PR #84, opened 2015-01-19
81 promote single classes

### LINKED ISSUE #81: Change ClassName mixin to promote single classes
## Need

We need to default to a single class
And right now passing a class to a component appends it
And composing classes is against our principles

## Deliverables

- Passing a class to a component replaces default one
- Prop should be named `className` instead of `class`
- No need for any class properties for the default class. It's more semantic to be called inside the render function

skidding:
Fixes #81


skidding:
@NiGhTTraX please review

NiGhTTraX:
LGTM

PR #93, opened 2015-02-02
Redesign PersistState mixin

skidding:
- [x] No longer deep clone anything
- [x] No longer alter props!
- [x] No longer clear childProps cache inside a getter!
  - [x] Do it on `componentDidUpdate` and `componentDidMount`
    - [x] Think about when shouldComponentUpdate returns false
- [x] Extend getInitialState when overriding state from snapshot
- [x] Name and order private methods accordingly
- [x] Refactor and improve style of unit tests (so hard :cry:)
- [x] Find a better name
  - [x] `ComponentTree`

Part 2:

- [x] No longer inject state on props
- [x] Update Docstring
- [x] Update Mixins wiki page
- [x] Process loadChild arguments without leaks
- New specs
  - [x] Inject state with custom refs
  - [x] ~~Serialize with custom refs~~ No need, serializing [always worked with refs directly](https://github.com/skidding/cosmos/blob/301a14b8f0ae25fa2516e66ada3d7776bc4ee0fc/mixins/persist-state.js#L37)

### DIFF
--- a/mixins/persist-state.js
+++ b/mixins/persist-state.js
@@ -37,7 +37,7 @@ Cosmos.mixins.PersistState = {
       _.each(this.refs, function(instance, ref) {
         // The child component needs to implement the PeristState mixin to be
         // able to serialize its children recursively as well
-        if (typeof(instance.serialize) == 'function') {
+        if (_.isFunction(instance.serialize)) {
           childSnapshot = instance.serialize(true);
 
           if (!_.isEmpty(childSnapshot.state)) {
--- a/mixins/persist-state.js
+++ b/mixins/persist-state.js
@@ -82,6 +82,7 @@ Cosmos.mixins.PersistState = {
      * @param {...*} [arguments] Optional extra arguments get passed to the
      *                           function that returns the component props
      */
+    // https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments
     var args = [];
     for (var i = 1; i < arguments.length; ++i) {
       args[i - 1] = arguments[i];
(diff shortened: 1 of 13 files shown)


skidding:
Please review @NiGhTTraX

skidding:
> Think about when shouldComponentUpdate returns false

There are two cases when state is flushed:
1. `componentDidMount`
2. `componentDidUpdate`

#### componentDidMount

`componentDidMount` will always be called, so `shouldComponentUpdate` returning false changes nothing here.

#### componentDidUpdate

Here is the worst case scenario:

Say there is one component with children and it has `shouldComponentUpdate` returning false.

1. Mount/render component
2. Load nested state through `setProps`
3. Change child state, either through `refs.child.setState` or by user interaction in the child
4. Calling `forceUpdate` on component will result in the children receiving the state passed in 2. and overriding the one set in 3.

Considering props.state is usually only used when mounting, when loading a serialized state from a fixture or some snapshot, this is very unlikely to happen. I think it's way better than the previous compromises.

Worst case, we remove the [functionality to pass state after mounting](https://github.com/skidding/cosmos/blob/b890f390180bc2c1fad931bbe015e0ac131784bd/mixins/persist-state.js#L129-L138) altogether.

mixins/persist-state.js
```diff
@@ -28,28 +26,38 @@ Cosmos.mixins.PersistState = {
...
 
     if (recursive) {
       _.each(this.refs, function(instance, ref) {
         // The child component needs to implement the PeristState mixin to be
         // able to serialize its children recursively as well
-        if (typeof(instance.generateSnapshot) == 'function') {
-          children[ref] = instance.generateSnapshot(true).state;
-        } else {
-          children[ref] = _.cloneDeep(instance.state) || {};
+        if (typeof(instance.serialize) == 'function') {
```
NiGhTTraX: Triple equal :)

PR #127, opened 2015-05-07
Add hot reloading for React components

### LINKED ISSUE #124: Playground integration with react-hot-loader

gaearon:
Fixes #124.
Note that hot reloading only works for React components; changing a fixture will force a full reload.

It may be possible to do hot reloading for fixtures too but to be honest I couldn't find where they were loaded in the source. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ specified via cli args:
 node_modules/.bin/component-playground --components-path src/components --fixtures-path tests/fixtures
 ```
 
+Finally, Cosmos includes [React Hot Loader](http://gaearon.github.io/react-hot-loader/) and has webpack's [hot module replacement](http://webpack.github.io/docs/hot-module-replacement.html) enabled so you can tweak the components and their styles without refreshing the browser:
+
+![React Hot Loader in Cosmos](https://cloud.githubusercontent.com/assets/250750/7522175/2e839f04-f4fc-11e4-890c-e60505c50f57.gif)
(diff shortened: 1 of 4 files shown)


skidding:
Awesome!

Tested and it works great
cosmos

Please bump the version to 0.6.2 and feel free to write a README paragraph about the react-hot-loader integration.

gaearon:
Done!

skidding:
Patch is already on [npm](https://www.npmjs.com/package/cosmos-js). Thanks! 

> changing a fixture will force a full reload.

That makes sense. At least for now, since it would interfere with the changes you make inside the playground fixture editor.

> It may be possible to do hot reloading for fixtures too but to be honest I couldn't find where they were loaded in the source. :-)

Some breadcrumbs...

- [(optional) Alias to fixtures folder in case you want a custom path](https://github.com/skidding/cosmos/blob/master/component-playground/webpack.config.js#L24)
- [Tree of fixtures grouped per component sent to ComponentPlayground](https://github.com/skidding/cosmos/blob/master/component-playground/entry.js#L11)
- [An entire folder is mapped recursively using webpack context](https://github.com/skidding/cosmos/blob/master/component-playground/lib/get-component-fixture-tree.js#L4)
- [The contents of each fixture is extracted](https://github.com/skidding/cosmos/blob/master/component-playground/lib/get-component-fixture-tree.js#L20)

PR #126, opened 2015-05-07
Use babel-loader instead of jsx-loader

skidding:
Even Facebook uses Babel. Time to switch!


skidding:
:hammer:

For the record: http://kangax.github.io/compat-table/es6/

PR #140, opened 2015-06-11
Added server config section

### LINKED ISSUE #139: Configure port and host of the server
I need to dockerize cosmos, but I can't, because it only listens on localhost, not '0.0.0.0'. 
What would be a preferred way to set port and host of the server? I am willing to work on this feature.

w0rm:
Closes #139

@skidding I've introduced another config section for the server, but I'm not sure about extracting getUserConfig part into separate lib, why do you think it makes sense here?

### DIFF
--- a/bin/component-playground.js
+++ b/bin/component-playground.js
@@ -7,7 +7,8 @@ var path = require('path'),
 var rootPath = path.join(__dirname, '..'),
     playgroundPath = path.join(rootPath, 'component-playground');
 
-var compiler = webpack(require(path.join(playgroundPath, 'config')).webpack);
+var userConfig = require(path.join(playgroundPath, 'config'));
+var compiler = webpack(userConfig.webpack);
 
 var server = new WebpackDevServer(compiler, {
   contentBase: path.join(playgroundPath, 'public'),
@@ -15,4 +16,4 @@ var server = new WebpackDevServer(compiler, {
   hot: true
 });
 
-server.listen(8989, 'localhost', function() {});
+server.listen(userConfig.server.port, userConfig.server.hostname, function() {});
--- a/bin/component-playground.js
+++ b/bin/component-playground.js
@@ -7,7 +7,7 @@ var path = require('path'),
 var rootPath = path.join(__dirname, '..'),
     playgroundPath = path.join(rootPath, 'component-playground');
 
-var userConfig = require(path.join(playgroundPath, 'config'));
+var config = require(path.join(playgroundPath, 'config'));
 var compiler = webpack(userConfig.webpack);
 
 var server = new WebpackDevServer(compiler, {
@@ -16,4 +16,4 @@ var server = new WebpackDevServer(compiler, {
   hot: true
 });
 
-server.listen(userConfig.server.port, userConfig.server.hostname, function() {});
+server.listen(config.server.port, config.server.hostname, function() {});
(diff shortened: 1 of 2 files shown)


bin/component-playground.js
```diff
@@ -7,7 +7,8 @@ var path = require('path'),
 var rootPath = path.join(__dirname, '..'),
     playgroundPath = path.join(rootPath, 'component-playground');
 
-var compiler = webpack(require(path.join(playgroundPath, 'config')).webpack);
+var userConfig = require(path.join(playgroundPath, 'config'));
```
skidding: This contains the user config plus all the Cosmos defaults, so let's call it just `config`.

bin/component-playground.js
```diff
@@ -15,4 +16,4 @@ var server = new WebpackDevServer(compiler, {
   hot: true
 });
 
-server.listen(8989, 'localhost', function() {});
+server.listen(userConfig.server.port, userConfig.server.hostname, function() {});
```
skidding: This line has over 80 chars, and it failed the build: https://travis-ci.org/skidding/cosmos/builds/66354068#L401

(Will probably be fixed once you rename to `config`, though)

skidding:
Finished review

bin/component-playground.js
w0rm: Oops, sorry for that. I've configured my atom with jscs linter plugin to avoid this in the future.

bin/component-playground.js
```diff
@@ -7,7 +7,8 @@ var path = require('path'),
 var rootPath = path.join(__dirname, '..'),
     playgroundPath = path.join(rootPath, 'component-playground');
 
-var compiler = webpack(require(path.join(playgroundPath, 'config')).webpack);
+var config = require(path.join(playgroundPath, 'config'));
+var compiler = webpack(userConfig.webpack);
```
skidding: Is this PR ready? This should be `config.webpack`.

w0rm: I'm sorry for this!

skidding:
There's still a problem with this PR. You need to set the custom port for the HMR entry point as well: https://github.com/skidding/cosmos/blob/master/component-playground/config.js#L30

w0rm:
@skidding fixed port setting for HMR.

component-playground/config.js
```diff
@@ -24,10 +26,15 @@ try {
...
+config.server = _.extend({
+  port: 8989,
+  hostname: 'localhost'
+}, userConfig.server);
+
+config.webpack = {
   context: playgroundPath,
   entry: [
-    'webpack-dev-server/client?http://localhost:8989',
+    'webpack-dev-server/client?http://localhost:' + config.server.port,
```
skidding: You also need to replace localhost with `' + config.server.hostname + '`.

Please test this yourself as well afterwards.

PS. It's going to be > 80 chars so something like this would do:

```js
    'webpack-dev-server/client?http://' + config.server.hostname + ':' + 
                                          config.server.port,
```

w0rm: Sorry, I just don't have the whole thing at home, will try to clone your cosmos-example project and test with it

skidding:
Everything seems in order now. Thanks for the contribution!

w0rm:
Sorry for not doing it right at the first time

PR #142, opened 2015-07-06
External stylesheets/scripts

pekala:
It might be a useful piece of advice for people who are not there yet with complete encapsulation of components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ module.exports.webpack = function(config) {
 You can use this functionality to inject external styles or scripts if your
 components need e.g. Bootstrap to work.
 ```js
-config.entry.push('../../../injectStyleTag.js');
+config.entry.push(path.join(process.cwd(), 'injectBootstrapTags.js'));
 ```
 
 #### Hot loading
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ module.exports.webpack = function(config) {
 };
 ```
 
+You can use this functionality to inject external styles or scripts if your
+components need e.g. Bootstrap to work.
+```js
+config.entry.push('../../../injectStyleTag.js');
+```
+
 #### Hot loading
 
 Cosmos includes [React Hot Loader](http://gaearon.github.io/react-hot-loader/)


README.md
```diff
@@ -87,6 +87,12 @@ module.exports.webpack = function(config) {
 };
 ```
 
+You can use this functionality to inject external styles or scripts if your
+components need e.g. Bootstrap to work.
+```js
+config.entry.push('../../../injectStyleTag.js');
```
skidding: This example is a bit confusing, why would be the extra file be outside the project (`../../..`), since the config file is located in the root of your project, right?

Also, since you gave Bootstrap as an example, it might be useful to have the bootstrap file in the example as well. (unless you need many extra files and it looks ugly, I guess...)

skidding:
I'm glad you managed to make this work, it will be useful for users I'm sure! I only have one [comment](https://github.com/skidding/cosmos/pull/142#discussion_r34087772) with the PR

README.md
pekala: > Also, since you gave Bootstrap as an example, it might be useful to have the bootstrap file in the example as well. (unless you need many extra files and it looks ugly, I guess...)

I suppose most use cases would include external file from CDN, in which case you need a custom JS file to inject actual script/style tags.

> This example is a bit confusing, why would be the extra file be outside the project (../../..), since the config file is located in the root of your project, right?

Using `./injectStyleTag.js` would look for the file in cosmos project inside `node_modules`. In this case the file `injectStyleTag.js` is actually in the root of my project. I agree it looks a bit odd, but I'm not sure how to get around it.

skidding: > I suppose most use cases would include external file from CDN, in which case you need a custom JS file to inject actual script/style tags.

So webpack doesn't support external files as entry points?

> Using ./injectStyleTag.js would look for the file in cosmos project inside node_modules. In this case the file injectStyleTag.js is actually in the root of my project. I agree it looks a bit odd, but I'm not sure how to get around it.

Maybe you could use `path.join(process.cwd(), 'injectStyleTag.js')`

pekala: > Maybe you could use path.join(process.cwd(), 'injectStyleTag.js')

Duh, that will work :)

> So webpack doesn't support external files as entry points?

Not that I'm aware of. I tried to search for this, but couldn't find anything built in.

skidding:
Merged :tada:

PR #146, opened 2015-07-26
Generates documentation from propTypes

pekala:
Discussed here: https://github.com/skidding/react-component-playground/issues/25

Adds a `docs` property to every component passed to React Component Playground. 
The value of the property is a string containing HTML with component's documentation generated using [`react-docgen`](https://github.com/reactjs/react-docgen). A custom Webpack loader is handling this generation using `react-docgen` and [JSON to Markdown converter](https://github.com/reactjs/react-docgen/blob/master/example/generateMarkdown.js) from `react-docgen` examples. Finally the resulting markdown is passed through markdown and html Webpack loaders to get the string.

This is not mergable yet, as the tests are not updated. But please, have a look and let me know if you like the approach.


skidding:
Do you have a working example? It would be nice if you could post a screeenshot or gif. I'm currently on vacation and am not able to try it out myself. I return in 2 weeks so sorry in advance for not being responsive.

A few notes:
- Passing html through props seems a bit heavy, and makes the component playground a bit harder to serialize. One thing that I wanted to for a while is install Cosmos inside the playground repo to test the playground with itself. This might make it harder. Did you consider passing the JSON or the Markdown string to the playground, for it to turn it into HTML (a declaration markdown to HTML conversion in the render function would be cool if possible...).
- Since the project already depends on Lodash, feel free to use it for brevity. It felt like there were some places where you could benefit from its functions. It also has a string template API which might make the markdown generating code more readable.

pekala:
> Passing html through props seems a bit heavy, and makes the component playground a bit harder to serialize. One thing that I wanted to for a while is install Cosmos inside the playground repo to test the playground with itself. This might make it harder. Did you consider passing the JSON or the Markdown string to the playground, for it to turn it into HTML (a declaration markdown to HTML conversion in the render function would be cool if possible...).

Sure, should be easy to defer conversion to component playground.

> Since the project already depends on Lodash, feel free to use it for brevity. It felt like there were some places where you could benefit from its functions. It also has a string template API which might make the markdown generating code more readable.

The markdown generation code is copied from `react-docgen` [example project](https://github.com/reactjs/react-docgen/blob/master/example/generateMarkdown.js). Maybe we could just require it from their package instead, although that might be a bit risky. But I can refactor the code to make use of lodash to make it nicer if you feel like that's the way to go.

skidding:
>  Maybe we could just require it from their package instead, although that might be a bit risky

Yeah, I think it's best to copy it since it's an example.

> But I can refactor the code to make use of lodash to make it nicer if you feel like that's the way to go.

I thought you wrote it by hand, but since it's taken from the react-docgen repo and needs no change there's no point in rewriting it.

skidding:
@pekala while the repo has been through major changes since this PR, I still think it's a good idea. 

It's also easier to implement this now. We could:
1. Add component meta data along with components and fixtures in the [webpack entry](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-cosmos-webpack/src/webpack-config.js#L38-L44)
2. Create a Proxy that renders the props info next to the rendered component (see the [ReduxProxy](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-cosmos-redux-proxy/src/index.jsx) for a Proxy example)
3. Bonus: Stop rendering component altogether if fixture is missing required props. This prevents nasty errors and can provide an onboarding experience for users, telling them which props are missing and overall instructions on creating fixtures.

Let me know if this still interests you, I can provide more assistance.

skidding:
Closing this due to inactivity. Feel free to revive.

ovidiubute:
I'd like to necro this PR, it's a very useful feature to have documentation alongside components. @skidding should we create a new issue? Do you feel that the code here is still a good starting point?

skidding:
The code here is definitely incompatible with the current state of RC, but I think it should be read by anyone looking to restart the same idea. Also keep in mind that `react-docgen` might've evolved since 2015. @ovidiubute A new issue that links to this PR would be best.

PR #161, opened 2015-08-20
resolve #155 adding support to newer webpack loaders to support css modules

### LINKED ISSUE #155: Cannot run `css-loader?modules`
I am trying to migrate a site to using CSS within components. One of the major reasons for doing so it so be able to use cosmos. The way I approach this problem is to first import the styles within the component:

```javascript
import styles from './bootstrap-layout.css';
```

Then use the style as the `className`:
```javascript
<div className={styles['col-xs-12']}>...</div>
```

I have the following in my `loaders` for `webpack`:
```
{test: /\.css$/, loader: "style-loader!css-loader?modules"}
```

This all works in regular `webpack`. I am trying to achieve the same using `cosmos`.

I added the following to the `component-playground.config.js` file:

```
module.exports.webpack = function(config) {
  config.module.loaders[1]= {
      test: /\.css$/,
      loader: 'style-loader!css-loader?modules'
    };
  return config;
};
```

But the imported `styles` object still remains `undefined`.

Any idea how to achieve this?  `Cosmos` seems to be the perfect use case for css-modules.

cef62:
Debugging with @fabiobiondi we found that the loaders used by the playground are older than the introduction of css modules support.
Please accept this small pull request to permit us to use cosmos with css modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,14 +9,14 @@
   "dependencies": {
     "babel-core": "^5.2.17",
     "babel-loader": "^5.0.0",
-    "css-loader": "^0.9.1",
-    "less": "^2.4.0",
+    "css-loader": "^0.15.6",
+    "less": "^2.5.1",
     "less-loader": "^2.2.0",
     "lodash": "^3.5.0",
     "react-component-playground": "^0.3.0",
     "react-hot-loader": "^1.2.7",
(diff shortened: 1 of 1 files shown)


skidding:
Published on npm under `cosmos-js@0.6.7`

PR #162, opened 2015-08-20
Use latest JS features in Babel

bbirand:
Since we're so cutting edge, might as well use the `stage=1` features of `babel`. I usually add this on my own version..

### DIFF
--- a/component-playground/config.js
+++ b/component-playground/config.js
@@ -53,7 +53,7 @@ config.webpack = {
     loaders: [{
       test: /\.jsx?$/,
       exclude: /node_modules/,
-      loaders: ['react-hot-loader', 'babel-loader']
+      loaders: ['react-hot-loader', 'babel-loader?stage=1']
     }, {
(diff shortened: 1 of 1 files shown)


skidding:
Stage 1 features are experimental and subject to change. They are cool, but I want Cosmos to be as accessible as possible by default. E.g. I'm using it at work and if I merged this I'd have to override the option because I don't want to enable my colleagues to use experimental features in Enterprise software.

I am thinking, however, on a way to make the lib easier to configure, and that's something that will be improved in the future.

lapanoid:
@skidding maybe there could be "next" branch which could be less conservative to contributions? There we could develop and iterate fast and merge at some point with master. I am thinking about io-node story as a successful example of such model.

skidding:
Cosmos is now a [boilerplate](https://github.com/skidding/cosmos/tree/master/webpack-boilerplate/react-0.13). You can now easily config Babel per project.

PR #172, opened 2015-11-10
Added support for url-loader.

n1k0:
This adds support for loading external assets, eg. for when requiring external stylesheets using images or fonts (think bootstrap for example).

Credit: https://github.com/webpack/css-loader/issues/38#issuecomment-72287584


skidding:
Thanks! This is included in `0.6.9`

PR #176, opened 2015-11-21
Abstract playground files under npm package

skidding:
## Problem

The Cosmos setup requires you to add two playground-specific files to your project:
- [Webpack entry file](https://github.com/skidding/cosmos/blob/d30ee586da6818f5a0059fb39384d22dcbf174dc/webpack-boilerplate/react-0.14/playground/index.js)
- [Function for traversing through your project files and builds a tree of component with fixtures](https://github.com/skidding/cosmos/blob/d30ee586da6818f5a0059fb39384d22dcbf174dc/webpack-boilerplate/react-0.14/playground/get-component-fixture-tree.js)

I was already thinking of abstracting these somehow as they make the integration less pleasant, when @goatslacker also suggested it on [Gitter.](https://gitter.im/skidding/cosmos)

The reason I was hesitant to *hide* these files was because they make customization options obvious. However, since I don't remember anybody customizing the way Cosmos expects to find components and fixtures within a project, I'm going to make the assumption that **customizing the files above is more an exception than a rule.**

## Solution

- [x] Bring back `cosmos-js` package (now becoming `0.7`), containing `entry.js` and `get-component-fixture-tree.js`
- [x] Use `cosmos-js` as entry inside the [playground webpack config](https://github.com/skidding/cosmos/blob/d30ee586da6818f5a0059fb39384d22dcbf174dc/webpack-boilerplate/react-0.14/webpack.config.playground.js#L8)
- [x] `cosmos-js` dependency replaces `react-component-playground` and `react-querystring-router` dependencies
- [x] Update [README section](https://github.com/skidding/cosmos/blob/d30ee586da6818f5a0059fb39384d22dcbf174dc/webpack-boilerplate/react-0.14/README.md#playground-files) to reflect how would one go about customizing these files once they're no longer part of the host project.
- [ ] ~~Update `react-0.14` boilerplate and add note under `react-0.13` stating that it no longer receives update and latest setup is found under `react-0.14`~~ #179 is a better idea!


skidding:
@goatslacker wdyt?

src/get-component-fixture-tree.js
```diff
@@ -43,7 +41,7 @@ module.exports = function() {
     };
 
     // Allow users to browse components before creating fixtures
-    if (_.isEmpty(components[componentName].fixtures)) {
```
catalinmiron: :+1:

goatslacker:
This is good I just have a note on React 0.14/0.13 compatibility. How about you support both?

You can create a getDOMNode function like

```js
function getDOMNode(node) {
  return React.version.indexOf('0.14') > -1 ? node : node.getDOMNode();
}
```

You can then also conditionally load ReactDOM and use it to render. That seems to be all you're using React 0.14 for.

Also need to update the peer dependency requirements. You can probably even get away with 0.12 support.

skidding:
> This is good I just have a note on React 0.14/0.13 compatibility. How about you support both?

 Good idea, created #179 for this.

PR #180, opened 2015-12-05
terminate regex for fixture files

DanLipsitt:
Fix the regex for fixture files so it only matches `*.js` files, not `*.js*` files. This was causing bugs for me because it was picking up my editor's backup files (e.g. `foo.js~`).


skidding:
Patch published under `0.7.1` :+1:

PR #181, opened 2016-02-11
account for es6 default classes (compiled with Babel)

narenranjit:
I was having troubling getting the following component to render within the playground:

```
import React from 'react';

export default function PP() {
    return (
        <h1>
            HI
        </h1>
    );
}
```
Turned out the reason was because this exports as `{ default : function }` instead of just `function`. 

Hand-tested with other non-es6 commonjs exports and it seems fine, happy to write other test-cases if needed.

### DIFF
--- a/src/get-component-fixture-tree.js
+++ b/src/get-component-fixture-tree.js
@@ -22,6 +22,10 @@
   *   }
   * }
   */
+ 
+var normalizeModule = function (module) {
+    return (module.default) ? module.default : module;
+};
 module.exports = function() {
   var requireComponent = require.context('COSMOS_COMPONENTS', true),
       isComponent = /^\.\/(.+)\.jsx?$/,
@@ -36,7 +40,7 @@ module.exports = function() {
     // Fixtures are grouped per component
     var componentName = match[1];
     components[componentName] = {
-      class: requireComponent(componentPath).default ? requireComponent(componentPath).default: requireComponent(componentPath),
+      class: normalizeModule(requireComponent(componentPath)),
       fixtures: getFixturesForComponent(componentName)
     };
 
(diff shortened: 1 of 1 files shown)


ovidiu-chereches-hs:
Hi @narenranjit,

Can you tweak one of the [components from the boilerplate](https://github.com/skidding/cosmos/tree/5e99788e2f1eb2907a7bcf965dfa372ec0098e56/webpack-boilerplate/react-0.14/src/components) to use the es6 `exports default` syntax, in a way that doesn't work without your patch?

I tried something like
```diff
-module.exports = StatelessButton;
+export default StatelessButton;
```
and it worked out of the box, probably because [Babel already defaults to the `.default` key](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=import%20x%20from%20'foo'%3B%0Avar%20bar%20%3D%20require('bar')%3B). 
```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }
```

But maybe I'm missing something, or does your setup have some particularity?

narenranjit:
Seems to be a Babel 5 vs. Babel 6 thing - works fine in 5 but breaks in 6. I'm having trouble finding the behavior change documented in an official changelog strangely but [this stackoverflow link](http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default) seems to be the canonical source everyone is pointing to.

Or of course you could just add a note saying babel 6 isn't completely supported in the readme instead.

narenranjit:
Will put up a 'breaking' branch later today to illustrate.

skidding:
> Will put up a 'breaking' branch later today to illustrate.

Any luck? It would be great to upgrade the boilerplate to Babel 6 altogether. If you already have the knowledge feel free to create a PR for it.

skidding:
Fixed in #184

PR #189, opened 2016-08-23
Turn repo into monolith, add support for React v0.12 to v15

### LINKED ISSUE #179: Unify 0.13 and 0.14 boilerplates
By @goatslacker in #176 

> You can create a getDOMNode function like

> ```js
function getDOMNode(node) {
  return React.version.indexOf('0.14') > -1 ? node : node.getDOMNode();
}
```

> You can then also conditionally load ReactDOM and use it to render. That seems to be all you're using React 0.14 for.

> Also need to update the peer dependency requirements. You can probably even get away with 0.12 support.

### PR DESCRIPTION
**Pull request by skidding**
Please take a look @NiGhTTraX @bogdan-stefan-hs @catalinmiron @flaviusone @ovidiubute @RadValentin 👯 

[**NEW react-cosmos BRANCH**](https://github.com/skidding/cosmos/tree/react-cosmos)

Two big changes:
- Added support for most React versions (both pre and post `react-dom`) in same package. In order to allow all the goodness from the uberVU branch to merge into master & to make the lib more available to other React devs who didn't yet upgrade to >=0.14
- Merged all packages into one repo. I realised Cosmos dev sucks when I started working on `react-dom-polyfill` so I rethought the entire project from scratch. Dev was always a pain. I suspect @catalinmiron will like the use of Lerna ;) - [See packages](https://github.com/skidding/cosmos/tree/react-cosmos/packages)

Also, the `cosmos-js` pkg becomes `react-cosmos`. Will rename the repo as well once this branch gets merged.

Can you please test if the new packages (all under npm v`1.0.0-beta.1`) still work on your use case? The public API should be identical. ~~with a single distinction: `ComponentPlayground` is no longer bundled, just transpiled, so the parent project needs to ensure the LESS bundling:~~ (https://github.com/skidding/cosmos/pull/189#issuecomment-242014157)

As for the development, it's now very enjoyable. Didn't get to write the docs yet, but here is the gist:

Starting full playground instance, built from the source:
```bash
npm install
npm run bootstrap
npm start
```

Working on a specific module (playground will rebundle on file change):
```bash
# After install & bootstrap
npm run build:component-tree -- -- -w
npm start
```

Building everything with different React version: 👌 
```bash
# After install & bootstrap
npm run install-react:0.13
npm start
```

Note: Webpack will throw a warning when bundling with React <0.14, but everything still works, including HMR. Maybe something can be done in the future to get rid of it altogether (a noop alias, or loader?).

Let me know if I can provide other info. Thanks!

Closes #179


skidding:
Update: Decided the LESS loader shouldn't be in the user land, so [went back](https://github.com/skidding/cosmos/commit/92920391c26b41f980df06d63a160ab085c256b9) to bundling CP. This means the API is now exactly as before. Published the lastest change under **`react-cosmos@1.0.0-beta.1`** **`react-component-playground@1.0.0-beta.1`**.

Note: CP [no lunger bundles any JS deps](https://github.com/skidding/cosmos/blob/acf22567ed063c0244fa9680a3fcecd6c47e2870/packages/react-component-playground/webpack.config.js#L16-L26), which makes it lighter and no longer duplicate libs in the user build.

RadValentin:
screen shot 2016-08-24 at 15 07 29
9270086a-662d-11e6-871b-2ec3d6e02164

skidding:
> +5,201 −1,847

@RadValentin those numbers bear little importance. Most of the code is the source & tests copied from each satellite repo. Look at the [file](https://github.com/skidding/cosmos/tree/react-cosmos) [structure](https://github.com/skidding/cosmos/tree/react-cosmos/packages) instead.

The most helpful thing right now would be to validate that the new packages `...@1.0.0-beta.1` work on your codebase(s) (hs-components is probably the easiest) without any intervention besides switching the package versions. If you do get the chance, please keep me posted. Thanks!

bogdan-stefan-hs:
@skidding I can confirm hs-components' playground environment is working with:

- `react-component-playground` version `1.0.0-beta.1`
- `react-querystring-router` version `1.0.0-beta.0`
- `react-component-tree` version `1.0.0-beta.0`

Does the new `react-cosmos` package need testing as well?

skidding:
@bogdan-stefan-hs awesome!

> Does the new react-cosmos package need testing as well?

Yes if you're using it internally, otherwise don't bother.

flaviusone:
❤️  the idea to turn this into a monolith. Amazing job! GG! 
swag

packages/react-component-tree/src/load-child-component.js
```diff
@@ -0,0 +1,10 @@
+var React = require('react'),
+    loadChild = require('./load-child.js');
+
+class LoadChildComponent extends React.Component {
+  loadChild(childName, a, b, c, d, e, f) {
```
catalinmiron: TODO for the future?

```javascript
loadChild(childname, ...args) {
  return loadChild.loadChild(this, childName, ...args);
}
```

test/karma.conf.js
```diff
@@ -0,0 +1,50 @@
...
+          helpers: path.join(__dirname, 'helpers'),
+          fixtures: path.join(__dirname, 'fixtures'),
+          'component-playground': path.join(__dirname, '../packages/react-component-playground/src')
+        }
+      },
+      module: {
+        loaders: [{
+          test: /\.jsx?$/,
+          exclude: /node_modules/,
+          loader: 'babel',
```
catalinmiron: I made a loader that wraps each describe into a describe with file path, you can use it: https://www.npmjs.com/package/describe-loader

```
// npm install --save-dev describe-loader

postLoaders: [{
    test: /\.js$/,
    exclude: /node_modules/,
    loader: 'describe'
  }]
```

catalinmiron:
LGTM. gg 🍶

packages/react-component-tree/src/load-child-component.js
skidding: This might actually be more efficient, stole it from the React source code a while ago. But I'm going to ES6 the s*it out of this repo before merging. Wanted to make sure everything works first before I "celebrate". :D

test/karma.conf.js
skidding: Pretty cool. Not sure if this repo would benefit, tho, since the test cases are pretty verbose and it doesn't have the crazy nesting you guys have (not to mention the repetitiveness between similar components).

image

bogdan-stefan-hs:
@skidding The uberVU playground additions are all written in pre-ES6 code, to be in line with the rest of the repo. Can I suggest we ES6-fy it after we've merged those features? (I'm only talking about `react-component-playground`) 

Also, shall I wait until you merge this, or can I submit PRs with `react-cosmos` as a base?

skidding:
> @skidding The uberVU playground additions are all written in pre-ES6 code, to be in line with the rest of the repo. Can I suggest we ES6-fy it after we've merged those features? (I'm only talking about react-component-playground)

Makes sense, I will restrain myself!

> Also, shall I wait until you merge this, or can I submit PRs with react-cosmos as a base?

Wait for the merge. I'm on it. 🚂

bogdan-stefan-hs:
@skidding sorry for being late, but i think you also need to add `react-addons-testutils` to `react-dom-polyfill`. It has been separated to another package in React v0.14. 
Because of this, `npm test` fails on React <0.14.

skidding:
@bogdan-stefan-hs done! The project will remain version as beta for a while, to make for a great 1.0 release, but everything is on master. Fire away! 💣 

Muchas gracias, amigos!

skidding:
> @skidding sorry for being late, but i think you also need to add `react-addons-testutils` to `react-dom-polyfill`. It has been separated to another package in React v0.14. 
Because of this, `npm test` fails on React <0.14.

I figure `react-dom-polyfill` is for libs, which only publish production code, which shouldn't include testing logic. It's the user's concern to know how to use those test utils in development, based on his or her React version.

Does the fact that it's no covered by `react-dom-polyfill` cause problems for you?

bogdan-stefan-hs:
>Does the fact that it's no covered by react-dom-polyfill cause problems for you?

Not really, i can modify the code accordingly.

It's just the thing that this command sequence:
```
npm install
npm run bootstrap
npm run install-react:0.13
npm test
```

will fail out of the box.

Maybe something in the docs to warn the devs is enough, then?

skidding:
@bogdan-stefan-hs you're right, I forgot to mention that. To be frank, this is the current status:

- The lib works with React 0.12 to 15
- It can be built & live tested visually via the dev setup with React 0.12 to 15
- The automated tests only run with React 0.14 and above. (This should be mentioned in the readme.)

Not ideal, but you can:
- Switch to 0.13
- Develop with playground open for visual feedback and live testing
- Switch back to 15 to run automated tests

I'm thinking it's safe enough. If you think it's feasible to cover the test utils by `react-dom-polyfill` then feel free to come up with a solution (spec or code, we can keep an issue open for it). I wanted to see this mono repo done and skipped that step. :)

PR #190, opened 2016-08-25
Make fixture navigation panel scroll to selected component after mount

bogdan-stefan-hs:
## Need
```gherkin
As a user
I want the fixture panel to scroll to the currently selected component
So that I can easily locate the fixture/component I am viewing.
```
## Deliverables
- [x] On page load, the component panel is scrolled exactly to the selected fixture.

## Solution
Scroll to selected fixture's component after the main component finishes rendering.

### DIFF
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -280,6 +280,11 @@ module.exports = React.createClass({
     window.addEventListener('resize', this.onWindowResize);
 
     this._updateContentFrameOrientation();
+
+    if (this.props.component) {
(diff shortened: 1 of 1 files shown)


bogdan-stefan-hs:
@skidding 
The question here is: are you ok with using `behavior: smooth`?

It's only supported in Firefox, but on other browsers it simply does nothing (only tested on Chrome).

skidding:
> The question here is: are you ok with using `behavior: smooth`

Sure. 

I was gonna say something about the missing whitespace...
where-is-the-space
...but I remembered I'm gonna ES6 & lint all the repo soon. 

Gg! ✋

PR #196, opened 2016-08-30
Dom polyfill docs & tests

skidding:


skidding:
@bogdan-stefan-hs please review

skidding:
Coverage is [fine](https://coveralls.io/builds/7669043/source?filename=packages%2Freact-dom-polyfill%2Fsrc%2Findex.js#L9), that try catch is there for webpack's static analysis.

flavius-tirnacop-hs:
Looking good 👍

bogdanstefan:
Looks great! 👍 
Did you get the chance to test it with another bundler besides webpack to see how it acts when `react-dom` is missing?

skidding:
> Did you get the chance to test it with another bundler besides webpack to see how it behaves when react-dom is missing?

No. But good idea. I'll try a Browserify setup before we submit any external PR. 👍

PR #197, opened 2016-09-01
Replace textarea editor with Codemirror

bogdanstefan:
## Need

```gherkin
As a user
I want a powerful and feature-rich editor
So that I can edit fixtures with ease.
```

## Deliverables 
- [X] A CodeMirror instance with code editor features, as opposed to a plain `textarea`. This editor will have the following features:

 - Code folding
 - Syntax highlighting
 - Smart indentation
 - Line numbers

## Solution
Use [`react-codemirror`](https://github.com/skidding/react-codemirror) for a CodeMirror React component and adapt it for Javascript syntax.


bogdanstefan:
@skidding Please review the changes.

You also need to take a look at the changes in the [`react-codemirror` fork](https://github.com/skidding/react-codemirror/tree/use-react-dom-polyfill), as I also needed to modify some things there to make the tests pass.

Coverage seems fine, it decreased because there's more code now.

skidding:
This is what I recommend for the CSS:

- Make the first style loader (that only includes `src`) accept both `.css` and `.less` and use the `:global` CSS modules modifier in `codemirror.css`:
```css
:global {
  .CodeMirror, .ReactCodeMirror {
    width: 100%;
    height: 100%;
  }
}
```
- Make the 2nd loader only include `node_modules`.

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -398,9 +398,8 @@ module.exports = React.createClass({
   },
 
   onFixtureChange: function(event) {
-    var userInput = event.target.value,
+    var userInput = event,
```
skidding: Let's name the argument better, it's no longer an event but a value.

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -11,8 +11,14 @@ var _ = require('lodash'),
     isSerializable = require('../lib/is-serializable.js').isSerializable,
     SplitPane = require('ubervu-react-split-pane'),
     localStorageLib = require('../lib/local-storage.js'),
+    CodeMirror = require('react-codemirror'),
     fuzzaldrinPlus = require('fuzzaldrin-plus');
 
+require('codemirror/lib/codemirror.css');
+require('codemirror/addon/fold/foldgutter.css');
+require('codemirror/theme/solarized.css');
```
flaviusone: I agree with keeping this theme for customisation but can't we move the other .css required into the react-code-mirror fork (as we did with the .js requires). WDYT?

flaviusone:
LGTM

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -359,21 +370,10 @@ module.exports = React.createClass({
...
     }
-
-    // Focus on the editor when changing fixture, to prevent overriding
-    // its contents with the state generated by the initial unfolding of the
-    // rendered component tree
-    if (this.props.editor) {
-      this._focusOnEditor();
-    }
-  },
-
```
bogdanstefan: @skidding This was causing some trouble, as focusing the editor on fixture click would only provide it with the fixture contents. `onFixtureUpdate` wouldn't do anything, because it was focused. Then, on editor blur, it would also add the previewed component's serialization (things such as `state`, for example) and it looked pretty bad.

Are you ok with removing it? Maybe I misunderstood it.

packages/react-component-playground/src/components/component-playground.jsx
skidding: Let's keep the `react-code-mirror` with minimum changes, besides the `react-dom-polyfill` integration, to be able to submit a PR upstream. I'll take a closer look on what you added in that fork as well.

packages/react-component-playground/src/components/component-playground.jsx
skidding: Sounds good. I also think it was overkill and am 👌  with removing it.

packages/react-component-playground/webpack.config.js
```diff
@@ -3,6 +3,7 @@ var path = require('path'),
 
 var src = path.join(__dirname, 'src');
 var lib = path.join(__dirname, 'lib');
+var node_modules = path.join(__dirname, 'node_modules');
```
skidding: `var nodeModules` 😬

skidding:
Finished review. Will take a look at the React CodeMirror fork.

q8q4j

packages/react-component-playground/src/components/component-playground.jsx
flaviusone: @bogdan-stefan-hs had to move [this](https://github.com/uberVU/react-component-playground/blob/cf52b775fb6feec7de2792c429bf45ca14f87138/src/components/component-playground.jsx#L21-L25) js requires into the `react-code-mirror` fork because of some webpack reason. @bogdan-stefan-hs maybe you can explain what was happening there 💃

bogdanstefan: The thing is `react-codemirror` doesn't bundle any `codemirror` addons or styles, to keep the bundle size to a minimum.
JS `require` calls (for syntax highlight keyword detection and code folding) inject code into the global `codemirror` object.

However, webpack bundles `codemirror` twice, once for CP and once for `react-codemirror` and requires in CP are injecting code to the wrong instance. All I could come up with is move the `require` calls to `react-codemirror`, since we had the fork.
Maybe some dedupe plugin would work, i am not sure though...

skidding: This seems to work:
image
image
😎

skidding:
@bogdanstefan this does the trick. Commit it yourself because it's your branch. GG @flaviusone for saying that the test is the problem! 

image

The rest LGTM!

PR #200, opened 2016-09-06
Eslint takeover

skidding:
Integrated ESLint with the Airbnb preset. 😓 

@bogdanstefan @flaviusone @RadValentin wdyt?


flaviusone:
```js
/* eslint-env browser, mocha */
/* eslint-disable
  global-require,
  no-unused-vars,
  no-unused-expressions,
  import/no-unresolved,
  import/no-extraneous-dependencies
*/
/* global expect, sinon */
```
It seems that you have something like this in almost every test file. Why don't you make a custom eslintrc file just for the test folder that will dedupe a lot of this rules?

flaviusone:
image
image

flaviusone:
LGTM otherwise ¯\_(ツ)_/¯

catalinmiron:
> I know that the airbnb wants it like this but it still hurts 😆.

Hurts because you're not writing javascript ma friend:) 👔 🎉

bogdanstefan:
Just curious, why disable `no-unused-vars` in tests and not remove them?

Anyway, looks beautiful now. 👏

test/helpers/render-component.js
```diff
@@ -1,25 +1,25 @@
-var React = require('react'),
-    ReactDOM = require('../../packages/react-dom-polyfill')(React),
-    $ = require('jquery'),
-    ComponentTree = require('../../packages/react-component-tree'),
-    ComponentPlayground = require('../../packages/react-component-playground/src');
+/* eslint-env browser */
+/* eslint-disable react/no-find-dom-node */
```
valentin-radulescu-hs: Can this rule be part of the global config?

packages/react-dom-polyfill/test/pre-react-dom.js
```diff
@@ -1,3 +1,7 @@
+/* eslint-env node, mocha */
+/* eslint-disable react/no-find-dom-node */
+/* global expect, sinon */
```
valentin-radulescu-hs: How about making a tests config and including it for all test folders?

valentin-radulescu-hs:
👍  to what @flaviusone said, maybe you could do something about the annotations? Tests could do with a separate config. Other that that looks good!

valentin-radulescu-hs:
👌

PR #201, opened 2016-09-08
Add proxy support

### LINKED ISSUE #195: Provide redux store for rendered components
## Need

```gherkin
As a user
I want to be able to load components/fixtures that need a Redux store
So that I can preview them.
```

## Deliverables
Fixtures with state residing in a redux store do not throw an error on selection, and load normally.

## Solution

- Receive a `reduxStore` prop in CP.
- On mount, create a Redux store, cloning the state from `reduxStore` prop.
- Wrap the rendered component in a `<Provider>` component, with the initially created store.
- On fixture change, reset the store to the initial state, then merge the fixture's `reduxStore` data in it, if it provides one.

That way, a store is provided for components belonging to Redux-enabled apps.

bogdanstefan:
Closes #195.


bogdanstefan:
@skidding Can you take a look, please?

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -213,11 +218,19 @@ module.exports = React.createClass({
...
+        {this.renderComponent()}
       </div>
     );
   },
 
+  renderComponent() {
+    return _.reduce(this.props.proxies, (accumulator, proxy) =>
+      React.createElement(proxy.component, _.assign({}, this.props, {
+        children: React.Children.only(accumulator),
+        ref: proxy.ref,
```
skidding: Why do you need this?

bogdanstefan: It's not needed right now (except in tests), but I'm thinking we should have a way to access these proxies in the future.

skidding: I think this is a bad move going forward. [String refs are almost deprecated](https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute), so we should at least use callback refs. But callback refs also don't make sense here, because the CP shouldn't be aware of what proxies it is being passed, otherwise we might as well embed them inside. I suggest we only keep component classes/functions in the list of proxies. 

Is there any other reason besides tests to keep them? There must be a different way to check that the proxies are being loaded. E.g. Checking that their dom node(s) include the dom node(s) of the next proxy.

skidding:
@bogdanstefan how about reopening the PR into `master`, since `eslint-takeover` was already merged. Maybe merge `master` into it as well.

bogdanstefan:
@skidding How's this? Can't think of a cleaner way to test them, maybe use jQuery?

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -43,10 +43,7 @@ module.exports = React.createClass({
     editor: React.PropTypes.bool,
     fixture: React.PropTypes.string,
     fullScreen: React.PropTypes.bool,
-    proxies: React.PropTypes.arrayOf(React.PropTypes.shape({
-      component: React.PropTypes.func.isRequired,
-      ref: React.PropTypes.string,
-    })),
+    proxies: React.PropTypes.arrayOf(React.PropTypes.func),
```
skidding: Do all three types of React components match `React.PropTypes.func`? (React.createClass, ES6Class and stateless functions)

skidding:
> @skidding How's this? Can't think of a cleaner way to test them, maybe use jQuery?

Looking great! I only have [one question](https://github.com/skidding/react-cosmos/pull/201#discussion_r78558527) left.

packages/react-component-playground/src/components/component-playground.jsx
bogdanstefan: Yes, they do. Wasn't sure about ES6 classes but it seems [they do too](http://jsfiddle.net/eb8wbwsk/).

skidding: Awesome. Cosmos becomes a platform for 3rd party middleware! 👯

PR #202, opened 2016-09-13
Pin searchbar to left nav header

bogdanstefan:
If the component/fixture list has a big length, it's too time-consuming and unnecessary to scroll to the top to find the search bar.

Thus, it needs to be fixed to the top of the component list while scrolling.


bogdanstefan:
@skidding @flaviusone
Do you guys like it better this way? Is the code alright?

fixed searchbar

skidding:
Sweet! 🍬 

@bogdanstefan Made you a collaborator so merge at will. Not relevant this time, since you have only 1 commit, but remember to Squash and merge (below) to minimise the noise and have meaningful commits looking back.

image

flaviusone:
Great!

PR #203, opened 2016-09-14
Implement redux proxy

bogdanstefan:
This adds a proxy which creates a Redux store using `createStore` and passes it to the children's context.

It will be published as a separate package, `react-cosmos-redux-proxy`.

### DIFF
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -43,7 +43,10 @@ module.exports = React.createClass({
     editor: React.PropTypes.bool,
     fixture: React.PropTypes.string,
     fullScreen: React.PropTypes.bool,
-    proxies: React.PropTypes.arrayOf(React.PropTypes.func),
+    proxies: React.PropTypes.arrayOf(React.PropTypes.shape({
+      component: React.PropTypes.func,
+      props: React.PropTypes.object,
+    })),
     router: React.PropTypes.object,
   },
 
@@ -225,7 +228,7 @@ module.exports = React.createClass({
 
   renderComponent() {
     return _.reduce(this.props.proxies, (accumulator, proxy) =>
-      React.createElement(proxy, _.assign({}, this.props, {
+      React.createElement(proxy.component, _.assign({}, proxy.props, {
         children: React.Children.only(accumulator),
       })), this.loadChild('preview'));
   },
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -390,10 +390,7 @@ module.exports = React.createClass({
   componentDidUpdate(prevProps, prevState) {
     if (this.previewComponent && (
         this.constructor.didFixtureChange(prevProps, this.props) ||
-        prevState.fixtureChange !== this.state.fixtureChange ||
-        // Toggling the editor re-creates the preview component from scratch,
-        // because SplitPane is added or removed from part of the dom tree
-        prevProps.editor !== this.props.editor)) {
+        prevState.fixtureChange !== this.state.fixtureChange)) {
       this.injectPreviewChildState();
     }
   },
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -225,6 +225,11 @@ module.exports = React.createClass({
 
   renderComponent() {
     return _.reduce(this.props.proxies, (accumulator, proxy) =>
+      React.createElement(proxy, _.assign({},
+        this.constructor.getSelectedFixtureContents(this.props), {
+          children: React.Children.only(accumulator),
+        }
+      )), this.loadChild('preview'));
   },
 
   renderFixtures() {
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -43,10 +43,7 @@ module.exports = React.createClass({
     editor: React.PropTypes.bool,
     fixture: React.PropTypes.string,
     fullScreen: React.PropTypes.bool,
-    proxies: React.PropTypes.arrayOf(React.PropTypes.shape({
-      component: React.PropTypes.func,
-      props: React.PropTypes.object,
-    })),
+    proxies: React.PropTypes.arrayOf(React.PropTypes.func),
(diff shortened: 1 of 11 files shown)


bogdanstefan:
@skidding 
Can you review this?

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+      <div>
+        {React.createElement(type, Object.assign({}, props, { ref }))}
+      </div>
+    );
+  }
+}
+
+ReduxProxy.propTypes = {
+  children: React.PropTypes.element.isRequired,
+  store: React.PropTypes.object,
```
skidding: Any reason why this isn't also required?

test/fixtures/component-playground/proxies.js
```diff
@@ -29,6 +29,12 @@ const MarkupProxy = React.createClass({
 });
 
 module.exports = _.merge({}, selectedFixture, {
-  proxies: [PropMutatorProxy, MarkupProxy],
+  proxies: [{
+    component: PropMutatorProxy,
+    props: {},
```
skidding: This is odd. Isn't `proxies` a prop of the Component Playground and not a fixture property?

skidding: OK. I got it now, this is a fixture for CP, not one used by CP. My bad.

That being said, this means you can only send the contents of the store once to the ReduxProxy, global for all component/fixtures loaded? Is this correct?

If yes, I suggest the following alternative: Revert the props thing and continue to send the entire fixture to all proxies. This way, ReduxProxy continues to reads the contents of the store from `props.store`, but `store` will be a separate property on every fixture. Wdyt?

skidding:
This PR should also come with:
- One or more component/fixture examples using this proxy inside the `dev` folder.
- Some unit testing for the proxy component.  This is what I think matters:
  - Component creates Store using `props.store` (Redux store should probably be mocked)
  - Component attaches that store instance in context
  - Component renders children with right props

test/fixtures/component-playground/proxies.js
bogdanstefan: > That being said, this means you can only send the contents of the store once to the ReduxProxy, global for all component/fixtures loaded? Is this correct?

Yes, you're right, this is a problem.

> If yes, I suggest the following alternative: Revert the props thing and continue to send the entire fixture to all proxies.

If we do this, aren't we decoupling props from the proxies too much?
What happens if another proxy needs a key named 'store' as well, for completely different reasons? It will use the same value.

skidding: > What happens if another proxy needs a key named 'store' as well, for completely different reasons? It will use the same value.

To avoid this problem completely, we can make proxies be functions that return components, instead of components. E.g.

```js
proxies: [
  // Default 
  reactCosmosReduxProxy()
  // Custom key
  reactCosmosReduxProxy({ storeKey: '__reduxStore' }),
]
```

And inside the proxy:

```js
const reactCosmosReduxProxy = ({ storeKey }) =>
  class ReduxProxy extends React.Component {
    constructor(fixture) {
      super();
        this.store = createStore((state) => state, fixture[storeKey || 'reduxStore']);
    }
```

Notes:
- We could default to "reduxStore" instead of "store", to minimize such conflicts from the onset. 
- I suggest using `fixture` instead of `props` inside proxies. It sounds unorthodox, but I think is easier to understand. After all, the props of the preview component just happen to be all the fixture keys minutes the *magical* ones (like "state", and now "store", or "reduxStore").
  - Moreover, I think this proxy should omit the redux store key from the fixture when passing it down to children, just like ComponentTree does with "state" key. Children should receive just props (remaining fixture keys after magical ones).

packages/react-cosmos-redux-proxy/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "MIT",
+  "main": "lib/index.js",
+  "dependencies": {
+    "redux": "^3.6.0"
+  },
+  "peerDependencies": {
+    "react": ">=0.12 <16"
+  },
+  "devDependencies": {
+    "lodash": "^4.15.0"
```
skidding: Lodash appears to be needed in `dependencies`, not `devDependencies`.

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+        <div>
+          {React.createElement(type, Object.assign({}, _.omit(props, storeKey), { ref }))}
+        </div>
+      );
+    }
+  }
+
+  ReduxProxy.propTypes = {
+    children: React.PropTypes.element.isRequired,
+  };
```
skidding: You should be able to add the store in `propTypes` as well:
```js
  [storeKey]: React.PropTypes.object,
```
And we either make it required or add `defaultProps`:
```js
ReduxProxy.defaultProps = {
    [storeKey]: {},
};
```

packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -225,9 +225,11 @@ module.exports = React.createClass({
 
   renderComponent() {
     return _.reduce(this.props.proxies, (accumulator, proxy) =>
-      React.createElement(proxy, _.assign({}, this.props, {
-        children: React.Children.only(accumulator),
-      })), this.loadChild('preview'));
+      React.createElement(proxy, _.assign({},
+        this.constructor.getSelectedFixtureContents(this.props), {
```
skidding: I've missed this the 1st time, but this needs further tweaking. We should be consistent in what we send to proxies vs what we send to the preview component. [THIS](https://github.com/skidding/react-cosmos/blob/42aaa56eafc738efe6a2daa5942cf8bcea6b8b94/packages/react-component-playground/src/components/component-playground.jsx#L132-L144) is what we send to the preview component. We don't need all of those keys, but we need `this.state.fixtureUnserializableProps` merged with `this.state.fixtureContents` to get the latest fixture contents (possibly updated by the user via the shinny editor 😉 ). We might need the component class as well in the future, in case proxies want to, say, display meta data of components on top of previewing them. 

Proposition: Create a method that returns the current fixture contents:
```js
return _.assign({
  component: this.constructor.getSelectedComponentClass(this.props)
}, this.state.fixtureUnserializableProps, this.state.fixtureContents);
```
Then use it as proxy props and inside `preview()` children function, where you add `key`, `ref` and omit `state`.

PR #204, opened 2016-09-14
Remove optional dep

skidding:
After seeing weird npm install warnings I realized putting `react-dom` in `optionalDependencies` would still yell at you when using older Reacts.

Then I re-read https://docs.npmjs.com/files/package.json#optionaldependencies and it seems `optionalDependencies` is more tied to `dependencies` than `peerDependencies`, which is were `react` & `react-dom` belong, so it seems like the cleanest thing to do is to omit it altogether. If you have React >=0.14 in your project & use React Cosmos you will have ReactDOM installed as well.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,8 +1,5 @@
 module.exports = {
   extends: [
     "airbnb"
-  ],
-  rules: {
-    "import/no-extraneous-dependencies": ["error", {"optionalDependencies": true}]
-  }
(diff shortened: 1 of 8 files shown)


skidding:
Please review @bogdanstefan @flaviusone @RadValentin

valentin-radulescu-hs:
LGTM

flaviusone:
LGTM

PR #206, opened 2016-09-15
Use UglifyJS compress to remove dead code

skidding:
@bogdanstefan check out if this gets rid of prod errors in your Redux setup.


skidding:
@bogdanstefan cool. I put it in your branch, your problem now! 🏃

PR #207, opened 2016-09-15
UI bugs

skidding:
- [x] Fix component state being lost when toggling editor
- [x] Fix full screen

@bogdanstefan please review. Any idea why did you remove the CSS bit from the latest commit (in 7fc47be8ae0eb34305233e748863277214a8e733)?

### DIFF
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -393,7 +393,10 @@ module.exports = React.createClass({
   componentDidUpdate(prevProps, prevState) {
     if (this.previewComponent && (
         this.constructor.didFixtureChange(prevProps, this.props) ||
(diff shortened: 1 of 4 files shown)


bogdanstefan:
Must've slipped by when I deleted the `top` and `bottom` parts not needed anymore because of `SplitPane`. Sorry.

IMHO the CSS needs refactoring, though. But maybe it should be correlated with dividing `component-playground.jsx`.

The code looks good, really nice that you added those tests, should prevent these things from happening in the future.

skidding:
> IMHO the CSS needs refactoring, though. But maybe it should be correlated with dividing component-playground.jsx into more files.

Definitely. Once everything is broken down into pieces it will become way nicer. The whole fixture craziness that happens should happen in a HoC, separated from the views.

PR #211, opened 2016-09-20
Integrate Flatris as Redux example

### LINKED ISSUE #210: Integrate Flatris as Redux example

skidding:
Fixes #210


skidding:
@bogdanjsx this is *it*. Beautiful JSX example and some changes to CP and the Redux proxy. Also, a million things to think about :). Please review. cc @flaviusone @RadValentin

`npm start` => [http://localhost:8989/?component=FlatrisGame&fixture=new-game-running](http://localhost:8989/?component=FlatrisGame&fixture=new-game-running)

example/src/components/Well.jsx
```diff
@@ -0,0 +1,84 @@
...
+   * A rectangular vertical shaft, where Tetriminos fall into during a Flatris
+   * game. The Well has configurable size and speed. Tetrimino pieces can be
+   * inserted inside the well and they will fall until they hit the bottom, and
+   * eventually fill it. Whenever the pieces form a straight horizontal line it
+   * will be cleared, emptying up space and allowing more pieces to enter
+   * afterwards.
+   */
+  shouldComponentUpdate(nextProps) {
+    return (
+      nextProps.grid !== this.props.grid ||
```
flaviusone: I think this will always be true because there are complex data structures that are compared. I think you should use some type of deep comparison like `_.isEqual()`.
Eg for the grid:
image

Am I wrong?

skidding: If the grids were cloned on every frame, yes. But I tried a more functional approach. The grid is only rebuilt [on start](https://github.com/skidding/react-cosmos/blob/763b9acfad69fb0e22a76b6a1080d6d4faf3bc71/example/src/reducer.js#L108) and [when a Tetrimino lands](https://github.com/skidding/react-cosmos/blob/763b9acfad69fb0e22a76b6a1080d6d4faf3bc71/example/src/reducer.js#L86). The rest of the time the exact same object is kept. What you're looking at is a reference check, which is way faster than comparing the values.

flaviusone: Nice

flaviusone:
LGTM

PR #214, opened 2016-10-14
webpack CLI

skidding:
Big one 💣 cc @flaviusone @RadValentin @bogdanjsx @catalinmiron 

New React Cosmos package: A CLI that makes integration as easy as possible for webpack users, using their existing webpack config and a new `cosmos.config.js` with just one required option (maybe zero in the future).

**Please start by reading the [Usage README section](https://github.com/skidding/react-cosmos/blob/2702d5babcd51169bd30d1645decd7e9a666ed70/README.md#usage), it serves as a deliverable for this PR as well.**

Other notable facts:
- Started testing new packages with Jest. It's pretty cool & definitely faster. ~~Will try to merge coverage reports between old and new suites in the future.~~ Done
- Changed the CP props a bit, separated `components` from `fixtures`.


packages/react-cosmos-webpack/src/build-module-paths.js
```diff
@@ -0,0 +1,60 @@
...
+
+const getFixtures = (componentsRootPath, componentCleanPath) => {
+  const fixturesRootPath = `${componentsRootPath}/__fixtures__/${componentCleanPath}`;
+
+  let fixtures = '{';
+  glob.sync(`${fixturesRootPath}/*.{js,json}`).forEach((fixturePath) => {
+    const nameMatcher = new RegExp(`^${fixturesRootPath}/(.+).(js|json)$`);
+    const matches = fixturePath.match(nameMatcher);
+    const name = matches[1];
+    const ext = matches[2];
```
flaviusone: Maybe [path.extname](https://nodejs.org/api/path.html#path_path_extname_path)  along with [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) can ease the work done here.

packages/react-cosmos-webpack/src/config.js
```diff
@@ -0,0 +1,17 @@
+export default function getConfig(userConfig) {
+  // The user config is not read in here because this function is used on both
+  // the server and client side and reading it is different per case.
+  // On the server side we read the path from the webpack CLI params while on
+  // the client receives the embedded path through webpack.DefinePlugin.
+  // These function just applies the defaults, which are "universal".
```
flaviusone: This* function

packages/react-cosmos-webpack/src/build-module-paths.js
skidding: `extname`  yes, but not `basename`, which a) contains the extension and b) is read from the last `/` to the right, whereas fixtures can be namespaced. E.g. `__fixtures__/Button/simple-states/default.js`, where `matches[1]` will be `simple-states/default` and basename will be `default.js`.

skidding:
cc @bogdanpetru you might want to take a look as well

examples/flatris/browserify/prepare-modules.js
```diff
@@ -0,0 +1,24 @@
...
+  const components = {};
+  Object.keys(modules).forEach(componentName => {
+    if (componentName.indexOf('_') !== 0) {
+      components[componentName] = () => modules[componentName];
+    }
+  });
+  return components;
+}
+
+export function prepareFixtures(modules, components) {
```
catalinmiron: From my POV, each component should have their own folder

```
MyComponent/
  __fixtures__
    state_1.js
    state_2.js
  __tests__
    state_1.spec.js
    state_2.spec.js
  index.js
  index.less
```

catalinmiron: This will also help you open sourcing the component with ease.

examples/flatris/webpack/webpack.config.js
```diff
@@ -0,0 +1,41 @@
+import path from 'path';
+import webpack from 'webpack';
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+
+// This config doesn't have entry and output set up because it's not meant to
+// work standalone. Cosmos adds an entry & output when extending this.
+module.exports = {
+  devtool: 'eval',
+  resolve: {
+    extensions: ['', '.js', '.jsx'],
```
catalinmiron: `root: path.resolve('./path_to_root')`

Reference: http://webpack.github.io/docs/configuration.html#resolve-modulesdirectories

This will remove the `../../../../../../constants/my_awesome_constant.js`

examples/flatris/browserify/prepare-modules.js
skidding: Yes! This has been on my mind for a while and has been [raised by many people](https://github.com/skidding/react-cosmos/issues/182). I plan to approach this as well in the future, maybe even included in 1.0, but the current structure will need to continue to work as well.

This is where it can be integrated in the webpack CLI: https://github.com/skidding/react-cosmos/blob/1432aeb487aaae938d767f2de3e0b71309c23170/packages/react-cosmos-webpack/src/build-module-paths.js#L9

E.g.
```js
// Current structure
const fixturesRootPathA = `${componentsRootPath}/__fixtures__/${componentCleanPath}`;
// Nested structure (like you proposed)
const fixturesRootPathB = `${componentsRootPath}/${componentCleanPath}/__fixtures__`;
```

I want to merge this PR before adding that functionality because I want to integrate Cosmos in a repo where it already works. If you want we can work together on adding nested fixture func in a subsequent PR.  👍

examples/flatris/webpack/webpack.config.js
skidding: > This will remove the ../../../../../../constants/my_awesome_constant.js

Where do you see paths like this in the Flatris example?

PR #218, opened 2016-10-18
Proxy API v2

### LINKED ISSUE #209: Proxy API revisited
Summary of my previous lunch with @bogdanjsx:

Main change: Proxies will no longer receive `children` to render, but just the fixture (and some other meta data) that they are to pass down to ReactElements created internally. Incidentally, this fixes [the Context problem in 0.13](https://github.com/skidding/react-cosmos/blob/da054785beea90a5932a1765548c10a5ab22c3ce/packages/react-cosmos-redux-proxy/src/index.jsx#L19-L22) by default.

Example:
```jsx
const ReactCosmosStateProxy = (props) => {
  const {
    // A linked list item with a pointer to the next element would be nice here (see below)
    nextProxy,
    // Can read meta data from it or even alter it with if necessary
    fixture,
    // Can be used to intercept when preview component is rendered (see below)
    previewRef,
  } = props;

  return React.createElement(nextProxy, Object.assign({}, props, {
    nextProxy: nextProxy.next(),
    previewRef: (previewComponent) => {
      if (fixture.state) {
        previewComponent.setState(fixture.state);
      }
      previewRef(previewComponent);
    }
  });
};
```

skidding:
Closes #209 

- ReduxProxy is now bound "two ways", meaning changes from the live instance are reflected in the fixture editor as well.
- StateProxy was extracted from CP, which is the first successful step in tearing down the CP *monster*.
- ReduxProxy now cancels StateProxy, so local state from inside connect() is no longer added to the fixture editor.
- Flatris is optimized :D – Planning to export static CP with Flatris and host it on github.io for anyone to be able to play with a live React Cosmos instance without any setup.
- Added another example to test local state injection & serialization.


skidding:
cc @flaviusone @RadValentin @bogdanjsx @catalinmiron 🙏 

redux-two-way

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -19,35 +28,68 @@ const ReduxProxyFactory = ({
       };
     }
 
+    componentWillMount() {
+      const {
+        store,
```
catalinmiron: 80 chars? this is hard to read.

catalinmiron: I know now:) `cmd+d` ftw. Lazy hax0r :)

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -19,35 +28,68 @@ const ReduxProxyFactory = ({
...
+        onPreviewRef,
+      } = this.props;
+
+      return React.createElement(nextProxy.value, { ...this.props,
+        nextProxy: nextProxy.next(),
+        // TODO: No longer omit when props will be read from fixture.props
+        // https://github.com/skidding/react-cosmos/issues/217
+        fixture: omit(fixture, fixtureKey),
+        onPreviewRef,
+        // Disable StateProxt when Redux state is available, otherwise the entire
```
catalinmiron: StateProxy

packages/react-cosmos/src/linked-list.js
```diff
@@ -0,0 +1,12 @@
+export default function createLinkedList(items) {
+  function advanceList(toIndex) {
+    return {
+      value: items[toIndex],
+      next: function getNextItem(nextIndex) {
+        return advanceList(nextIndex);
+      }.bind(null, toIndex + 1),
```
catalinmiron: maybe we'll use generators in a future release. wdyt?

packages/react-cosmos/src/proxies/StateProxy.jsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import omit from 'object.omit';
+import deepEqual from 'deep-equal';
```
catalinmiron: import {omit, isEqual} from 'lodash'

packages/react-cosmos/src/proxies/StateProxy.jsx
```diff
@@ -0,0 +1,122 @@
...
+    render() {
+      const {
+        props,
+        onPreviewRender,
+      } = this;
+      const {
+        nextProxy,
+        fixture,
+        disableLocalState,
+      } = this.props;
```
catalinmiron: `props` without `this` because you're taking the `props` above

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -19,35 +28,68 @@ const ReduxProxyFactory = ({
...
+        store,
+        onStoreChange,
+      } = this;
+      if (store) {
+        this.storeUnsubscribe = store.subscribe(onStoreChange);
+      }
+    }
+
+    componentWillUnmount() {
+      if (this.storeUnsubscribe) {
```
catalinmiron: typeof "function" and declare `this.storeUnsubscribe = null` in constructor?

skidding: Why? It's a private affair, no body is gonna set an `storeUnsubscribe` property of a different type...

packages/react-cosmos/src/linked-list.js
skidding: Tried using generators, but realised they work different. Tried my best to explain in this commit message: 3a02cfbe06d55ce39a6ec97bf1c9b3bf56dacd7e

packages/react-cosmos-redux-proxy/src/index.jsx
skidding: ¯\_(ツ)_/¯

packages/react-cosmos-redux-proxy/src/index.jsx
```diff
@@ -1,16 +1,25 @@
 import React from 'react';
-import _ from 'lodash';
+import omit from 'object.omit';
```
skidding: Is this new? I remember you had to use paths (e.g. `lodash/omit`) to benefit from this in the past.

flaviusone: I think that only webpack 2 has support for tree-shaking. On the other hand I guess you can use https://www.npmjs.com/package/lodash.omit.

flaviusone:
Looking good 👍

PR #229, opened 2016-11-21
Infer public path from webpackConfig.devServer.contentBase

skidding:


skidding:
@BTMPL I believe this is what you had in mind. I'll work on docs in a separate PR. Wdyt?

cc @xyx2011

BTMPL:
Looks good to me. I know this will only be worth for people using webpack-dev-server, but it's still something good to have IMO :)

PR #232, opened 2016-12-02
Isolate loader inside iframe

skidding:
Per Andrei's suggestion, I went ahead and separated the loader part from the playground.

Pros:
- Supports components with `display: fixed`.
- Complete JS/CSS encapsulation, playground doesn't affect user bundle.
- Smaller hmr overhead.
- Better isolation of concerns in codebase. Another step in making the Playground component approachable.

Cons:
- Drag and drop is sluggish because one of the panes is an iframe (will be fixed once ubervu-split-pane fork is up to date).
- Loaded component blinks for a second when opening/closing editor (because iframe dom is recreated) – state is preserved so it's just a minor detail.

@NiGhTTraX @flaviusone @RadValentin can you take a look, at least at the high level? Thanks!


flaviusone:
👍

valentin-radulescu-hs:
LGTM

PR #236, opened 2016-12-05
Make react-component-tree bundle smaller

### LINKED ISSUE #235: Make react-component-tree bundle smaller
When using [source-map-explorer](https://www.npmjs.com/package/source-map-explorer) on our project bundle the `react-component-tree` pulls the entire lodash and the result looks like this:
image

We can shrink its size by using individual lodash packages like [lodash.omit](https://www.npmjs.com/package/lodash.omit)

flavius-tirnacop-hs:
Closes #235

### DIFF
--- a/packages/react-component-tree/test/unit/load-child.js
+++ b/packages/react-component-tree/test/unit/load-child.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import _ from 'lodash';
+import cloneDeep from 'lodash.clonedeep';
 import { loadChild } from '../../src/load-child.js';
 
 const ReactComponent = {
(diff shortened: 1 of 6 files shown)


flavius-tirnacop-hs:
@skidding please review

I didn't bother to replace all the lodash calls for now because most of them are in `react-component-playground` that is used only for dev purposes. Also the `component-playground.jsx` calls quite a lot of different lodash methods so we may get quite a lot of imports. Maybe when this gets split into components we can refactor those calls.

packages/react-component-tree/test/unit/load-child.js
```diff
@@ -12,9 +12,9 @@ const StatelessComponent = {
 };
 
 describe('UNIT Load child', () => {
-  const FirstComponent = _.cloneDeep(ReactComponent);
-  const SecondComponent = _.cloneDeep(ReactComponent);
-  const ThirdComponent = _.cloneDeep(StatelessComponent);
+  const FirstComponent = cloneDeep(ReactComponent);
+  const SecondComponent = cloneDeep(ReactComponent);
+  const ThirdComponent = cloneDeep(StatelessComponent);
```
skidding: This is odd, what must've I been thinking. Can you duplicate `ReactComponent` and remove these `cloneDeep` calls along with the peerDependency?

flavius-tirnacop-hs:
@skidding addressed comments

PR #241, opened 2016-12-23
Removed deprecated lodash.isarray

raulanatol:
npm WARN deprecated lodash.isarray@4.0.0: This package is deprecated. Use Array.isArray.


skidding:
Seems like the build failed due to a Lerna version mismatch. Can you freeze the Lerna version in package.json (by removing the `^` at https://github.com/react-cosmos/react-cosmos/blob/master/package.json#L65) to fix it and prevent future failures like this?

I would but am currently on the road:
img_20161223_154758

Cheers.

skidding:
Thanks! 🎄

PR #243, opened 2016-12-29
Reload page when hot reloading fails

skidding:


skidding:
@ovidiubute please review

This auto reloads the playground when a component or fixture is added or edited (if the patch can't be applied via hot reloading. i.e. all fixtures and stateless components).

Issue #217, opened 2016-10-18
Redesign fixture to keep props separate from meta fields


skidding:
### Problem

Props are now root level fixture attributes. This means attrs like `state` or `reduxState` or other flags and modifiers that can power Proxies are "magic fields" next to props and you can't have props with those names. This wasn't a problem before Proxies because `state` was the only magic attr and it wasn't a common prop name anyway. The new Proxies might be interested in all sorts of attrs like `width`, `height` or `backgroundColor`, which would easily be in conflict with actual component props. 
### Solution

Have dedicated nesting for props inside fixture.

Current fixture format:

``` js
{
  fooProp: 'bar',
  state: {
    fooState: 'bar'
  },
  // meta data for Proxies. e.g. reduxState
}
```

Next fixture format:

``` js
{
  props: {
    fooProp: 'bar',
  },
  state: {
    fooState: 'bar'
  },
  // meta data for Proxies. e.g. reduxState
}
```

Also, remove component from fixture and send it as a separate prop to Proxies.

This is a small change but spans across many packages:
- [No longer attach component class to fixture in ComponentPlayground](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-component-playground/src/components/component-playground.jsx#L532)
- [Send component as separate prop to proxies in ComponentPlayground](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-component-playground/src/components/component-playground.jsx#L218-L225)
- [Use component proxy instead of fixture.component in PreviewLoader](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-cosmos/src/proxies/PreviewLoader.jsx#L11)
- [Use fixture.props instead mutated fixture in PreviewLoader](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-cosmos/src/proxies/PreviewLoader.jsx#L12-L14)
- [No longer mutate fixture in Redux proxy](https://github.com/react-cosmos/react-cosmos/blob/c3e023aec675079dfa952d5e391d521d745650c3/packages/react-cosmos-redux-proxy/src/index.jsx#L67-L69)
[...]

PR #248, opened 2016-12-31
Upgrade to AirBnB ESLint v13

### LINKED ISSUE #246: Fix eslint warning
### Problem
- Console warning `The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.`

### Solution
- It looks like we can upgrade `eslint-config-airbnb` to version `12` to fix this. See issue [#978](https://github.com/airbnb/javascript/issues/978)

ovidiubute:
Closes #246 

Had to do more work than I thought, as version 13 has some changes to the ESLint rules.


ovidiubute:
@skidding please review

.eslintrc.js
```diff
@@ -5,7 +5,16 @@ module.exports = {
...
     // React Cosmos is designed to work with React >= 0.13 and stateless
     // functions were invented in 0.14
-    "react/prefer-stateless-function": 0
+    "react/prefer-stateless-function": 0,
+    // Dynamic require is needed in Playground
+    "import/no-dynamic-require": 0,
+    // Object and Array are forbidden by default too
+    "react/forbid-prop-types": [2, { "forbid": [ "any" ] }],
+    // Passing children as prop is needed in React Cosmos
+    "react/no-children-prop": 0,
```
skidding: I'd add a TODO comment to fix this, with a link to your comment: https://github.com/react-cosmos/react-cosmos/issues/217#issuecomment-269708146

.eslintrc.js
```diff
@@ -5,7 +5,16 @@ module.exports = {
...
+    // Defining unused props is required in Proxies
+    "react/no-unused-prop-types": 0,
+    "jsx-a11y/anchor-has-content": 0
```
skidding: How about putting this at the beginning of the `component-playground.jsx` file because it only occurs there?

examples/flatris/browserify/prepare-modules.js
```diff
@@ -2,7 +2,7 @@
 
 const prepareComponents = (modules) => {
   const components = {};
-  Object.keys(modules).forEach(componentName => {
+  Object.keys(modules).forEach((componentName) => {
```
skidding: This is odd. In some case you have to remove the parenthesis and in others (like here) you have to add it. Does it have smth to do with the curly brackets?

packages/react-cosmos-utils/src/unserializable-parts.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-bitwise */
```
skidding: This is an error! Please replace `|` with `||`. 😥

packages/react-cosmos/src/components/Loader.jsx
```diff
@@ -1,3 +1,5 @@
+/* global parent */
```
skidding: Is this needed? ESLint doesn't render any errors for me when I remove it (because [this](https://github.com/react-cosmos/react-cosmos/blob/7a90876a11ea257a122c2d25295c482abef2e30c/packages/react-cosmos/src/.eslintrc.json#L3))

ovidiubute:
Valid points. Happy new year! 🎉

skidding:
Happy new year! 🎆

examples/flatris/browserify/prepare-modules.js
ovidiubute: The rule is:
- If you write a function that takes one argument, and its body does not have curly braces, then do not apply parenthesis to the argument. If its body does have curly braces, then the argument must be surrounded by parenthesis.

ovidiubute:
@skidding addressed review

PR #250, opened 2017-01-01
Pass fixture.children as 3rd argument to createElement

### LINKED ISSUE #249: [PropsProxy] Pass children as 3rd argument to React.createElement

ovidiubute:
Closes #249

### DIFF
--- a/packages/react-cosmos/src/components/proxies/PropsProxy.jsx
+++ b/packages/react-cosmos/src/components/proxies/PropsProxy.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import omit from 'lodash.omit';
 
 export default class PropsProxy extends React.Component {
   /**
@@ -10,15 +11,11 @@ export default class PropsProxy extends React.Component {
       fixture,
       onComponentRef,
     } = this.props;
-    let children;
-    if (Array.isArray(fixture.children)) {
-      children = [].slice.call(fixture.children);
-    }
 
     return React.createElement(component, {
-      ...fixture,
+      ...omit(fixture, 'children'),
       ref: onComponentRef,
-    }, children);
+    }, fixture.children);
   }
 }
 
--- a/packages/react-cosmos/src/components/proxies/PropsProxy.jsx
+++ b/packages/react-cosmos/src/components/proxies/PropsProxy.jsx
@@ -10,11 +10,15 @@ export default class PropsProxy extends React.Component {
       fixture,
       onComponentRef,
     } = this.props;
+    let children;
+    if (Array.isArray(fixture.children)) {
+      children = [].slice.call(fixture.children);
+    }
 
     return React.createElement(component, {
       ...fixture,
       ref: onComponentRef,
-    });
+    }, children);
   }
 }
 
--- a/packages/react-cosmos/src/components/proxies/__tests__/PropsProxy.jsx
+++ b/packages/react-cosmos/src/components/proxies/__tests__/PropsProxy.jsx
@@ -5,12 +5,16 @@ import PropsProxy from '../PropsProxy';
 const Component = () => {};
 const fixture = {
   foo: 'bar',
+  children: [
+    <p key="1">Child 1</p>,
+  ],
 };
 const onComponentRef = jest.fn();
 
 let wrapper;
 let childWrapper;
 let childProps;
+let childChildren;
 
 beforeAll(() => {
   wrapper = shallow(
@@ -22,6 +26,7 @@ beforeAll(() => {
   );
   childWrapper = wrapper.at(0);
   childProps = childWrapper.props();
+  childChildren = childWrapper.children();
 });
 
 test('renders component', () => {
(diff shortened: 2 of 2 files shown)


ovidiubute:
@skidding please review

packages/react-cosmos/src/components/proxies/PropsProxy.jsx
```diff
@@ -10,11 +10,15 @@ export default class PropsProxy extends React.Component {
       fixture,
       onComponentRef,
     } = this.props;
+    let children;
+    if (Array.isArray(fixture.children)) {
+      children = [].slice.call(fixture.children);
```
skidding: No need to clone `children`, because the fixture is never mutated. If we went down this road we'd have to clone props and other stuff as well. Let's just send `fixture.children` directly to `React.createElement`.

packages/react-cosmos/src/components/proxies/PropsProxy.jsx
```diff
@@ -10,11 +10,15 @@ export default class PropsProxy extends React.Component {
...
+    }
 
     return React.createElement(component, {
       ...fixture,
       ref: onComponentRef,
-    });
+    }, children);
```
skidding: This needs a [test](https://github.com/react-cosmos/react-cosmos/blob/b3be0680336724bce8830dae2f367df260394364/packages/react-cosmos/src/components/proxies/__tests__/PropsProxy.jsx). Check out how [props are tested](https://github.com/react-cosmos/react-cosmos/blob/b3be0680336724bce8830dae2f367df260394364/packages/react-cosmos/src/components/proxies/__tests__/PropsProxy.jsx#L31-L33). The [children()](http://airbnb.io/enzyme/docs/api/ShallowWrapper/children.html) method from the Enzyme wrapper should do the trick.

packages/react-cosmos/src/components/proxies/PropsProxy.jsx
```diff
@@ -10,11 +10,15 @@ export default class PropsProxy extends React.Component {
       fixture,
       onComponentRef,
     } = this.props;
+    let children;
+    if (Array.isArray(fixture.children)) {
+      children = [].slice.call(fixture.children);
+    }
 
     return React.createElement(component, {
       ...fixture,
```
skidding: We need to omit the `children` field from the fixture, otherwise it continues to be sent as a prop as well as the 3rd arg from `createElement`.  See [how StateProxy omits `state` field](https://github.com/react-cosmos/react-cosmos/blob/b3be0680336724bce8830dae2f367df260394364/packages/react-cosmos/src/components/proxies/StateProxy.jsx#L100).

You also need to add a test that checks that `fixture.children` is no longer sent as a prop.

ovidiubute: The reason this code exists is because `createElement` takes a variable number of arguments, but `children` is an `Array`. Ideally you'd destructure it with `...children`, but this fails if the fixture does not have the `children` field (you cannot destructure `undefined`).

skidding: I think the destructuring of children is optional: https://jsbin.com/yikulexefe/edit?js,output – which means the 3rd argument from `createElement` can be a list (or single element, or `undefined`).

With this in mind, do you see a problem with this?
```js
    React.createElement(component, {
      ...omit(fixture, 'children'),
      ref: onComponentRef,
    }, fixture.children);
```

ovidiubute: Yeah, you're right, destructuring is optional. 🐧

ovidiubute: While writing the test I realized that `children` in `React-Cosmos` is not the same as children in React. Those children need to be React Elements, not objects with props. Back to the drawing board. 📚

skidding: > While writing the test I realized that children in React-Cosmos is not the same as children in React.

Not really. `fixture.children` are React Elements, not object with props. Not a very used pattern at HS, but useful when you create fixtures for wrapper components like modal windows.

I think you're thinking of the [loadChild](https://github.com/react-cosmos/react-cosmos/blob/b3be0680336724bce8830dae2f367df260394364/packages/react-component-tree/src/load-child.js) mixin where you define elements at the beginning of the component and then insert them in the render function using `loadChild`. That's just an optional utility, not used in any React Cosmos example.

ovidiubute: My problem is due to a lack of experience using `enzyme`. I tried to modify the fixture to the following:

```
const fixture = {
  foo: 'bar',
  children: [{
    c1: {
      baz: 42,
    },
  }],
};
```

But I get the following error:

Invariant Violation: Objects are not valid as a React child (found: object with keys {c1}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons.

That's why I assumed I was doing something wrong in `createElement`, but it's most likely to do with the shallow rendering of `enzyme`.

skidding: That's not a valid fixture.

This is the type of fixture relevant to this PR: Mocking React children.
```jsx
const fixture = {
  foo: 'bar',
  children: [
     <p>I am wrapped inside a parent component</p>
  ],
};
```
You may be thinking about this: Mocking the state of child components (that have string *refs*).
```js
const fixture = {
  foo: 'bar',
  state: {
    children: {
      c1: {
        baz: 42,
      },
    },
  },
};
```
We can use a different medium like Slack if you have further questions.

ovidiubute:
@skidding addressed review

PR #254, opened 2017-01-03
Improve npm targets

### LINKED ISSUE #247: Add generic build npm task
### Problem
- The list of npm targets is getting pretty big, and is unsustainable now that proxies are being added to `react-cosmos`. 
- The `build` target is error-prone since you need to call all of the above targets one by one

### Solution
- Have a generic `build:package` target that accepts CLI arguments and calls a separate script
- The script will call what `build-webpack` is calling, but instead of taking the argument from the environment, it will use the CLI argument
- The `build` target could also be a script that lists the packages and calls the `build:package` tasks one by one, eliminating the need to maintain it

ovidiubute:
Closes #247 

- To build all packages:

`npm run build`

- To build a single package:

`npm run build react-cosmos-utils`

- To build & watch:

`npm run build react-component-playground -- --watch`

- Trying to build a non-existing package will print an error message.
- Removed extra npm targets, and cleaned up the `build` target.
- The only restriction is the package name must start with `react-`.
- No unit-tests 😢


ovidiubute:
@skidding please review

scripts/build-package.js
```diff
@@ -0,0 +1,36 @@
...
+      'build-webpack';
+  const stdout = npmRun.execSync(`PACKAGE=${packageName} npm run ${npmTask}`, {
+    cwd: __dirname,
+  });
+
+  /* eslint-disable no-console */
+  console.log(stdout.toString('utf-8'));
+}
+
+glob('./packages/*react-*', null, (err, files) => {
```
skidding: Shouldn't this be `./packages/react-*`to only match packages that *start* with `react`?

scripts/build-package.js
```diff
@@ -0,0 +1,36 @@
...
+  const allPackages = files.map(f => path.basename(f));
+  const argv = yargs
+      .usage('Usage: $0 <package>')
+      .help()
+      .argv;
+  const targetPackage = argv._[0];
+
+  if (!targetPackage) {
+    allPackages.forEach(pkg => (runBuildTask(pkg)));
```
skidding: This needs to be a bit smarter. I.e. packages bundled via webpack need to be built after other ones they depend on.

image


Suggestion:
1. Keep a list of packages bundled with webpack (just CP for now)
2. Run packages that aren't included in that list before those that are
3. Bonus: Use same list to check which task to run (line 11)

Accepted limitation: This might still fail when more builds that bundle with webpack depend on each other. There's only one webpack build now (CP) and there are no plans for any other in the near/mid future.

ovidiu-bute-hs: A config file with the dependencies would be useful for this. Example:

`
module.exports = {
  "react-component-playground": {
    "depenendents": [
        "react-cosmos-utils",
        "react-component-tree",
        ...
      ] 
  }
}
`

skidding: > A config file with the dependencies would be useful for this

At scale, I agree, but here are few things to consider:
1. The dependency graph doesn't matter between packages that are just Babel-compiled (all but the playground), because no bundling occurs and the dependencies are resolved later in the user's codebase
2. There is only one package that bundles and there aren't any plans to add another (the reason why CP bundles with webpack is its use of CSS)
3. This dep graph would still not be needed until we'd have a webpack-bundled package depend on another webpack-bundled package, which we can assume will never be the case

That said, I think we should go for the easy solution, even if that means another `if` for `react-component-playground` and dropping the list idea altogether. What would be useful is a comment leaving breadcrumbs for future generations, stating that this solution doesn't work when other packages start depending on each other to build (only one does now, see point 1.)

ovidiubute:
@skidding fixed everything but the `watch`. It doesn't work anymore so I'll have to bake that into the new script too.

ovidiubute:
To watch for changes, `npm run build:react-component-playground -- --watch`. Also switched to Node `spawn` which provides the benefit of building all the packages in parallel (except CP which is built at the end). @skidding please review.

scripts/build-package.js
```diff
@@ -0,0 +1,76 @@
...
+  if (!targetPackage || targetPackage === 'react-component-playground') {
+    // NOTE: The Playground needs to be built after everything else
+    runBuildPlayground(allPackagesNames, applyWatch);
+  } else if (allPackagesNames.includes(targetPackage)) {
+    runBuildTask({
+      packageName: targetPackage,
+      watch: applyWatch,
+    });
+  } else {
+    throw new Error(`Invalid package! Can only build the following packages: ${allPackagesNames}`);
```
skidding: This creates a lot of noise:
image

Let's use `console.error`, it's much nicer:

image

And to make it even nicer, let's do something like `${[''].concat(allPackagesNames).join('\n -')}`:

image

I'd also replace "Can only build the following packages:" with "These are the existing packages:"

Wdyt?

package.json
```diff
@@ -11,16 +11,8 @@
...
-    "build:cosmos": "PACKAGE=react-cosmos npm run build-babel",
-    "build:utils": "PACKAGE=react-cosmos-utils npm run build-babel",
-    "build:webpack": "PACKAGE=react-cosmos-webpack npm run build-babel",
-    "build:dom-polyfill": "PACKAGE=react-dom-polyfill npm run build-babel",
-    "build:querystring-router": "PACKAGE=react-querystring-router npm run build-babel",
-    "build:redux-proxy": "PACKAGE=react-cosmos-redux-proxy npm run build-babel",
-    "build:context-proxy": "PACKAGE=react-cosmos-context-proxy npm run build-babel",
-    "build": "rm -rf ./packages/*/lib && npm run build:utils && npm run build:playground && npm run build:component-tree && npm run build:cosmos && npm run build:webpack && npm run build:dom-polyfill && npm run build:querystring-router && npm run build:redux-proxy && npm run build:context-proxy",
+    "build:package": "node ./scripts/build-package.js",
+    "build": "rm -rf ./packages/*/lib && npm run build:package",
```
skidding: Having these two scripts is confusing, it took me a bit to understand what happens. E.g.

```bash
# Both build only the react-cosmos-utils script, except in the first case the watch flag is ignored
npm run build react-cosmos-utils -- --watch
npm run build:package react-cosmos-utils -- --watch
```

Let's get rid of `build:package`. You can use rimraf (see [example](https://github.com/react-cosmos/react-cosmos/blob/abb3dedfaa07e72a6c2b95a9e898d11d579579a2/scripts/prune-react-copies.js#L9-L11)) to run wipe out `./packages/*/lib` when `targetPackage` is missing.

scripts/build-package.js
```diff
@@ -0,0 +1,76 @@
...
+  });
+}
+
+const targetPackage = argv._[0];
+const applyWatch = Boolean(argv.watch);
+
+glob('./packages/react-*', null, (err, files) => {
+  const allPackagesNames = files.map(f => path.basename(f));
+
+  if (!targetPackage || targetPackage === 'react-component-playground') {
```
skidding: You shouldn't compile all other packages every time you bundle CP.

You need to ensure order when we build all of them `npm run build`, but other than that you don't want to re-build other packages when you run `npm run build react-component-playground` when working exclusively on the playground. Not to mention that some packages are not deps of the playground such as `react-cosmos` and `react-cosmos-webpack`.

Having three separate cases seems like the clean solution. Wdyt?

scripts/build-package.js
ovidiu-bute-hs: Sure!

package.json
ovidiu-bute-hs: Passing arguments to `npm run build` was not intended. You could do that to the old target and pass `watch` to the last package `build` target. It's interesting how it turned out though, running just `build` would definitely be easier to understand 👍

ovidiubute:
I really should `nvm install 5` 😄

ovidiubute:
@skidding please review

scripts/build-package.js
```diff
@@ -0,0 +1,101 @@
...
+  // Build all packages except CP
+  packageNames
+      .filter(pkg => pkg !== 'react-component-playground')
+      .forEach(packageName => (
+          runBuildTask({
+            packageName,
+          })
+        ));
+
+  runBuildPlaygroundTask();
```
skidding: Since all build cmds are now async and start at the same time, how can we know that the other packages are built when the playground starts bundling?

skidding: Maybe use [child-process-promise](https://github.com/patrick-steele-idem/child-process-promise), have `runBuildTask` return a promise, and then

```js
Promise.all(
  packageNames
    .filter(pkg => pkg !== 'react-component-playground')
    .map(packageName => runBuildTask({ packageName })
).then(runBuildPlaygroundTask);
```

scripts/build-package.js
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Run the build for all packages, the CP requires all other components
+ * to be built first.
+ * @param packageNames List of package names
+ */
+function runBuildAllTask(packageNames) {
+  // Cleanup
+  glob.sync('./packages/*/lib').forEach((packageLibPath) => {
+    rimraf.sync(packageLibPath);
+    console.log('WARNING: Removed lib directory for', packageLibPath);
```
skidding: This is normal behavior so let's remove `WARNING: ` prefix to decrease CLI anxiety ✌️

ovidiubute: I'd change it to `INFO:` just to be clear on what's happening.

skidding: Sure

ovidiubute:
raw

skidding:
Congrats, @ovidiubute! Tricky one but you made it 🏁

ovidiubute:
Team effort, @skidding 😄 

59a

PR #260, opened 2017-01-21
Integrate webpack HMR

### LINKED ISSUE #255: Hot reload all components and fixtures
## Problem

Cosmos has hot reloading, but it uses [react-transform-hmr](https://github.com/gaearon/react-transform-hmr/blob/master/README.md) which is deprecated and lacks in features (Support for functional components is one).

## Solution

Follow [Hot Module Replacement - React](https://webpack.js.org/guides/hmr-react/) webpack tutorial.

Warning: The tutorial revolves around webpack@2 and RC uses webpack@1 for now, so only extract the HMR-related bits.

> This issue was initially about integrating React Hot Loader 3, but after learning more about the HMR API I concluded that it is *good enough* for Cosmos. RHL goes to great lengths to preserve state between hot patches, but CP already preserves state and ensures deterministic re-renders. Preserving DOM state (or avoiding DOM trashing) is not of interest for Cosmos. On the other hand, not adding RHL (at least for for now) has two benefits:
>
> - Keeps things simpler
> - Cosmos continues to work with 0.13 (RHL needs 0.14 or above)
>
> What's more, the HMR integration alone also includes **hot reloading for fixtures**!

  ```js
  // Hot Module Replacement API example
  if (module.hot) {
    module.hot.accept('./components/App', () => {
      const NewApp = require('./components/App').default
      render(NewApp)
    });
  }
  ```

skidding:
Fixes #255 (Hot reloads all components and fixtures)

### DIFF
--- a/examples/context/webpack.config.js
+++ b/examples/context/webpack.config.js
@@ -23,10 +23,6 @@ module.exports = {
       test: /\.jsx?$/,
       loader: 'babel',
       exclude: /node_modules/,
-      query: {
-        // 'react-hmre' only works with React >=0.14
-        presets: ['es2015', 'react', 'react-hmre'],
-      },
     }],
   },
   plugins: [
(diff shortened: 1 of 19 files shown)


skidding:
cc @flavius-tirnacop-hs @ovidiubute @catalinmiron @RadValentin hot reloading fixtures!

lnlaifqdenmxw

Issue #234, opened 2016-12-02
Document requirement for babel-plugin-transform-es2015-modules-commonjs


gajus:
It is not clear that `babel-plugin-transform-es2015-modules-commonjs` is required.

Note that webpack@2 does not need the latter module. Therefore, most webpack@2 users will not have it.

skidding:
Are you referring to the fact that webpack@2 users will have es2015 modules disabled in their `.babelrc` (and let webpack handle es2015 modules), like this?
```js
{
  "presets": [
    ["es2015", { "modules": false }]
  ]
}
```
The react-cosmos modules are already published as es5, so the only problem I can anticipate is disabling Babel modules and using es2015 `import` inside `cosmos.config.js`, which isn't processed through webpack. Is this the problem you identified or something different?

gajus:
> Are you referring to the fact that webpack@2 users will have es2015 modules disabled in their .babelrc (and let webpack handle es2015 modules), like this?

Yes.

> The react-cosmos modules are already published as es5, so the only problem I can anticipate is disabling Babel modules and using es2015 import inside cosmos.config.js, which isn't processed through webpack. Is this the problem you identified or something different?

I have a component, e.g. `src/components/TestComponent.js`

```js
import React from 'react';

export () => {
  return <div>Hello, World!</div>;
};
```

Suppose my `.babelrc` is:

```json
{
  "presets": [
    ["es2015", { "modules": false }]
  ]
}
```

With this setup, if I use `react-cosmos`, I am getting an error on `import React from 'react';` in the `src/components/TestComponent.js`.

I am assuming this is because `react-cosmos` directly depends on `webpack@1`?

https://github.com/react-cosmos/react-cosmos/blob/81ad59e73911faf8117ba3f02fb6786b316e2963/package.json#L74

skidding:
> I am assuming this is because react-cosmos directly depends on webpack@1?

Yes, `react-cosmos-webpack` more specifically, has been built for and tested with `webpack@1`. We could make this explicit in the [Requirements](https://github.com/react-cosmos/react-cosmos/blob/81ad59e73911faf8117ba3f02fb6786b316e2963/README.md#requirements), but at the same time it would be great to find a way to make Cosmos work with both webpack 1 and 2. Publishing separate versions of `react-cosmos-webpack` for each webpack version is undesirable. 

[...]

PR #262, opened 2017-01-22
Support webpack 2 (as well as webpack 1)

### LINKED ISSUE #261: Support webpack 1 and 2, use peerDependencies
Fix #234

skidding:
Fixes #261

### DIFF
--- a/examples/context/webpack.config.js
+++ b/examples/context/webpack.config.js
@@ -1,32 +1,23 @@
-/* eslint-disable import/extensions */
-import path from 'path';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import webpack from 'webpack';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 
 // This config doesn't have entry and output set up because it's not meant to
-// work standalone. Cosmos adds an entry & output when extending this.
+// work standalone. react-cosmos-webpack adds an entry & output when extending this.
 module.exports = {
   devtool: 'eval',
   resolve: {
-    extensions: ['', '.js', '.jsx'],
-  },
-  resolveLoader: {
-    // This can be removed in a real life scenario. react-cosmos-webpack is a
-    // sibling directory in this setup, whereas a regular npm module is nested
-    // within the app. In the latter case, node modules installed in the host
-    // app are found by react-cosmos-webpack due to Node's bubble up require
-    // strategy.
-    root: path.join(__dirname, 'node_modules'),
+    extensions: ['.js', '.jsx'],
   },
   module: {
     loaders: [{
       test: /\.jsx?$/,
-      loader: 'babel',
+      loader: 'babel-loader',
       exclude: /node_modules/,
     }],
   },
   plugins: [
-    new webpack.NoErrorsPlugin(),
+    new webpack.NoEmitOnErrorsPlugin(),
     new HtmlWebpackPlugin({
       title: 'React Cosmos',
     }),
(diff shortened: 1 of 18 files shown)


skidding:
In this PR:
- Moved webpack to peerDependencies of `react-cosmos-webpack`, allowing users to use the webpack version from their project
- Upgraded entire repo and examples to work with both webpack 1 and 2
- Added [mention](https://github.com/react-cosmos/react-cosmos/tree/c2e6e9c9a6b92aa24e2c22f347d5102bbd3ec1e5#using-webpack-2) about disabling Babel modules with webpack 2 (suggested in #234)

Please take a look @gajus @ovidiubute

packages/react-component-playground/webpack.config.js
```diff
@@ -35,22 +35,22 @@ module.exports = {
     'react-dom': 'react-dom',
   },
   resolve: {
-    extensions: ['', '.js', '.jsx'],
+    extensions: ['.js', '.jsx'],
   },
   module: {
```
gajus: `webpack` v2 wraps this in [rules](https://webpack.js.org/configuration/module/#module-rules) configuration.

skidding: It also continues to support `loaders`, which is great because it allows Cosmos to build webpack configs that work universally without version checking.

That being said, I can start using `rules` here because it's part of the build process, not lib code (same with examples).

examples/context/webpack.config.js
```diff
@@ -1,32 +1,23 @@
...
     loaders: [{
       test: /\.jsx?$/,
-      loader: 'babel',
+      loader: 'babel-loader',
       exclude: /node_modules/,
     }],
   },
   plugins: [
-    new webpack.NoErrorsPlugin(),
+    new webpack.NoEmitOnErrorsPlugin(),
```
ovidiubute: @skidding This makes Flatris impossible to build without webpack2. The peerDependency in `react-cosmos-webpack` specifies a version between 1.13 and 3, but that is no longer the case. I tried to build it locally and couldn't do it with webpack 1.14.

ovidiubute: I realise this is the `context` config file, but you get my point 😄

skidding:
Right. I forgot examples work with the webpack in the root dir. We can just
remove the plugin from the examples for now. Feel free to submit a PR if
that's the only thing stopping the examples from working on webpack 1.x.

examples/context/webpack.config.js
skidding: > I tried to build it locally and couldn't do it with webpack 1.14.

@ovidiubute How? Flatris (all examples) use the webpack copy installed in the root dir, which is [pinned to v2](https://github.com/react-cosmos/react-cosmos/blob/da85709af02f4ddf5b8d9908c52678f802a9f214/package.json#L66).

ovidiubute: I guess I didn't delete the root `node_modules` before I started working and still had webpack version 1.14 installed. If you perform a clean install it should be fine.

skidding: We could do for webpack versioning what [we did for React](https://github.com/react-cosmos/react-cosmos/blob/da85709af02f4ddf5b8d9908c52678f802a9f214/package.json#L7-L10), but I wouldn't go that route tbh. I think it's safe to keep webpack 2 in the repo and examples as long as we maintain webpack 1 compatibility.

PR #274, opened 2017-01-31
Isolated config

### LINKED ISSUE #272: Isolate and no longer import config in client bundle
#### What's up?

Three reasons for this:

- We need a modular way to read the cosmos config for the Jest snapshot integration
- So far the config file has been "universal" (loaded on the server as well as client side). It seemed like a good idea, but this restricts us from doing any fs operations in the config (lest the client bundle explode). Thus we should select the settings that are client facing, serialize and inject them into the client bundle using webpack DefinePlugin. `proxies` is the only client-facing setting that can't be serialized, but we can go around this by moving the Proxy code outside of configs and only keeping proxy file paths in config/proxies
- Support package paths in `globalImports` (e.g. `babel-polyfill`)

#### Mkay, tell me more

Here's how proxies will be configured:

```js
// .cosmos.config.js
proxies: [
  'react-cosmos-no-options-proxy',
  './custom-proxy.js'
]
```
```js
// .custom-proxy.js
import configurableProxy from 'react-cosmos-conf-proxy';

export default function() {
  return configurableProxy({
    myOption: 3,
    myOtherOption: require('./foobar.js')
  });
}
```

Moreover, there will be a new `react-cosmos-config` package, used like this:

```js
import getCosmosConfig from 'react-cosmos-config';

const cosmosConfig = getCosmosConfig({
  // optional
  configPath: './custom/cosmos/config/path.js'
});
```

This package will:

- [x] Load the config path, default to `CWD/cosmos.config.js`
- [x] Resolve `componentPaths` and `fixturePaths` to absolute paths
- [x] Resolve `globalImports` to absolute paths (or package paths, e.g. `babel-polyfill`)
- [x] Resolve `webpackConfigPath` to absolute path (default to `webpack.config.js` first)
- [x] Resolve `publicPath`
- [x] Provide other defaults for the rest of the options (e.g. enable `hot` and `hmrPlugin`)

skidding:
Fixed #272 

Note: Also made hot reloading enabled by default.

TODO:
- [x] Add Proxies section to README


README.md
```diff
@@ -194,6 +179,52 @@ From [the new webpack docs](https://webpack.js.org/guides/migrating/#mixing-es20
...
+
+#### react-cosmos-redux-proxy
+
+Most components in a Redux app depend on Redux state–either they're a *container* or one of their descendants is. This proxy creates the store context required for any component you load, just like [Provider](http://redux.js.org/docs/basics/UsageWithReact.html#passing-the-store) does for your root component. Writing Redux fixtures almost feels too easy. Because Redux state is global, once you have one state mock you can render any component you want!
+
+```js
+// redux-proxy.js
+import createReduxProxy from 'react-cosmos-redux-proxy';
+
+export function () {
```
ovidiubute: Why not go all the way and use a fat arrow here? Everything else is ES6.

skidding: Because you can't export a fat arrow function directly. Or can you?

ovidiubute: Sure you can, it'll just get transpiled to `function ()`.

ovidiubute: Why aren't you exporting `default`, now that I think of it?

skidding: Ouch, by mistake. The examples are [fine](https://github.com/react-cosmos/react-cosmos/blob/d12a20eacdfe44b37f0478ab7e978fdacf501809/examples/context/context-proxy.js#L4).

What I was saying is that you can do
```js
export const createProxy = () => {
```
but not
```js
export default const createProxy = () => {
```

What am I missing?

ovidiubute: You can do 

```
export default () => {
```

PR #275, opened 2017-01-31
Add user webpack config validation

nik-kor:
I spent an hour getting to know why my build wasn't working. It failed with an error
```
get-contexts.js:14 Uncaught ReferenceError: COMPONENT_CONTEXTS is not defined
```

So the problem was with webpack configuration - I had to use `module.loaders` in my webpack.config. I know that there is a note about it in readme. But i think it's not obvious.

### DIFF
--- a/packages/react-cosmos-webpack/src/__tests__/webpack-config.js
+++ b/packages/react-cosmos-webpack/src/__tests__/webpack-config.js
@@ -59,30 +59,6 @@ describe('without hmr', () => {
     webpackConfig = getWebpackConfig(userWebpackConfig, cosmosConfigPath);
   });
 
-  test('choose `loaders` or `rules` from user config', () => {
-    const userRule = { foo: 'bar' };
-
-    webpackConfig = getWebpackConfig(
-      {
-        module: {
-          rules: [userRule]
-        }
-      },
-      cosmosConfigPath
-    );
-    expect(webpackConfig.module.rules[0]).toEqual(userRule);
-
-    webpackConfig = getWebpackConfig(
-      {
-        module: {
-          loaders: [userRule]
-        }
-      },
-      cosmosConfigPath
-    );
-    expect(webpackConfig.module.loaders[0]).toEqual(userRule);
-  });
-
   test('keeps user rules', () => {
     expect(webpackConfig.rules).toBe(userWebpackConfig.rules);
   });
@@ -198,5 +174,50 @@ describe('with absolute paths', () => {
     mockCosmosConfig.globalImports.forEach(globalImport => {
       expect(webpackConfig.entry).toContain(globalImport);
     });
+  });
+});
+
+describe('webpack config basic', () => {
+  test('choose `loaders` or `rules` from user config', () => {
+    const userRule = { foo: 'bar' };
+
+    webpackConfig = getWebpackConfig(
+      {
+        module: {
+          rules: [userRule]
+        }
+      },
+      cosmosConfigPath
+    );
+    expect(webpackConfig.module.rules[0]).toEqual(userRule);
+
+    webpackConfig = getWebpackConfig(
+      {
+        module: {
+          loaders: [userRule]
+        }
+      },
+      cosmosConfigPath
+    );
+    expect(webpackConfig.module.loaders[0]).toEqual(userRule);
+  });
+
+  test('passing user data to module config', () => {
+    const additionalOption = {
+      something: 'foo',
+    };
+    const userRule = { foo: 'bar' };
+    webpackConfig = getWebpackConfig(
+      {
+        module: {
+          additionalOption,
+          rules: [userRule],
+        }
+      },
+      cosmosConfigPath
+    );
+
+    expect(webpackConfig.module.additionalOption).toEqual(additionalOption);
+    expect(webpackConfig.module.rules[0]).toEqual(userRule);
   });
 });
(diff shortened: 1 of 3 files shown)


skidding:
@nik-kor Can you please tell me what version of React Cosmos were you using?

This shouldn't be a problem since `v2.0.0-beta.5` because we moved webpack to `peerDependencies` and use the user's version of webpack (be it 1 or 2). Before–and what I think you experienced–you could have a webpack 2 config (using *rules*) passed to Cosmos which would run it using webpack 1 (only working with *loaders*).

Sorry to hear you wasted so much time on this. A much more stable 2.0 is [almost here](https://github.com/react-cosmos/react-cosmos/projects/3).

nik-kor:
hey, @skidding 
I use `2.0.0-beta.5`. The thing is if you pass `{ modules.rules } ` then final config will be something like `{ modules: { rules: [], loaders: [] } }`

nik-kor:
And also here https://github.com/nik-kor/react-cosmos/blob/v2.0.0-beta.5/packages/react-cosmos-webpack/src/webpack-config.js#L55 you expect  `loaders`

skidding:
> hey, @skidding
I use 2.0.0-beta.5. The thing is if you pass { module.rules } then final config will be something like { module: { rules: [], loaders: [] } }

You are correct about the output. I was under the impression that since webpack 2 supports both `loaders` (old) and `rules` (new), it would process both when found. But from your experience it seems like once it encounters `loaders`, it stops looking for `rules`. If this is so then that line you referenced needs to be changed to extending either `rules` or `loaders` based on what the user already defined. What do you think?

nik-kor:
I got your point. Maybe the guess that it will process both is correct. I can check it. 
But also the connection between the original problem(`webpack` cannot inline constants via `DefinePlugin`) and loader/rules is not obvious.

nik-kor:
i believe it will take the one https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L46

skidding:
> i believe it will take the one https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L46

It seems so. Thanks for digging into this!

Would you like to replace this PR with changing the logic [here](https://github.com/nik-kor/react-cosmos/blob/a9f430cab2d0272ccb4460e203f9726b5f3facae/packages/react-cosmos-webpack/src/webpack-config.js#L51-L61)? They best way would be to see what is the user using–loaders or rules–and extend that with the cosmos loader, using the appropriate syntax according to the case.

I'm wondering what's the best failing test we should create for this first...

Also, it might be worth worth waiting until I merge #274 today to avoid conflicts, because I refactored the server tests.

packages/react-cosmos-webpack/src/webpack-config.js
```diff
@@ -69,8 +71,7 @@ export default function getWebpackConfig(
     entry,
     output,
     module: {
-      ...userWebpackConfig.module,
```
nik-kor: not sure if we need it. But in current point it will make config invalid

packages/react-cosmos-webpack/src/webpack-config.js
```diff
@@ -44,11 +44,13 @@ export default function getWebpackConfig(
     publicPath: '/loader/',
   };
 
-  const loaders = userWebpackConfig.module && userWebpackConfig.module.loaders ?
-    [...userWebpackConfig.module.loaders] : [];
+  // use what user wants
+  const webpackRulesOptionName = userWebpackConfig.module && userWebpackConfig.module.rules ? 'rules' : 'loaders';
```
nik-kor: why the webpack guys decided to use `rules`....

nik-kor: can't see any profit from this renaming

packages/react-cosmos-webpack/src/webpack-config.js
```diff
@@ -44,11 +44,13 @@ export default function getWebpackConfig(
     publicPath: '/loader/',
   };
 
-  const loaders = userWebpackConfig.module && userWebpackConfig.module.loaders ?
-    [...userWebpackConfig.module.loaders] : [];
+  // use what user wants
```
skidding: Can you please improve this comment to explain the reason behind this is to supporting webpack 1 and 2?

packages/react-cosmos-webpack/src/webpack-config.js
skidding: I guess they wanted old configs to still work while providing an improved loader API for those who have the time to upgrade.

packages/react-cosmos-webpack/src/__tests__/webpack-config.js
```diff
@@ -57,8 +59,32 @@ describe('without hmr', () => {
     webpackConfig = getWebpackConfig(userWebpackConfig, cosmosConfigPath);
   });
 
-  test('keeps user loaders', () => {
-    expect(webpackConfig.loaders).toBe(userWebpackConfig.loaders);
+  test('choose `loaders` or `rules` from user config', () => {
```
skidding: Please move this test outside of the [`without-hmr` describe block](https://github.com/nik-kor/react-cosmos/blob/7abc2236f024217690c6dcfa0b2e09004263ae40/packages/react-cosmos-webpack/src/__tests__/webpack-config.js#L51) as it doesn't benefit from the beforeEach setup. Best to put it at the end of the file.

skidding:
@nik-kor Thanks!

nik-kor:
thank you too!
great project btw!

skidding:
Published under `2.0.0-beta.6`

Issue #283, opened 2017-02-04
how to deploy playground?


jlc467:
### What's up?

Want to achieve playground deployment like https://github.com/react-cosmos/react-cosmos.github.io

### Mkay, tell me more...

I'm using the default webpack config. I don't see any options for writing a bundle so i can deploy.

skidding:
There's no official API for it at the moment. I changed the source code of react-cosmos-webpack on my machine to save the bundle on disk instead of memory to create that export. Are you interested in playing with the source code (maybe even contribute) to achieve this? I can provide pointers to the minimum changes required and how to run a local copy of Cosmos on your codebase. Otherwise I'll keep this open as a feature request until somebody has the time to implement it properly.

jlc467:
> Are you interested in playing with the source code (maybe even contribute) to achieve this?

Sure. Any pointers are appreciated. I have the react-cosmos repo up and running locally. I added a view source button [here](https://github.com/jlc467/react-cosmos/pull/2), but it only really shines when you deploy the playground for viewing when the source isn't on your local.
![screen shot 2017-02-04 at 2 48 39 pm](https://cloud.githubusercontent.com/assets/6405557/22621112/cf36ecaa-eae9-11e6-9131-98a7a82aa720.png)

My end goal is to use react-cosmos as more of a code notebook and less as a component testing solution.

skidding:
I just read your PR, very cool! Are you planning to support code changes as well in the source editor?

Will come back with instructions. 💨

skidding:
The code that bundles webpack (currently into memory via webpack dev middleware) is inside [server.js](https://github.com/react-cosmos/react-cosmos/blob/ed8ef3e09645cd69d0086339ea6e169294bd4aef/packages/react-cosmos-webpack/src/server.js). Right off the bat, I'd start by duplicating this file into something like `export.js`

Inside export.js, you can get rid of webpack-dev-middleware, webpack-hot-middleware and express stuff, since you're not going to start any server.

This is the key part that you want to preserve (among other path-related logic):

```js
const cosmosWebpackConfig = getWebpackConfig(userWebpackConfig, cosmosConfigPath);
const compiler = webpack(cosmosWebpackConfig);
```

Now, instead of passing that `compiler` instance to the dev middleware, you [run](https://webpack.js.org/api/node/#run) it:

```js
compiler.run((err, stats) => {
  // ...
});
```

[...]

PR #284, opened 2017-02-05
Ability to export bundle for deployment

jlc467:
Running `/packages/react-cosmos-webpack/bin/cosmos-export.js` will export an index.html and bundle.js that can be deployed, e.g. https://react-cosmos.github.io/?component=FlatrisGame&fixture=new-game-running

See https://github.com/react-cosmos/react-cosmos/issues/283 for background

Todo:
- [x] Tests will need to be updated/added


packages/react-cosmos-webpack/src/export.js
```diff
@@ -0,0 +1,58 @@
...
+  const compiler = webpack(cosmosWebpackConfig);
+
+  compiler.run((err, stats) => {
+    if (err) {
+      console.error('Export Failed! See error below:');
+      console.error(err);
+    } else {
+      console.log(stats);
+      try {
+        fs.copySync(path.join(__dirname, 'static/favicon.ico'), `${cosmosWebpackConfig.output.path}/favicon.ico`);
```
skidding: This is a bit more tricky. Cosmos uses two index.htmls:
- [Index of the playground](https://github.com/react-cosmos/react-cosmos/blob/ed8ef3e09645cd69d0086339ea6e169294bd4aef/packages/react-cosmos-webpack/src/static/index.html), used to render nav and controls (this one is missing from the export at the moment)
- Index of the user, used for the contents of the loader frame (this one is generated by the export, but should be placed in the `loader/` dir, next to the `bundle.js` file)

jlc467: > Index of the user, used for the contents of the loader frame (this one is generated by the export, but should be placed in the loader/ dir, next to the bundle.js file)

Seems to work fine without this one, but then I use a web server that is redirecting to the index.html of the playground. Can you elaborate on when the `index of the user` comes into play?

jlc467:
> I managed to tweak the URLs and make it work locally, but I'm still thinking of an elegant way to fix this. How is this working on your end?

The only way I could get it working is by redirecting everything to index.html with a web server. I tried by copying the export to my root and, while the playground shell would load correctly, the iframe would just show the directory browser which does have the `index of the user` in it. Sorry I can't be much help here.

skidding:
Hey @jlc467, happy to see you're still working on this. Looking very good, let me know if you want me to review.

Regarding the other changes required for this to "just work", I started debugging on your branch and found all the missing pieces! Here are the changes that fixed every previous issue on my machine (tests not included):

- Removed comments about saving to memory, because we now save to disk as well
- Make paths relative
- Add production flag to minify/optimize bundle when exporting

image

- More relative paths

image

- More relative paths
- Disabled dev tools hook in production (export) because browser security yells at you when you try to access `window.parent` on the `file:///` protocol

image

Now, you have two options:
- We leave these for a separate PR to keep diffs smaller 
- You integrate these changes into this PR and ship the complete feature (home run)

Up to you. Also let me know if you need any more help. 

PS. This is going to make a great addition to the [upcoming 2.0 release](https://github.com/react-cosmos/react-cosmos/projects/3)!

jlc467:
@skidding Thanks! This has been educational and fun to work on. I'll implement these changes sometime within the next few days, after which i think it would be good for you to review so we can wrap this PR up! 😎

flaviusone:
@jlc467 are you planning to add anything else or is this ready? We are really excited and waiting for this feature!

jlc467:
@flaviusone @skidding  Ready for code review

packages/react-cosmos-webpack/src/__tests__/entry.js
```diff
@@ -43,7 +43,7 @@ describe('loader path', () => {
   let options;
 
   beforeAll(() => {
-    init('/loader/');
+    init('./loader/index.html');
```
skidding: This value mocks `window.location.pathname`, so it should be `/loader/index.html`. Tests should continue to pass.

packages/react-cosmos-webpack/src/__tests__/webpack-config.js
```diff
@@ -225,3 +225,81 @@ describe('loaders', () => {
...
+    mockCosmosConfig = {
+      componentPaths: ['src/components'],
+      fixturePaths: ['test/fixtures'],
+      ignore: [],
+      globalImports: ['./global.css'],
+      hot: true,
+      outputPath: '__mock__outputPath'
+    };
+  });
+  describe('with shouldExport false', () => {
```
skidding: It's annoying that I'm enforcing this without having an ESLint rule, but could you please add a newline between each `describe` and `test` blocks? The extra air improves readability.

jlc467:
@skidding Changes made 😎

skidding:
Congrats @jlc467! 🎉  🎆

PR #299, opened 2017-03-02
Document custom component paths

### LINKED ISSUE #298: componentPaths cannot have a folder with a `index.js` in it
### What's up?

I happened to put one folder with `index.js` in `componentPaths`. The `cosmos.config.js` is like below.

```
module.exports = {
  componentPaths: [
      'src/components/Home'
  ]
};
```

The `src/components/Home` folder has one `index.js` file in it. After launching cosmos, it fails with error message 
>ERROR in ./~/react-cosmos-webpack/lib/user-modules.js
Module build failed: Error: Must implement getComponentName when using exact file paths in componentPaths

I dug into the code and found [this line](https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-voyager/src/index.js#L56) throws the Error. It turns out that [`react-cosmos-config`](https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-config/src/index.js#L7) utilizes `resolveFrom` to normalize component path. The npm `resolveFrom` returns path with `index.js` if there is `index.js` in given folder, which make `react-cosmos-voyager` think it is exact file paths.

IMHO, `componentPaths` should allow existence of `index.js` without requiring `getComponentName`.

### Mkay, tell me more...

There is no documentation about `getComponentName` found.

skidding:
Fixes #298

Added a separate section instead of putting `getComponentName` with all the other options because I'm planning to split all options per use case sections in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,29 @@ module.exports = {
 };
 ```
 
+#### Custom component paths
+
+The `componentPaths` option supports both dir and file paths. Most of the times using a dir path is enough (e.g. "src/components"), but we might need to target the exact file paths when component files sit next non-component files. Here's an example for this type of setup:
+
+```js
+// cosmos.config.js
+componentPaths: [
+  'src/components/A.jsx',
+  'src/components/B.jsx'
+],
+getComponentName: componentPath =>
+  componentPath.match(/src\/components\/(.+)\.jsx$/)[1]
+```
+
+> **Note: The `getComponentName` option needs to be added when using file paths.**
+
+We can also use [glob](https://github.com/isaacs/node-glob) instead of adding each component by hand:
+
+```js
+// cosmos.config.js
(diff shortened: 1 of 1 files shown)


skidding:
@mocheng please review

README.md
```diff
@@ -160,6 +160,29 @@ module.exports = {
 };
 ```
 
+#### Custom component paths
+
+The `componentPaths` option supports both dir and file paths. Most of the times using a dir path is enough (e.g. "src/components"), but we might need to target the exact file paths when component files sit next non-component files. Here's an example for this type of setup:
```
ovidiubute: > when component files sit next non-component files

Should be: "when component files sit next **to** non-component files"

skidding: Fixed!

mocheng:
+1

Issue #288, opened 2017-02-13
Module resolution error on Windows


reinei:
### What's up?

I am trying to load proxies in my (new and clean) project.
However, the webpack packing process fails with the following error:
```
ERROR in ./~/react-cosmos-webpack/lib/user-modules.js
edux-proxy.js' in 'D:\Development\centrum-sys\node_modules\react-cosmos-webpack\lib'
 @ ./~/react-cosmos-webpack/lib/user-modules.js 21:14-66
 @ ./~/react-cosmos-webpack/lib/entry.js
 @ multi (webpack)-hot-middleware/client.js?reload=true ./~/react-cosmos-webpack/lib/entry.js
```
(the error is literally stripped, but trying to load cosmos with only one proxy results in the full line being shown:
```Module not found: Error: Can't resolve 'D:Developmentcentrum-syscontext-proxy.js' in ```
note the missing slashes)

### Mkay, tell me more...

I am using the following libraries

* react-cosmos-context-proxy: 2.0.0-beta.6
* react-cosmos-redux-proxy: 2.0.0-beta.6
* react-cosmos-webpack: 2.0.0-beta.7
* webpack: 2.2.1
* webpack-dev-server: 2.3.0
* react: 15.4.2
* redux: 3.6.0

(all have been installed without version tags, which I assume means the beta builds are fine OR the last npm publish did not tag them correctly)

I am running node v7.2.0 on Win10 64bit.

my config files are attached below.
[config.zip](https://github.com/react-cosmos/react-cosmos/files/771136/config.zip)

I hope this is enough information to allow someone to point at my mistake

PS: is this even the right place to ask these questions?

skidding:
Hi @reinei 

This is the right place. Unfortunately most of the users (myself included) are on OSX, so I apologize if the current version is broken on Windows. Maybe we can improve on this together while fixing your installation.

I'll start with some off the top of my head suggestions:
- What's weird about `'D:Developmentcentrum-syscontext-proxy.js'` is that you only have `./redux-proxy.js` in your `cosmos.config.js`, so where could `...context-proxy.js` come from?
- Can you try cloning react-cosmos and [running one of the examples](https://github.com/react-cosmos/react-cosmos/blob/master/CONTRIBUTING.md#get-familiar-with-codebase)? This should tell us whether we have a generic problem on Windows or if we should debug your config more.

Thanks!

reinei:
Hey skidding,
[...]

PR #301, opened 2017-03-02
Link to standalone examples

skidding:
Fixes #300

PR #303, opened 2017-03-04
Fix dev environment on Windows

### LINKED ISSUE #289: Fix dev environment on Windows
### What's up?

As [reported](https://github.com/react-cosmos/react-cosmos/issues/288#issuecomment-279400489) by @reinei it seems like the build scripts don't work on Windows environments.

### Mkay, tell me more...

It might have something to do with how `PACKAGE` is passed via [environment var](https://github.com/react-cosmos/react-cosmos/blob/ed8ef3e09645cd69d0086339ea6e169294bd4aef/scripts/build.js#L26-L28) to [npm script](https://github.com/react-cosmos/react-cosmos/blob/ed8ef3e09645cd69d0086339ea6e169294bd4aef/package.json#L13). 

@ovidiubute any ideas or do you have a Windows machine to test on?

RadValentin:
Fixes #289 and #288


RadValentin:
@skidding @ovidiubute Can you guys take a look? Windows users should be OK now, the only thing that I couldn't get to work were the tests for the Flatris example as they [reference an external `node_modules`](https://github.com/react-cosmos/react-cosmos/blob/master/examples/flatris/package.json#L26) but that should be fixed with #301.

packages/react-cosmos-config/src/index.js
```diff
@@ -1,11 +1,12 @@
 import path from 'path';
+import slash from 'slash';
 import resolveFrom from 'resolve-from';
 import importModule from 'react-cosmos-utils/lib/import-module';
 
 const resolveUserPath = (userPath, rootPath) =>
-  path.isAbsolute(userPath) ? userPath : (
+  slash(path.isAbsolute(userPath) ? userPath : (
```
ovidiubute: You don't need to use `slash` for the entire branch. `path.join` already uses the platform specific separator. In this particular case you could call `slash` just for the `userPath`. [Source](https://nodejs.org/api/path.html#path_path_join_paths)

packages/react-cosmos-voyager/src/__tests__/index.js
```diff
@@ -1,8 +1,9 @@
 import path from 'path';
+import slash from 'slash';
 import traverse from 'traverse';
 import getFilePaths from '../index';
 
-const resolvePath = relPath => path.join(__dirname, '../use-cases', relPath);
+const resolvePath = relPath => slash(path.join(__dirname, '../use-cases', relPath));
```
ovidiubute: `relPath` seems to be the only possible offender, so call `slash` only on that.

packages/react-cosmos-config/src/index.js
skidding: > path.join already uses the platform specific separator

@ovidiubute yes, but it doesn't normalize the existing slashes from the concatenated paths. Am I right?

ovidiubute: Yeah I see what you mean, `path.join` is called with `userPath` which could have any combination of forward/back slashes. So `slash` is indeed required. 👍

skidding:
@RadValentin I agree with https://github.com/react-cosmos/react-cosmos/pull/303#discussion_r104289808, otherwise looks good! Also, you have merge conflicts to solve. Sorry bout that! 👨‍👩‍👧‍👦

packages/react-cosmos-voyager/src/__tests__/index.js
RadValentin: Unfortunately what ends up happening is that `path.join` will try to use OS specific slashes so it will use backslashes for Windows but we're testing against paths that were processed via glob which [insists on forward slashes](https://github.com/isaacs/node-glob#windows). That's why I'm running everything through `slash`

RadValentin:
@skidding @ovidiubute Adressed comments, can you take another look?

skidding:
@reinei can you please try @RadValentin's branch on your machine to make sure the initial issues were corrected. It would help confirm Cosmos works for Windows users. Thanks!

reinei:
```npm install``` is successful on Windows 10 node v7.2.0 npm 3.10.9
and I am able to start the context example without any problems. (If you want, I can see if the changes fixed my initial problems in my own repo as well, but I very much assume thats the case, seeing as I was using a context proxy as well.)

skidding:
@reinei That's good enough, thanks! Let us know if you encounter any other issues in the future.

@RadValentin saved the day!

PR #312, opened 2017-03-11
Separate fixture props

jlc467:
Issue: https://github.com/react-cosmos/react-cosmos/issues/217

With this PR, fixture props are expected to be attributes of `fixture.props` instead of the the root-level `fixture`.

**Also in this PR:** local-state example shows how to use props in a fixture.

To test backwards compatibility, I copied the local-state example and put props at the root-level, then added a `FixFixturePropsProxy` to cosmos config to confirm the props come through. Before merging, I imagine this example will be deleted and the proxy extracted into some kind of markdown documentation in the release notes.

Haven't looked into codemods yet, but figure that can come later.


jlc467:
Well explained... I'll make these changes before the end of the weekend. Wondering how many levels deep I need to check to find unserializable props? I'll start with props root for starters.

skidding:
One level deep (root props) is as good as before. Ideally the splitting would be recursive and would be any level deep. Wondering if there's an easy algorithm for this. The key here is how we merge the serializable/unserializable parts back, which [until now was trivial](https://github.com/react-cosmos/react-cosmos/blob/c3ff73fd47a56c4ebb1314ace41aab4587cebb71/packages/react-cosmos/src/components/Loader.jsx#L125-L128).

Finally, as further context, my wish is to someday add a notice at the bottom of the fixture editor like this:
> Some fixture attributes are unserializable and not shown here: `props.onClick`, `foo.bar`, `baz`

jlc467:
Still todo:
- [ ]  Make `NormalizeStateProps` a working package (not sure what is needed here as far as lerna and tooling config)
- [ ]  Tests for `NormalizeStateProps`
- [ ]  More tests for `StateProxy` and `splitUnserializableParts`
- [ ]  Finish backwards compatibility example (my vote is to move it in a folder separate from /examples to avoid confusion with old API)
- [ ]  Code review on latest changes (see below changes that I'm iffy about)

- I use _.pick + _.omit in the proxy because I thought we don't want to leave behind what we transfer to fixture.props
- `StateProxy` passes fixture untouched to `PropsProxy` now

packages/react-cosmos/src/components/proxies/__tests__/StateProxy.jsx
```diff
@@ -83,7 +83,7 @@ describe('fixture without state and component without initial state', () => {
 
   commonTests();
 
-  test('omits state from fixture sent to next proxy', () => {
+  test('does not omit state from fixture sent to next proxy', () => {
```
skidding: I think this test case is useless and can be removed

packages/react-cosmos-utils/src/unserializable-parts.js
```diff
@@ -28,20 +28,36 @@ const isSerializable = obj => {
...
-  // TODO: Perform nested separation of unserializable parts. Required if props
-  // will move to fixture.props https://github.com/react-cosmos/react-cosmos/issues/217
-  const serializable = {};
-  const unserializable = {};
+  const splitResult = {
+    serializable: { },
+    unserializable: { }
+  };
+
+  const { serializable, unserializable } = splitResult;
```
skidding: This looks convoluted. Why not go back to just
```js
const serializable = {};
const unserializable = {};
// ...
return { serializable, unserializable };
```

packages/react-cosmos-utils/src/unserializable-parts.js
```diff
@@ -28,20 +28,36 @@ const isSerializable = obj => {
...
 
   Object.keys(obj).forEach(key => {
     if (isSerializable(obj[key])) {
       serializable[key] = obj[key];
+    } else if (key === 'props' && isPlainObject(obj[key])) {
+      Object.keys(obj.props).forEach(propKey => {
+        let propHome = 'serializable';
```
skidding: Moreover, you could do something like 

```js
const propHome = !isSerializable(obj.props[propKey]) ? unserializable : serializable;

if (!propHome.props) {
  propHome.props = {};
}
propHome.props[propKey] = obj.props[propKey];
```

skidding: But it would be wise to add the new tests first and then beautify the implementation.

packages/react-cosmos-normalize-props-proxy/src/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+import omit from 'lodash.omit';
+import pick from 'lodash.pick';
+
+const getFixedFixture = fixture => {
+  if (fixture.props) {
+    // the fixture doesn't need fixing if it has props
+    return fixture;
+  }
+
+  const notProps = ['children', 'state', 'context', 'reduxState'];
```
skidding: You can put this outside of `getFixedFixture` since they never change.

packages/react-cosmos-normalize-props-proxy/src/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const fixedFixture = Object.assign(pick(fixture, notProps), {
+    props: omit(fixture, notProps)
+  });
+  return fixedFixture;
```
skidding: How about

```js
return {
  ...pick(fixture, notProps),
  props: omit(fixture, notProps)
};
```

packages/react-cosmos-normalize-props-proxy/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "react-cosmos-normalize-props-proxy",
+  "version": "2.0.0-beta.6",
+  "description": "Seperate props in fixtures",
```
skidding: "Proxy for backwards compatibility with old react-cosmos fixture format"

packages/react-cosmos-normalize-props-proxy/src/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+class NormalizePropsProxy extends React.Component {
+  render() {
+    const { nextProxy, fixture } = this.props;
+
```
skidding: Remove line

examples/local-state-with-normalize-props-proxy/cosmos.config.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  componentPaths: ['components'],
+  // TODO: use package for normalize proxy
+  proxies: ['../../packages/react-cosmos-normalize-props-proxy/src']
```
skidding: No need for a TODO since examples are explicitly working with relative packages, but you should rely on the compiled lib by just pointing to `'../../packages/react-cosmos-normalize-props-proxy'`

jlc467:
Awesome... looks like this is pretty close to done. Review makes sense to me, so I'll make the changes as suggested. Should be able to get it done during the week, if not definitely over next weekend.

jlc467:
@skidding Implemented feedback and added tests. Not sure where you want the NormalizePropsProxy example to live. I've seen projects split up the examples folder into e.g. `simple-examples-start-here`, `advanced-examples`, `real-world`. I think this is a `real-world` example of how to achieve backwards compatibility, but `simple` everything else.

packages/react-cosmos-normalize-props-proxy/src/__tests__/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+  test('sends nextProxy.next() to next proxy', () => {
+    expect(childProps.nextProxy).toBe(nextProxyNext);
+  });
+
+  test('sends designated root level fixture fields to next proxy unmodified', () => {
+    expect(childProps.fixture.context).toEqual({});
+    expect(childProps.fixture.children).toEqual([]);
+    expect(childProps.fixture.state).toEqual({});
+    expect(childProps.fixture.reduxState).toEqual({});
```
skidding: These checks can become false positives in time. How about creating some constants like `const mockContext = {}`, use them when constructing the fixture in `renderProxy` calls and then do strict checks like `expect(childProps.fixture.context).toBe(mockContext);`

skidding:
Discovered a problem with [how fixture parts are merged](https://github.com/jlc467/react-cosmos-notebook/blob/596ce2eb197702516498374d95cbb87200f65dce/packages/react-cosmos/src/components/Loader.jsx#L125-L128). The problem is when you have one or more unserializable props and one or more serializable props, the unserializable props are overridden by the serializable ones. We need to merge the parts recursively. [_.merge](https://lodash.com/docs/4.17.4#merge) should do the trick.

Additionally, while trying to integrate in own projects I realized I had more *magic* fixture fields that aren't props (designed to be picked up by custom proxies), which are being normalized into props. It would help to make `notProps` customizable via proxy option (e.g. [fixture key for context proxy](https://github.com/react-cosmos/react-cosmos/blob/c3ff73fd47a56c4ebb1314ace41aab4587cebb71/packages/react-cosmos-context-proxy/src/index.jsx#L10)). Is it possible to add this as well?

Thanks! 🙌

jlc467:
>  It would help to make notProps customizable via proxy option (e.g. fixture key for context proxy). Is it possible to add this as well?

Yeah, sure. No telling what kind of magical custom proxies people are running. To pass the options, the correct approach is to wrap the proxy, e.g. [as seen here](https://github.com/skidding/illustrated-algorithms/blob/master/cosmos/proxies/context-proxy.js), correct?

Will fix the bug too. Not totally clear on why the fixture is split and then merged again (why not just use the original fixture?), so will read more into that.

Should be able to get this done tomorrow.

skidding:
> Yeah, sure. No telling what kind of magical custom proxies people are running. To pass the options, the correct approach is to wrap the proxy, e.g. [as seen here](https://github.com/skidding/illustrated-algorithms/blob/master/cosmos/proxies/context-proxy.js), correct?

Yes... but in that example we're composing proxies. I think the Redux proxy is also good inspiration. The idea is to put the whole proxy component inside a create function, in order to have access to the options passed to that create function.

image

> Will fix the bug too. Not totally clear on why the fixture is split and then merged again (why not just use the original fixture?), so will read more into that.

Because of the fixture editor. Users can edit the serializable part of the fixture, which is merged with the unserializable part (you could call that part immutable) to form the new fixture.

image

Let me know if this makes sense. Thanks!

jlc467:
> Because of the fixture editor. Users can edit the serializable part of the fixture, which is merged with the unserializable part (you could call that part immutable) to form the new fixture.

Makes sense!

> Yes... but in that example we're composing proxies. I think the Redux proxy is also good inspiration. The idea is to put the whole proxy component inside a create function, in order to have access to the options passed to that create function.

I think my question is geared towards composing proxies. How would the user pass options to the new `NormalizePropsProxy`. I will definitely reference the Redux proxy to implement the ability to pass `notProps` option to `NormalizePropsProxy`

skidding:
Oh, you're right about the proxy wrapping, that's how you pass options. It's also seen in the two proxy examples from the README: https://github.com/react-cosmos/react-cosmos#react-cosmos-redux-proxy

In the future we could allow proxy options inside cosmos.config (similar to webpack loader options), but I don't think Cosmos is popular enough yet for that sort of customization :P

jlc467:
> In the future we could allow proxy options inside cosmos.config (similar to webpack loader options), but I don't think Cosmos is popular enough yet for that sort of customization :P

Yeah, I'm not sure the overall idea of developing components in isolation has hit mainstream yet. People seem more keen on the UI explorer/documentation aspect. Sometimes the killer feature of a project is misidentified early on, and it isn't until later that people realize that the better benefit is something else. Personally, I'm only doing cosmos-first development from now on, where cosmos playground is the entry point and you delay writing nav / router type components that can distract from the app's idea.

skidding:
> Yeah, I'm not sure the overall idea of developing components in isolation has hit mainstream yet. People seem more keen on the UI explorer/documentation aspect. Sometimes the killer feature of a project is misidentified early on, and it isn't until later that people realize that the better benefit is something else.

Yes, I couldn't agree more! ❤️ 

> Personally, I'm only doing cosmos-first development from now on, where cosmos playground is the entry point and you delay writing nav / router type components that can distract from the app's idea.

Same here! 🥂

jlc467:
@skidding Let me know if you think of anything else. A new test to cover the serializable overwriting unserializable crossed my mind, but wasn't immediately obvious how to tackle that. I imagine such a test would live with the fixture editor tests.

skidding:
I think this is great and ready for prime time. 🥇 

@ovidiubute @flaviusone @RadValentin you might want to take a look at this! It is breaking, so we need to make sure your react-cosmos version(s) are locked at `2.0.0-beta.8` until you upgrade, or be able to upgrade right away. Since I think think you have any custom proxies, adding `react-cosmos-normalize-props-proxy` to your proxy list should be enough. Let me know!

valentin-radulescu-hs:
We're good on our end, the other guys are still on `1.0.1` and we're locked to `2.0.0-beta.8` with yarn. Worst case we'll have to add the normalize proxy in one place which isn't a problem.

Fire away! 💣

skidding:
Great contribution @jlc467, thanks again making this happen! Oh, and I sent you an invitation to become a collaborator 🙌

jlc467:
@skidding Sweet! And thanks for helping me with the contributions 🙏.  I know README updates are still missing for both this and cosmos-export, so if that hasn't been picked up in the next couple weeks, I will try to contribute there.

skidding:
@jlc467 Welcome on board! ⚔️ 

Here are two examples where I've already upgraded:
- [Flatris](https://github.com/skidding/flatris/commit/b402f4e90eb780d52d11f14c988abf41f1cb7a75)
- [Illustrated Algorithms](https://github.com/skidding/illustrated-algorithms/commit/5cf3851ceada3389e4a17f7c82daf71570e73264)

Regarding docs, sure! I hope the latest structural README changes make it easier to contribute. You're also invited to join the [Slack channel](https://join-react-cosmos.now.sh/).

PR #332, opened 2017-04-10
Replace includes with indexOf for resolvingUserPath in Config

jnishiyama:
For compatibility with Node.js 5+


skidding:
Thanks @jnishiyama!

PR #339, opened 2017-04-15
Upgrade Lerna and switch to Yarn 📦

### LINKED ISSUE #245: Add the possibility of installing deps using yarn
### Need
As a developer, I'd like to quickly install dependencies when developing `react-cosmos`, so that I can be more productive.

### Deliverables
- `yarn.lock` files for all packages, and top-level
- Modify the npm `install-react` task to use `yarn` if available

### Problems
- Lerna and Yarn don't play together... yet. There is however hope, as the code required measures about [two lines](https://github.com/motion/lerna/pull/4) and there is [active discussion](https://github.com/lerna/lerna/issues/371) around the topic.

skidding:
Fixes #245

Lerna supports Yarn!

PS. Started dropping support for `0.13`. It became harder to switch between React versions that don't have `react-dom` and I realized we weren't fully supporting it anyway.


CONTRIBUTING.md
```diff
@@ -49,21 +49,21 @@ git clone git@github.com:react-cosmos/react-cosmos.git
...
 # Build example from source and test React Cosmos holistically
 cd examples/context
-npm install
-npm start
+yarn install
+yarn start
 
 # Watch & build single package (running example will live reload)
-npm run build react-component-playground -- --watch
+yarn run build react-component-playground -- --watch
```
ovidiubute: There is no need to use `run` when using `yarn`, it's implicit afaik.

CONTRIBUTING.md
```diff
@@ -49,21 +49,21 @@ git clone git@github.com:react-cosmos/react-cosmos.git
 cd react-cosmos
 
 # Install deps and link child packages (using Lerna)
-npm install
+yarn install
```
ovidiubute: You can simply call `yarn`, `install` is implicit.

ovidiubute:
This will be a huge improvement in speed when developing cosmos 👍

skidding:
@ovidiubute I like the implicit form because it saves key strokes, but I prefer keeping the explicit form in the docs because you never know the level of familiarity the user has with Yarn.

ovidiubute:
@skidding the `yarn` docs list `yarn install` as an alternative way of installing deps, I'd say the idiomatic way is to just write `yarn`. For running targets, yeah, it may be clearer to use it the same way as you would npm.

skidding:
@ovidiubute addressed comments!

Issue #282, opened 2017-02-04
404 page


skidding:
### What's up?

Getting a blank page after you remove or rename a fixture is annoying.

### Mkay, tell me more...

CP should check whether the component and fixture paths are valid and if not render a new 404 screen, similar to [how the welcome screen is shown](https://github.com/react-cosmos/react-cosmos/blob/ed8ef3e09645cd69d0086339ea6e169294bd4aef/packages/react-component-playground/src/components/component-playground.jsx#L197). Would look great on the new space bg @RadValentin created.

E.g.

> Invalid coordinates
> No astronomical object found at ${COMPONENT}:${FIXTURE}

ganderzz:
@skidding If no one is opposed, I'll try my hand at this. 😄

skidding:
It's all you @ganderzz!

skidding:
Thank you @ganderzz! This has been a long time pain, especially when switching between projects.

PR #341, opened 2017-04-15
Implement 404 page

ganderzz:
#282

Implement 404 page.

### DIFF
--- a/packages/react-component-playground/src/components/component-playground.jsx
+++ b/packages/react-component-playground/src/components/component-playground.jsx
@@ -388,7 +388,7 @@ module.exports = React.createClass({
 
     this.updateContentFrameOrientation();
 
-    if (this.isFixtureSelected && this.doesSelectedFixtureExist()) {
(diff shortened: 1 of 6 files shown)


packages/react-component-playground/src/components/component-playground.jsx
```diff
@@ -60,7 +61,13 @@ module.exports = React.createClass({
...
 
       // This returns the fixture contents as it is initially defined, excluding any modifications.
-      return component && fixture && fixtures[component][fixture];
+      return component && fixture && fixtures[component] && fixtures[component][fixture];
+    },
+
+    isValidComponentAndFixture(props) {
+      const { fixtures, component, fixture } = props;
+
+      return Boolean(fixtures && fixtures[component] && fixtures[component][fixture]);
```
ganderzz: Valid point! Mostly there as a safe guard, but really isn't needed.

ganderzz:
@skidding Adjustments implemented!

PR #342, opened 2017-04-16
Drop React 0.13 support and replace dep forks #340

### LINKED ISSUE #340: Drop React 0.13 support
### What's up?

It became apparent in #339 that we no longer fully support 0.13 so it's time to move on and remove it completely.

### Mkay, tell me more...

This has a number of advantages:
- Smaller maintenance cost
- We can get rid of `react-dom-polyfill`
- We can get rid of `react-split-pane` and `react-codemirror` forks
- Unit tests can be run on all supported React versions (they weren't running on 0.13 for a long time)

skidding:
Closes #340 

RIP `react-dom-polyfill` ⚰️  

*What is dead may never die*


flaviusone:
I don't know how much usage `react-dom-polyfill` had out there but maybe you should have in mind moving the code to a separate legacy repo. The package served us well 🗡 

*And now his watch is ended.*

skidding:
Will do if anybody asks. The npm package will not go away so nothing should break even if anybody used it outside RC. Onwards! ⚔️

Issue #351, opened 2017-05-03
Playground Fullscreen/Editor on Welcome & 404 screens


ganderzz:
### What's up?

When on the 404 page, the editor button is visible on the left nav though clicking it provides no editor. Like on the welcome page, both the editor and fullscreen buttons probably shouldn't be visible.

Also, navigating to a page with a starry bg (welcome & 404) and `fullScreen=true` in the url causes the page to go fullscreen, but the content area doesn't go fullscreen. Even though the buttons are disabled, this should still be a caught case with url parameters.

Thoughts?

### Mkay, tell me more...

- Update `component-playground` to only render the renderMenu when both `isFixtureSelected` and `doesSelectedFixtureExist` is true
- `display-screen__display-screen` has a css left property of 250px. This should be 0px during fullscreen mode.

skidding:
Good catch! Your action items are sound. Feel free to take a stab at it ⚔️

ganderzz:
Sounds good. I'll have a PR tonight! 👍

PR #353, opened 2017-05-04
Implement fixes for 351

ganderzz:
Fix issues mentioned in #351


skidding:
This works well, but I think we can get away without making global classes. The `content-frame` class already has the desired behavior (`left: 250px` normally, `left: 0` in full screen). We just need to wrap all content (Welcome screen and Error screen as well) inside `content-frame` in the playground code. Then we can set `left: 0` to display-screen for all cases, as it will take the space of its parent node. What do you think?

ganderzz:
I'm all for removing the globals.

Let me mess around with it, and I'll push the changes up!

skidding:
👌

Issue #314, opened 2017-03-14
Group components in a single view


ovidiubute:
### What's up?

Non-technical people have trouble navigating CP if your project has a lot of components and fixtures. Ideally they'd like to be able to see similar components grouped together on one page, or have all of the fixtures in a single page for one component.

### Mkay, tell me more...

I'm thinking we have a tree structure in the left nav menu, where if you click on the directory name you'll get a page with all of the components within it rendered under a specific fixture (this could be configurable, let's say that all of your components have a `base.js`, this would be configured in `cosmos.config.js`).

In addition to that, if you click on the name of a component, you should see a view of your component rendered in all states (or a subset of them). This would also be configurable, maybe we say render 3 fixtures per component, or maybe, wait for it: specify which fixtures are the most interesting to render (effectively showcasing your component in the most interesting states).

Obviously the pain points here are going to be:
* Working on the nav menu is a bit of a pain
* Extracting the directory from the name of the components
* Rendering more than one fixture into CP at a time (what happens to modals, for example, which render on the body and take over your screen?)

There are more questions here but I'd love to hear your opinion on this, @skidding. I think it's a worthwhile investment to have this feature and it's something I've seen in other playgrounds as well, the difference being that they all require the users to write JSX, which is very difficult to maintain in the long term.

maciej-ka:
Uber happy about it, I'm starting now, thanks.

Issue #297, opened 2017-02-27
Improve publicPath


skidding:
### What's up?

`publicPath` currently maps a static dir to the `/loader/` path (which is `/` as far as the user bundle is concerned). This works for some, but a common pattern is to map the static folder to a path other than `/`, like `/static/`), which means `publicPath` should maybe become a dict, like:

```js
publicPath: {
  '/static/': '/path/to/static/files/'
}
```

### Mkay, tell me more...

This is going to make sense after #284 is merged. Exporting needs to be taken into account fully in this issue.

TODO: Provide more context

reinei:
Another suggestion from my side would be to somehow check (maybe hasOwnKey?) if the webpack-dev-middleware can serve an index.html and rewrite requests to that address instead of serving your own static index.html .

In my use case I have 3 bundled files to serve and am using WebpackHtmlPlugin as well as the AddAsset plugin to build a nice index.html to load all required files in the required order, however simply executing ```npm run cosmos``` will serve the static index.html and thus fail due to my main bundle not containing the react libs (they are in a special vendor.dll chunk).

So I currently rigged my local copy of the server.js file to serve my own copy of index.html from the memory-fs instance, but as anyone can guess, this is a bad solution due to using private properties of the dev-middleware and requiring modification of files inside node_modules.

skidding:
Let's see if I got this straight. You have an `index.html` in your publicPath, which serves as a template for `html-webpack-plugin`. Normally, webpack would use the `index.html` generated by the `html-webpack-plugin` but Cosmos instead uses the `index.html` template from your publicPath. Is this right?

reinei:
Not quite,
I have the following Plugin stack:
```
DllReferencePlugin(s)
WebpackHtmlPlugin (using src/index.ejs and outputting to index.html in output dir)
AddAssetPlugin (to add the required dll files)
```

now my `src/index.ejs` is pretty simple and looks like:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
    <title><%= htmlWebpackPlugin.options.title %></title>
  </head>
  <body>
    <div id="main">
    </div>
  </body>
</html>
```

Its just there so I can deploy with my bundle instantly (my app renders into #main)

Now my problem comes from cosmos mounting the webpack-dev-middleware into /loader and serving its own static index.html which (obviously) does not include my dll files and thus has no react loaded.

[...]

PR #364, opened 2017-05-20
Separate loader and playground builds

### LINKED ISSUE #360: Separate loader and playground frames completely
### What's up?

This is a better (and opposite) solution for the problems mentioned in #313 (please read this issue before continuing).

The previous solution was to "bring all user code inside the playground frame", but after talking with @NiGhTTraX in depth I've now been convinced that we need to do the opposite "get all user code out of the playground frame".

Reasons to bring all user code inside the playground frame:
- The playground loads the component fixture contents (inside the fixture editor) so it needs the fixture source, which means it needs ALL the user source. This assumption has been debunked, because in fact the fixture editor only shows the serializable part of the fixture, which means we can send it from the loader frame via `window.sendMessage`.

Reasons to get all user code out of the playground frame:
- User code cannot conflict with playground code.
- User can use different React version than the playground component, which means we don't have to update the playground component whenever something changes in the React API and most users upgrade
- Finally, this is in line with [the vision to make Cosmos React-agnostic](https://github.com/react-cosmos/react-cosmos/issues/286). At the moment the playground depends on the user's React copy, but we want to decouple that completely and allow users to use different engines like Preact in the loader frame.

### Mkay, tell me more...

How to get all user code out of the playground frame:
- `react-cosmos-webpack` needs to create 2 separate bundles. This is how RC worked for a short while. We can see the 2 bundles being merged back into 1 in the [diff between 2.0.0-beta.1 and 2.0.0-beta.2](https://github.com/react-cosmos/react-cosmos/compare/v2.0.0-beta.1...v2.0.0-beta.2). Another source of inspiration is the [Browserify example](https://github.com/react-cosmos/react-cosmos/blob/master/examples/browserify/server.js), which starts 2 separate web servers for the playground and loader frames.

To clarify:
- **The playground frame will no longer share the same webpack config as the loader frame**, hence no more DllPlugin issue
- **The playground frame will no longer share the same index.html as the loader frame**

TBD:
- How/when does the loader frame send the fixture contents to the playground frame?

skidding:
Fixes #360


skidding:
@reinei you might want to take a look. This should not longer require you to [hack server.js to make it work for your setup](https://github.com/react-cosmos/react-cosmos/issues/297#issuecomment-284510403). Thanks!

packages/react-cosmos/src/components/Loader.jsx
```diff
@@ -127,20 +161,24 @@ class Loader extends React.Component {
...
     }
 
     return (
       <firstProxy.value
-        key={fixtureUpdateId}
+        key={fixtureId}
         nextProxy={firstProxy.next()}
         component={components[component]}
-        fixture={merge(fixture.unserializable, fixture.serializable)}
+        fixture={merge({}, unserializable, serializable)}
```
jlc467: 👍 Thanks for fixing that.

jlc467:
@skidding Checked out your branch and tried some examples, as well as the cosmos export -- all good! Once this makes it into a beta, I can start testing it everyday.

skidding:
Update: for those of us watching this issue, I did not stop working on it. On the contrary, I went two extra levels deep to fix long standing issues and clean up ancient code. Since we're doing major changes, I figured now or never.

That said, here are the main areas I've refacored
- Completely rewrote Loader. There's now Loader and RemoteLoader. The former is used for testing purposes (more on this later) and receives component and fixture via props. The latter is meant to be controlled remotely via window.postMessage (i.e. by Playground)
- Completely rewrote Playground. Changing the Loader-Playground communication required changes to the Playground, which was always a pain because it was a monolithic component written like 2 years ago, tested wierdly using the old test framework. I've taken it into pieces and rewrote it via latest ES6 & React features, split into a number of smaller components and tested nicely using Jest. **Moreover**, I went the extra mile and added fixtures for CP, which means by the end of this we will work on CP inside CP!

In effect:
- react-cosmos-tree will go away
- Entire Mocha/Karma test suite will go away

WIP:
- Fixture editor now yet added to new CP
- react-cosmos package will also disappear, because it contains both Loader & Playground logic. Most likely we'll have react-cosmos-loader and react-cosmos-playground, each with its own entry point. This also paves the way for non-React loaders.

There are even more goodies included, so stay tuned!

skidding:
Guys, it's time! ⌛️

image

Looking at the diff is insane, although I encourage you to scan it for key parts (tests, fixtures, yarn.lock files, etc. can be ignored). Best to explore [the branch](https://github.com/react-cosmos/react-cosmos/tree/360-separate-playground) and try it out locally.

**Warning: The fixture editor has not yet been added to the new Component Playground yet**, but I wanted to unlock other issues like #314 by merging this gianormous branch first. There are two things that I will submit (small) separate PRs for after this:

- Adding FixtureEditor back (it won't take long, but I didn't want to rush it either since I want to add some enhancements along the way)
- Updating the Contributing Guide (although this branch already adds to it)

Here's an overview of what this branch includes:
- **Complete separation between Loader & Playground.** Playground bundle is now served statically so we only have one build after all)
  - No more `react-cosmos` package. Packages are now better defined with `react-cosmos-loader` and `react-cosmos-component-playground`.
- **Complete rewrite of Playground. `yarn start:playground`** will fire up some Playground Inception right there! 🎇)
  - Removed `react-component-tree` dependency, and got rid of a lot of old code.
- **Loader rewrite.** Split between RemoteLoader (using window.postMessage to communicate) and Loader (uses props). The latter is useful for unit test setup ([see CP tests](https://github.com/react-cosmos/react-cosmos/blob/827e34c8632a2d59246a0c8e2d98b86b6ee8a557/packages/react-cosmos-component-playground/src/components/DisplayScreen/__tests__/index.jsx#L16-L19)).
- **No more Mocha/Karma tests**. Old test suite was removed along with old code, which makes for a cleaner repo and faster build times.
  - Coverage dropped because of `collectCoverageFrom` Jest option. It probably increased, but we weren't properly calculating it before by ignoring files that weren't touched at all by tests. The new metric is fine since it includes new/experimental code like exporting or `react-cosmos-telescope`
- Bonus: First key bindings to ComponentPlayground: `s` will focus on the fixture search input and `ESC` will clear & blur it. @flaviusone will like this.

There's probably more than what I mentioned for feel free to ask any questions. Looking forward to feedback. Thanks!

packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
```diff
@@ -0,0 +1,222 @@
...
+  renderContent() {
+    const { loaderUri, component, fixture } = this.props;
+    const { waitingForLoader, fixtures } = this.state;
+    const isFixtureSelected = !waitingForLoader && Boolean(fixture);
+    const isMissingFixtureSelected =
+      isFixtureSelected && !fixtureExists(fixtures, component, fixture);
+
+    return (
+      <div key="loader" className={styles.loader}>
+        <StarryBg>
```
flaviusone: It would be nice to not render this all the time in the background (for any loaded component). StarryBg is nice but GPU is still working in the background for no reason:
image

skidding: Good catch!

packages/react-cosmos-component-playground/src/components/DisplayScreen/__tests__/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+  shallow(element)
+    .dive() // Loader
+    .dive(); // PropsProxy
+
+describe('DisplayScreen', () => {
+  let wrapper;
+
+  beforeAll(() => {
+    wrapper = shallowLoader(
+      <Loader component={DisplayScreen} fixture={hiThereFixture} />
```
flaviusone: I don't understand why did you chose to render `DisplayScreen` through 'react-cosmos-loader' (and maybe also other components in this PR). The `DisplayScreen` is a pretty basic component independent of the loader and this test will fail if someone breaks the loader. Maybe I'm missing something though.

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -0,0 +1,157 @@
...
+      const componentAndFixture = componentName + fixtureName;
+      const fixtureAndComponent = fixtureName + componentName;
+
+      // Ensure that the fuzzy search is working in both direction.
+      // component + fixture and fixture + component. That's because the user
+      // can search for fixture name and afterwards for component name and
+      // we want to show the correct result.
+      return (
+        match(componentAndFixture, searchText).length !== 0 ||
+        match(fixtureAndComponent, searchText).length !== 0
```
flaviusone: I know this is old code but I vote that we remove matching for `fixtureAndComponent`. 

In time I noticed that this doesn't work that well as it won't prioritize your specific search and the result you want will be further down the list. Also one-way matching (componentAndFixture) is how all major editors work (Sublime, VS Code etc).

I don't know if this change will break a lot of tests but I it's a quick win I think we should do it. WDYT?

packages/react-cosmos-component-playground/src/components/StarryBg/index.jsx
```diff
@@ -0,0 +1,19 @@
...
+        {children &&
+          <div className={styles.content}>
+            {children}
+          </div>}
+      </div>
+    );
+  }
+}
+
+export default StarryBg;
```
flaviusone: This kinda does the same thing as `DisplayScreen` (loads some children) but it doesn't seem to have any tests even if Jest coverage reports 100% (but that's because of the way coverage works). 

Does anything stop us from adding Jest Snapshots?

packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
skidding: Done in #380

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
skidding: Agreed, I also find it confusing. Also fixed in #380

packages/react-cosmos-component-playground/src/components/StarryBg/index.jsx
skidding: Added in #380

packages/react-cosmos-component-playground/src/components/DisplayScreen/__tests__/index.jsx
skidding: We're using the Loader as a test util here. And we're importing the compiled `lib`, not the source. `Loader` is similar to `ComponentTree` back in the day. It allows us to reuse fixtures as test setup. This is a simple scenario indeed but I wanted to use it consistently in all Component Playground tests to showcase its use internally ("eating our own dog food"). It becomes more useful when we use something like a Redux fixture to create a complex test setup in one line.

skidding:
Merged & published under `react-cosmos-webpack@2.0.0-beta.16` :tada: :sparkler:

PR #366, opened 2017-05-31
Fix trailing slash bug and add e2e Cypress tests

### LINKED ISSUE #295: react-cosmos componentPaths should handle trailing '/' well
### What's up?

It takes an hour to detect why my "Home" component is displayed as "ome" component and no fixture is found. At last, it turns out to because `componentPaths` requires path not have trailing '/'.

In a word, `componentPaths: ['src/components]` works, while `componentPaths: ['src/components/]` doesn't work.

This looks really... annoying.

### Mkay, tell me more...

BTW, the latest example in this repo doesn't work. It dumps a bunch of error message.
>ERROR in multi main
Module not found: Error: Cannot resolve 'file' or 'directory'

skidding:
Fixes #295, improves code readability and cherry picks Cypress tests from #364 to bring it sooner into master.


skidding:
@RadValentin @flaviusone @ovidiubute @NiGhTTraX @catalinmiron you might want to take a look, at least at the Cypress tests ✌️

cypress/integration/local-state_spec.js
```diff
@@ -0,0 +1,102 @@
...
+      cy
+        .get(fixtureSel)
+        .should('have.length', 1)
+        .eq(0)
+        .should('contain', 'oneTwoThree');
+    });
+  });
+
+  context('select fixture', () => {
+    // There's only one fixture at this point
```
NiGhTTraX: Because the list is filtered in the previous test? Are Cypress tests not atomic?

cypress/integration/local-state_spec.js
```diff
@@ -0,0 +1,102 @@
...
+    it('should render Loader iframe', () => {
+      cy
+        .get('iframe')
+        .should('have.exist')
+        .should('have.attr', 'src', './loader/index.html');
+    });
+  });
+
+  context('fixture editor', () => {
+    // The first menu button is the fixture editor toggle
```
NiGhTTraX: Maybe give it an identifier?

cypress/integration/local-state_spec.js
```diff
@@ -0,0 +1,102 @@
...
+    const editorButtonSel = `${getSelector('component-playground__button')}:eq(1)`;
+
+    before(() => {
+      cy.get(editorButtonSel).click();
+    });
+
+    it('should add active class to editor button', () => {
```
NiGhTTraX: Are these tests valuable? Isn't this logic checked through unit tests?

cypress/integration/local-state_spec.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    it('should display fixture inside editor', () => {
+      cy.get('.CodeMirror-line:eq(0)').should('have.text', '{');
+      cy.get('.CodeMirror-line:eq(1)').should('have.text', '  "state": {');
+      cy.get('.CodeMirror-line:eq(2)').should('have.text', '    "children": {');
+      cy.get('.CodeMirror-line:eq(3)').should('have.text', '      "c1": {');
+      cy.get('.CodeMirror-line:eq(4)').should('have.text', '        "value": 1');
+      cy.get('.CodeMirror-line:eq(7)').should('have.text', '        "value": 2');
+      cy.get('.CodeMirror-line:eq(10)').should('have.text', '        "value": 3');
```
NiGhTTraX: This seems brittle. I don't think this test should check indentation.

package.json
```diff
@@ -8,13 +8,16 @@
...
+    "start": "./packages/react-cosmos-webpack/bin/cosmos.js --config examples/local-state/cosmos.config.js",
     "build": "node ./scripts/build.js",
     "postinstall": "yarn run bootstrap && yarn run build",
     "publish-lerna": "yarn run test && yarn run build && lerna publish",
     "test-karma": "karma start test/karma.conf.js --single-run",
     "test-jest": "jest --coverage --config test/jest.config.json",
     "test": "yarn run lint && yarn run test-karma && yarn run test-jest",
-    "coveralls": "lcov-result-merger 'test/coverage/*/lcov.info' | node_modules/coveralls/bin/coveralls.js"
+    "coveralls": "lcov-result-merger 'test/coverage/*/lcov.info' | node_modules/coveralls/bin/coveralls.js",
+    "test-cypress": "cypress run",
```
NiGhTTraX: Cypress is a detail, these are acceptance tests, are they not?

package.json
```diff
@@ -108,6 +112,7 @@
       "unicorn/filename-case": 0,
       "import/no-unassigned-import": 0,
       "import/no-extraneous-dependencies": 0,
+      "import/prefer-default-export": 0,
```
NiGhTTraX: Is this the fad these days? To include .eslintrc in package.json?

packages/react-cosmos-voyager/src/__tests__/index.js
```diff
@@ -5,14 +5,20 @@ import getFilePaths from '../index';
...
+  input: {
+    componentPaths = [],
+    fixturePaths = [],
+    ignore,
+    getComponentName,
+    getFixturePathsForComponent,
+  },
+  output
+}) => {
+  const testName = variantion ? `${useCase} (${variantion})` : useCase;
```
NiGhTTraX: Variantion? :grin:

package.json
skidding: It's actually [xo](https://github.com/sindresorhus/xo) in this case, but yes people do that.

package.json
skidding: Will rename. I see them as smoke tests as they don't (and can't, because Cypress doesn't work with iframes 😞 ) cover all functionality.

cypress/integration/local-state_spec.js
skidding: Right.

cypress/integration/local-state_spec.js
skidding: They're not. I added a bit warning at the top of the file.

cypress/integration/local-state_spec.js
skidding: It's going to change for sure when I revamp the CP.

cypress/integration/local-state_spec.js
skidding: Not very. They were just easy to write :). This suite is an experiment, I see it as a trial version. On the other hand, aren't all e2e test already checked through unit tests in theory (if you have 100% coverage)?

skidding:
Thanks @NiGhTTraX!

cypress/integration/local-state_spec.js
NiGhTTraX: Even though your unit tests cover unit A and unit B, there could be assumptions made on either side that become out of date (think mocks). e2e tests cover for that. They should also cover the app's composition root where you wire up the dependencies. In either case, they shouldn't check for correctness. They should only check the broad requirements.

Issue #287, opened 2017-02-10
Make __fixtures__ directory configurable


ovidiubute:
### What's up?

The `__fixtures__` directory cannot be changed from any configuration, even if you can provide additional directories to scan for fixtures, there is no way to have fixtures inside your `components` directory and call it something else. I've seen some users confused by this behavior and I believe we should provide an option to override it with whatever they like. This could also accommodate large projects that already have their own fixture structure and would never change it just to use `react-cosmos`.

### Mkay, tell me more...

Just off the top of my head, we could have a field in `cosmos.config.js` (naming could be better):

```
{
  inlineFixturesPath: '__fixtures__' (default option, can be overridden by user config)
}
```

This is going to non trivial impact, as it's used in quite a few places:

- `react-cosmos-voyager`: [here](https://github.com/react-cosmos/react-cosmos/blob/da85709af02f4ddf5b8d9908c52678f802a9f214/packages/react-cosmos-voyager/src/index.js#L6), [here](https://github.com/react-cosmos/react-cosmos/blob/ca471cb75cd1dab556a3ea68fcf3864e62015fd1/packages/react-cosmos-voyager/src/match-fixture-path.js#L2), [here](https://github.com/react-cosmos/react-cosmos/blob/da85709af02f4ddf5b8d9908c52678f802a9f214/packages/react-cosmos-voyager/scripts/generate-use-cases.js#L11), some tests as well.
- `flatris`: [here](https://github.com/react-cosmos/react-cosmos/blob/da85709af02f4ddf5b8d9908c52678f802a9f214/examples/flatris/browserify/prepare-modules.js#L27)

skidding:
> even if you can provide additional directories to scan for fixtures, there is no way to have fixtures inside your components directory and call it something else.

You should be able to achieve this using the `getFixturePathsForComponent` option, which is not documented:

```js
// cosmos.config.js

getFixturePathsForComponent: (componentName) => {
  return glob.sync(`src/components/custom-fixture-dir/${componentName}/**/*.js`);
}
// or, if you keep fixtures per component
getFixturePathsForComponent: (componentName) => {
  return glob.sync(`src/components/${componentName}/custom-fixture-dir/**/*.js`);
}
```

Wdyt? Shall we add some docs for this option? :) 
[...]

PR #368, opened 2017-06-02
add fixturesDir configuration option

maciej-ka:
https://github.com/react-cosmos/react-cosmos/issues/287
Enables to change fixtures directory name in configuration, default value is `__fixtures__`.

### DIFF
--- a/packages/react-cosmos-voyager/scripts/generate-use-cases.js
+++ b/packages/react-cosmos-voyager/scripts/generate-use-cases.js
@@ -62,6 +62,16 @@ const useCases = {
     pkgs/nested/Baz/src/index.jsx
     pkgs/nested/Baz/fixtures/blank.js
   `),
+  'fixtures-dir-setting': (`
+    components/Foo.js
+    components/nested/Bar.jsx
+    components/nested/Baz.jsx
+    components/__fixtures__/Foo/blank.js
+    components/__fixtures__/nested/Bar/one.js
+    components/__fixtures__/nested/Bar/two.json
+    components/__fixtures__/nested/Bar/three.jsx
+    components/__fixtures__/nested/Baz/blank.js
+  `),
 };
 
 const useCasePath = path.join(__dirname, '../src/use-cases');
--- a/packages/react-cosmos-voyager/scripts/generate-use-cases.js
+++ b/packages/react-cosmos-voyager/scripts/generate-use-cases.js
@@ -66,11 +66,11 @@ const useCases = {
     components/Foo.js
     components/nested/Bar.jsx
     components/nested/Baz.jsx
-    components/__fixtures__/Foo/blank.js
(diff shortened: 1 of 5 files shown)


skidding:
@maciej-ka thank you, this looks good!

I wonder why didn't Travis got triggered on this PR. Looking at #353 and #332 it seems like it should work out outside PRs already. @ganderzz @jnishiyama did you do anything extra to ensure Travis ran on your fork PRs?

jnishiyama:
@skidding Naw, Didn't do anything extra. Why don't you or someone else try making the PR to attempt to isolate the issue?

skidding:
@maciej-ka how about adding a unit test for this? Tests for `react-cosmos-voyager` have a twist to them but are then very fun to write.

[Here](https://github.com/react-cosmos/react-cosmos/blob/10402ea8afdf626713f993fbcd1832649678e091/packages/react-cosmos-voyager/scripts/generate-use-cases.js#L6-L65) are the use cases. After you add another use case to that map, you go to `packages/react-cosmos-voyager` and run `yarn generate-use-cases`. This will write the corresponding files to the disk. Then in the tests file you add a use case to match those files. I'd start from a simple example like [relative-fixtures](https://github.com/react-cosmos/react-cosmos/blob/10402ea8afdf626713f993fbcd1832649678e091/packages/react-cosmos-voyager/src/__tests__/index.js#L47-L69).

maciej-ka:
I was thinking about adding them too, big thanks for instructions

packages/react-cosmos-voyager/scripts/generate-use-cases.js
```diff
@@ -62,6 +62,16 @@ const useCases = {
     pkgs/nested/Baz/src/index.jsx
     pkgs/nested/Baz/fixtures/blank.js
   `),
+  'fixtures-dir-setting': (`
+    components/Foo.js
+    components/nested/Bar.jsx
+    components/nested/Baz.jsx
+    components/__fixtures__/Foo/blank.js
```
skidding: Shouldn't this be `components/non-default-fixtures-dir/Foo/blank.js`

ganderzz:
@skidding Late to the party, but no, didn't have to do anything to trigger Travis. A newer PR has a build, so seems to be isolated to this PR..?

packages/react-cosmos-voyager/scripts/generate-use-cases.js
maciej-ka: Right, should be fixed now. I changed generated folder and forgot to update here.

skidding:
This is great. Will try it out locally and merge it this weekend.

@maciej-ka maybe you need to have a Travis account (login via Github) in order for Travis to trigger builds when you open a PR.

skidding:
Beautiful! ✨

PR #381, opened 2017-06-28
Upgrade deps June

skidding:


skidding:
For some reason Cypress fails (`type` command in search input no longer triggers change event in FixtureList component, so the fixtures don't get filtered) due to some minor/patch package upgrades. See 724a17114f05386beacdf125e8f7909b73d8c1bf for the diff (from broken to passing). package.json is identical.

skidding:
Never mind. I though I had downgraded React but the caret has failed me again, `@^15.5.4` installed `15.6.1` 😡

It's probably [this regression bug](https://github.com/facebook/react/issues/10030).

PR #383, opened 2017-06-30
[WIP] Component page

maciej-ka:
**Component page**
- [X] Allow component to be selectable in nav
- [X] Display a component page
- [ ] Create frameset for each fixture
- [ ] Insert different components into framesets
- [ ] Display legend

partially closes: https://github.com/react-cosmos/react-cosmos/issues/314


maciej-ka:
@skidding I tried to run `yarn run test-smoke-ci` locally and result was the same error as on Travis: ```You passed the --record flag but did not provide us your Record Key.``` Can I do something for it to pass?

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -97,8 +101,17 @@ export default class FixtureList extends Component {
...
+              [styles.componentSelected]:
+                component === urlParams.component &&
+                !urlParams.fixture,
+            });
             return (
-              <div key={i} className={styles.component}>
+              <div
+                key={i}
+                className={componentClassNames}
+                onClick={() => this.onComponentClick(component)}
```
skidding: Wouldn't it make more sense to put the onClick handler on the `componentName` div? This way you don't need `stopPropagation` anymore as the fixture buttons will no longer be contained in the component button.

packages/react-cosmos-component-playground/src/components/FixtureList/index.less
```diff
@@ -87,6 +87,15 @@
   }
 }
 
+.componentSelected .componentName {
```
skidding: It might look nicer to nest these

```less
.componentSelected {
  .componentName {
```

packages/react-cosmos-component-playground/src/components/LoadersGrid/index.jsx
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react';
+
+class LoadersGrid extends Component {
```
skidding: How about naming this something like `ComponentPage`? This page could do many things in the future:
- If there are no fixtures it can be a landing page for that component
- Could display component meta data like name, prop types and description extracted from source comments (e.g. JSDoc)
- Could display a grid of loaders or a list of loaders, a "view modes"

maciej-ka: Thanks for bringing this, I did feel need to talk about it. After your proposal I see two different sets of responsibility:

LoadersGrid
---
To `LoadersGrid` you say: "here is a list of [component, fixture, label] triples" and it will:
- create many iframes
- take care of communication with them (send fixtures)
- layout those iframes and display labels
- provide an interface for changing from horizontal list to grid view

Would `LoadersManager` be a better name?

It would be an abstract tool for handling group views, situations where we render more than one component. It would be needed again on "project page" with all components rendered using their default fixture (also described in https://github.com/react-cosmos/react-cosmos/issues/314)

ComponentPage
---

I see a need for `ComponentPage` also. It would:
- prepare data for `LoadersGrid` [{component, fixture, description}, ...]
- handle information display for a component without fixtures
- host other component page on functionalities in the future

skidding: Excellent! 

I like `LoadersGrid` more than `LoadersManager`, although I'd opt for `LoaderGrid` (singular) as "grid" already suggests a list of items.

maciej-ka:
@skidding I could use a little help:
1. where in source files is html that iframe points to
1. what is flow of control from a moment we render iframe to sending fixture data

skidding:
> 1.  where in source files is html that iframe points to

The iframe loads the index.html that is part of the user bundle. If the user doesn't specify a webpackConfigPath. [This](https://github.com/react-cosmos/react-cosmos/blob/f0edb47bee29ed1f00461c5d404690e8e2313331/packages/react-cosmos-webpack/src/default-webpack-config.js#L24-L26) default config is used, which serves the index.html via HtmlWebpackPlugin.

[This section](https://github.com/react-cosmos/react-cosmos/blob/393-webpack-plugins/README.md#configuring-indexhtml) explains how the index.html is configured in iframe contents (aka Loader).

> 2.  what is flow of control from a moment we render iframe to sending fixture data

In the latest code the fixture data is actually sent from inside the iframe to the playground (the other way around). Inside the iframe – the Loader – is part of the user bundle, which has all component and fixtures. Here is the code that runs when the iframe renders: [react-cosmos-loader/src/mount.js](https://github.com/react-cosmos/react-cosmos/blob/f0edb47bee29ed1f00461c5d404690e8e2313331/packages/react-cosmos-loader/src/mount.js#L17-L39)

Here is where the Loader sends the fixture contents to the Playground (outer frame): [react-cosmos-loader/src/components/RemoteLoader/index.jsx](https://github.com/react-cosmos/react-cosmos/blob/f0edb47bee29ed1f00461c5d404690e8e2313331/packages/react-cosmos-loader/src/components/RemoteLoader/index.jsx#L153-L157)

Take a look at this updated list of events and let me know if it makes sense (or it there's any terminology that you don't understand): [Order of events](https://github.com/react-cosmos/react-cosmos/blob/master/CONTRIBUTING.md#order-of-events)

image

Thanks!

skidding:
Hi @maciej-ka, I see you updated your branch one day ago. I just merged some styling PRs (#409 and #410) so you have conflicts again – Sorry! 🙈

maciej-ka:
That's nothing to worry. However, my problem is I cannot pass `yarn` anymore. Tried with few node, yarn and npm downgraded versions.

Error is: `lerna ERR! execute error An unexpected error occurred: "https://registry.yarnpkg.com/@skidding/react-codemirror/-/react-codemirror-1.0.1.tgz: Request failed \"404 Not Found\"".`

```> yarn --version && node --version && npm --version
0.23.4
v6.11.0
5.3.0
```

skidding:
That is annoying. It's right [here](https://www.npmjs.com/package/@skidding/react-codemirror). Anyway, https://registry.yarnpkg.com/@skidding/react-codemirror/-/react-codemirror-1.0.1.tgz doesn't 404 for me, so maybe it works now?

maciej-ka:
Sadly it's still the same. I have this problem for around three weeks, all that time link was correct, I tried many versions of yarn between 0.23 and 0.28, `yarn cache clean`, `rm -rf node_modules`, `yarn upgrade`, node 6 and 8. Tried to install on a fresh clone of Cosmos repo. I have it on both of my computers. So frustrating.

skidding:
That's messed up. I just did a fresh install and it works, and Travis also works so it's definitely not a global thing. 

You can just replace `@skidding/react-codemirror` with `react-codemirror` in your package.json and not commit it–my patch fixes a bug that only affects the fixture editor which you wouldn't use in your work. We wouldn't even need the fork if https://github.com/JedWatson/react-codemirror/pull/107 got merged...

maciej-ka:
Thanks! Build is passing now. I'll change `ComponentPlayground.state.fixtureData` into structure `ComponentPlayground.state.fixturesData[component][fixtureName]` to have bodies of several fixtures on one page and see what next move I can do.

maciej-ka:
In case anyone has this install error too:
`lerna ERR! execute error An unexpected error occurred: "https://registry.yarnpkg.com/@skidding/react-codemirror/-/react-codemirror-1.0.1.tgz: Request failed \"404 Not Found\"".`

Downgrade to yarn 0.23.4 and rm `~/.yarnrc` may be solution (it was for me).

maciej-ka:
@skidding `ComponentPlayground` stores many fixture bodies in a structure now, organized by component and fixture name. While I was working on it, I realized, that this change maybe wasn't necessary, as those fixture bodies are sent to the editor. The visible component gets a fixture body from Loader.

skidding:
> While I was working on it, I realized, that this change maybe wasn't necessary, as those fixture bodies are sent to the editor. The visible component gets a fixture body from Loader.

That's true... Unless we want to show an editor for each fixture, which I don't think we do, we don't need to store the fixture bodies in the component page. Sorry for not saying something sooner!

maciej-ka:
Dear and patient @skidding,

Recent `fixtureBodies` are reverted. Loader grid and separate component are removed too.

I wish the Playground would be able:
a. to scan a list of components and fixtures without a Loader.
b. to create iframes with syntax `<CosmosFrame component={...} fixtureBody={...} />`.

Since we plan to have more than one Loader on a page, it's probably sound to move scanning to a different place.

Whatever we change with iframes, I assume that initiating `<iframe src={...}>` will take some time if src is pointing to a remote server. Because of that, unused iframes are valuable and should be not removed from dom, but hidden.

I restarted coding with an idea of Loader pool with this scenario in mind: 
1. I am on a single fixture page, one iframe with loader is created. 
2. I go to a component page with five fixtures. The previous iframe is reused plus four more are created.
3. I go back to single fixture page. Four frames are hidden, one is reused and visible.

I'm very interested in your opinion.

And, traditionally, I have few questions: 
- where is a line of code that creates RemoteLoader component in an iframe?
- where is a line of code that reads fixture body?
- would it be possible to give iframes some kind of id, that they would report back when sending loader ready event?

skidding:
Hi @maciej-ka,

First of all, I want you to know that this PR is going to be monumental, so right back at you when it comes to patience! :)

> I wish the Playground would be able:
a. to scan a list of components and fixtures without a Loader.

Yes and no. The idea is this: The Playground must stay decoupled from the user codebase (aka Loader build). The reason for this is to enable:

1. User to use different version of React than the Playground component
2. Future: User to use different view layer (something else than React) than the Playground component (see [Goals](https://github.com/react-cosmos/react-cosmos/blob/fb512e6f3ca48ad28afb65b72bd73cab4ab41905/CONTRIBUTING.md#goals))

In other words, think of the Playground as a remote control.

That said, I just came up with a radical idea while writing this reply. **How about this?**

The whole idea of instantiating a Loader to receive the list of fixtures seems wrong, because if you have X Loaders, you need to have like a "master" Loader. Ugh. I think we agree so far. So if the Playground is a "remote control", not tied to the user bundle, how can it elegantly receive the list of fixtures only once? Maybe it doesn't receive them from the Loader. After all, the Loader is meant to do much more: Render proxies + component and then report back on state changes.

Soo, if all we need is a large JSON object, why not talk to the Cosmos server instead? Cosmos already starts an Express server. We can attach a new method, something like `localhost:8989/_fixtures`, to return the complete list of component fixtures. The Playground could do a fetch request to there method first, add the data to its state (as it does now from the `loaderReady` event), and then carry on as usual.

What do you think?

PS. I will happily answer your specific questions as well after we decide on the core idea above.

PPS. Regarding the loader pool iframe reusability thing: I'm not a big fan of that optimization to be honest. Feels like we're reinventing the Virtual DOM 😄. We can worry about optimizations later if it really makes sense, but iframes are cheap and for now we need to fix this architectural conundrum.

skidding:
Even more futuristic: 
- We could attach Socket.IO to our Express server and keep a web socket connection open
- Once we have a socket connection open, we can move the entire [Playground ⇆ Loader communication](https://github.com/react-cosmos/react-cosmos/blob/master/docs/playground-loader.md) from window.postMessage to web sockets. This works even when we have more loaders open at the same time if we assigned an ID to each loader frame, like you suggested.

Why would this be even better? Going back to the `#UniversalComponentPlayground` goal, this could enable something like a React Native integration, where the Loader isn't an iframe in the same page with the Playground, but runs in a completely different client/platform.

Of course this is cool stuff to think about, but for now a new GET method on the existing server is *much* more realistic.

maciej-ka:
I have a son! (born this weekend) :). I'll be back to this branch around one week. `/_fixtures` endpoint is next task and we say bye bye to loader pool, hiding the one in current source misled me that they are precious.

One more question to add to above:
Under which user interaction Playground `fixtureUpdate` and `fixtureListUpdate` messages are triggered from Loader?

skidding:
> I have a son! (born this weekend) :). 

That is great! Congratulations! 👶🎉

> /_fixtures endpoint is next task and we say bye bye to loader pool


> hiding the one in current source misled me that they are precious.

The reason why it's currently hidden is UX, so the screen doesn't flicker when you switch between fixtures. Not that I think of it your idea was in line with this 🤔, so good thinking! Might be easier to keep the main loader hidden for ever and have a different grid of loaders for component page, but I'm not ruling out your initial idea, especially after we have the basics in place.

> Under which user interaction Playground fixtureUpdate and fixtureListUpdate messages are triggered from Loader?

- `fixtureUpdate` comes from state changes captured by proxies. So component state change, or app state changes (when using Redux proxy), or URL changes (when using React Router proxy), etc. This would continue to be received from the Loader.
- `fixtureListUpdate` is tricker, and I've been thinking about this intensively last night. The fixture list can change if a component or fixture file is added/renamed/changed on disk. The Loader [reloads on webpack HMR events](https://github.com/react-cosmos/react-cosmos/blob/fb512e6f3ca48ad28afb65b72bd73cab4ab41905/packages/react-cosmos-webpack/src/loader-entry.js#L43), which generates `fixtureListUpdate` event in the end, [updating the Playground](https://github.com/react-cosmos/react-cosmos/blob/fb512e6f3ca48ad28afb65b72bd73cab4ab41905/packages/react-cosmos-loader/src/components/RemoteLoader/index.js#L107-L110) with the latest components and fixtures. If we move to a new endpoint, like we discussed, we'll lose this and we'll have to restart the playground whenever adding a new fixture file. So it seems like a websocket connection as opposed to a plain endpoint might be required after all. The problem is we still need to watch the file system and effectively duplicate what webpack is doing, so I'm not yet sure of the best solution here. I have a few more thoughts on this but we should probably sync in greater detail before you invest more time on this. I'd happily talk more on Slack, including having a call to speed up the conversation.

skidding:
So, I've been thinking about this some more. @maciej-ka I wish I could've guided you better from the beginning, but unfortunately this feature is outside the scope of the original design. Now, we can hack it, which is possible in more than one way I'm sure, or we can lay the foundation for a better, more flexible architecture. I would choose the latter.

Cosmos v2 has already entered feature freeze, and I'm planing a release this month. Anything new is either 2.1 or 3 if needed. So no rush to get this out the door if we're not comfortable with the outcome.

Unless we come up with a better idea (feel free to propose), a web socket connect still sounds like the best approach. Here is a dump of things I imagine connected to this idea:
- New package: react-cosmos-server. Node package that receives a cosmos.config, creates an Express server with a socket.io connection. It absorts a lot of what react-cosmos-webpack does now (minus the sockets), meaning that react-cosmos-webpack imports newly created react-cosmos-server, starts it and attaches the webpack compiler to it
- The socket.io client is present in both the playground and loader bundles. With this in place, the current window.postMessage communication turns into socket events. Existing events can be reused with little change.
- If the communication becomes centralized, this means there will be two types of messages: global messages (like "fixtureListUpdate" when the fixture files change) and Loader messages (like "select fixture"). This brings us back to something you mentioned, which is assigning ID to Loaders, so that each Loader message contains a loaderId.
- Speaking of global messages, another thing that could later be implemented is opening source files into user's editor when clicking on a button in the playground (via `opn` node package, which would be called by the server). I already hacked a prototype of this and it feels amazing to click next to a component and have it pop up in the default code editor. This functionality would fit perfectly with the socket server.

But, this is A LOT, and it's probably going to be a while until you get to actually finish the component page you started with. It's the only clean solution that I can imagine, though, and maybe you find it interesting as well. If you do, you should break this mammoth into pieces and implement it one by one. I'm also pretty sure there's more to it than what I described, so think about it deeply and ask anything.

If this plan actually comes to life, tho, it will have the highest impact as it opens many other doors.

Take care!

skidding:
I left this open so long for posterity and for sentimental reasons, but it's time to move on. The PR is obviously outdated. As for the solution, some things to consider:

- The concept of "component" is almost opaque to Cosmos nowadays. I realized over the years that we often want to put more than one component in a fixture, and just as often the root component is not the most important component, but merely a decorator. Which is why fixtures are now just an element tree, which allows enough flexibility to build a custom "component page".
- I do envision a way to select multiple fixtures in split screen mode (up to 4), and to potentially bookmark a multi selection. But this is only partly related.

Thank you @maciej-ka for all your work on this!

Issue #387, opened 2017-07-04
All fixtures matched under component with same name as project


flaviusone:
### What's up?

All fixtures are matched under a component that has the same name as the project (found in absolute path).

### Mkay, tell me more...

This helps recreate the failing scenario:
```
{
  base: '/Users/foo/PROJECTNAME/tests/fixtures/components/PROJECTNAME/base.js',
  filtered: '/Users/foo/PROJECTNAME/src/scripts/components/somecomponent/__fixtures__/filtered.js',
}
```

Failing condition: https://github.com/react-cosmos/react-cosmos/blob/a2f5a2f3ab1ae1d82675c3ab663103883d118405/packages/react-cosmos-voyager/src/match-fixture-path.js#L5

skidding:
Fixed by #398

PR #389, opened 2017-07-05
[WIP] Create blank fixtures for components

jlc467:
With https://github.com/react-cosmos/react-cosmos/pull/364 (merged), blank fixtures are no longer provided automatically.

**Requirements are still a WIP, but so far, this PR:**
Passes the result of `getComponentsWithoutFixtures` to `createBlankFixturesForComponents`, which creates an `empty.js` fixture file for each component without a fixture.

@skidding has described the end goal as
> In a pretty future, when the user visits the playground, he or she is greeted with "There are 3 new components since you last visited, would you like to create blank fixtures for them?". Then, the user checks some components from the list and submits.

Here's a mockup for review:
screen shot 2017-07-10 at 9 04 35 am

### DIFF
--- a/examples/local-state/package.json
+++ b/examples/local-state/package.json
@@ -4,7 +4,8 @@
   "repository": "https://github.com/react-cosmos/react-cosmos/tree/master/examples/local-state",
   "license": "MIT",
   "scripts": {
-    "start": "node ../../packages/react-cosmos-webpack/bin/cosmos.js"
+    "start": "node ../../packages/react-cosmos-webpack/bin/cosmos.js",
+    "create-fixtures": "node ../../packages/react-cosmos-webpack/bin/cosmos-create-fixtures.js"
   },
   "xo": false
 }
--- a/packages/react-cosmos-voyager/package.json
+++ b/packages/react-cosmos-voyager/package.json
@@ -19,7 +19,8 @@
     "touch": "^1.0.0"
   },
   "dependencies": {
-    "glob": "^7.1.1"
+    "glob": "^7.1.1",
+    "react-cosmos-config": "^2.0.0-beta.15"
   },
   "xo": false
 }
--- a/packages/react-cosmos-voyager/package.json
+++ b/packages/react-cosmos-voyager/package.json
@@ -19,7 +19,9 @@
     "touch": "^1.0.0"
   },
   "dependencies": {
+    "fs-extra": "^3.0.1",
     "glob": "^7.1.1",
+    "lodash.find": "^4.6.0",
     "react-cosmos-config": "^2.0.0-beta.15"
   },
   "xo": false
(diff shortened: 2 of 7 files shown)


skidding:
Hey. Sorry it took so long to reply. I'm looking for an apartment these days :)

I love the sketch, it looks very promising. You're right about the external paths, but you know what, I'd rather we don't support that use case since it's there for historical reasons (old codebases) and the flow is complicated enough. If user wants to create a fixture via this flow it will go in `__fixtures__`. What do you think?

We still have some complexity, because you can have `components/__fixtures__/Button/standard.js` and `components/Button/__fixtures__/standard.js`. I think we have to support these 2 forever because from what I saw only ~50% of users create a folder for each component.

The code looks good, but `getFixtureFolder` is a bit messy. It's maybe WIP, but I can't say I understand it completely after reading it 3 times. The truth is, paths in cosmos are a bit of a mess. I believe `react-cosmos-voyager` has great tests, but the implementation ([this](https://github.com/react-cosmos/react-cosmos/blob/a2f5a2f3ab1ae1d82675c3ab663103883d118405/packages/react-cosmos-voyager/src/index.js#L63-L95) and [this](https://github.com/react-cosmos/react-cosmos/blob/a2f5a2f3ab1ae1d82675c3ab663103883d118405/packages/react-cosmos-voyager/src/match-fixture-path.js#L3-L7)) is convoluted. This embarrassing bug is also proof: https://github.com/react-cosmos/react-cosmos/issues/387.

Please let me know if you have any ideas on how to simplify this. We should probably take some time to write a paragraph or two about how paths work in the CONTRIBUTING guide while we're at it. Finally, part of the paths confusion is also created by the fact that `getComponentName` is only called when componentPaths contains file paths and not dirs. Maybe we can clean up this situation.

jlc467:
> If user wants to create a fixture via this flow it will go in __fixtures__
Works for me. For some reason I thought external fixtures folder was a new feature, but I was actually confusing `fixturesDir` and `fixturePaths`

`getFixtureFolder` is definitely a WIP. The goal is determine where to create the fixture. I might try to have `getFilePaths` return the components with missing fixtures, along with the path(s) the fixtures were looked for in. That way, create-fixtures does not need `getFixturesFolder` and is always creating fixtures where they will be found. This will require studying paths in voyager a little more, and I'll make comments as I go. The final approach may be to leave `getFilePaths` alone, but reuse some of the functions for finding fixtures as a way to get the fixture folder. Will play with it some more. Thanks!

skidding:
> For some reason I thought external fixtures folder was a new feature, but I was actually confusing `fixturesDir` and `fixturePaths`

Yeah, naming is part of the confusion. `componentPath` can also be a path where components are found or a path to a specific component 😕

skidding:
@jlc467 GitHub is a bit confusing me, are you expecting a review on this or did the review request just remained there from a previous time? In any case, I'd be curious to find out your opinion on this proposal: https://github.com/react-cosmos/react-cosmos/issues/440 – It ties in into this and am hoping to find a future-proof solution for component discovery/configuration.

jlc467:
Will open new pull request with new approach to fixture generation at some point in the future.

PR #391, opened 2017-07-07
Link examples via Lerna

### LINKED ISSUE #390: Link packages in examples via Lerna
### What's up?

Examples use relative paths to use Cosmos packages (e.g. `../../../packages/react-cosmos-webpack/bin/cosmos.js`). This is ugly and different than a real life examples. Making examples Lerna packages will allow linking packages normally and make the examples closer to standalone ones.

### Mkay, tell me more...

Also
- [x] Document monorepo structure
- [x] Add example using DllPLugin https://webpack.js.org/plugins/dll-plugin/

skidding:
Closes #390


skidding:
@flaviusone finally added a DllPlugin example. It works as expected!

PS. Added a verbose example to become familiar with DllPlugin, but fyi there's now [AutoDllPlugin](https://github.com/asfktz/autodll-webpack-plugin) with makes it much easier.

CONTRIBUTING.md
```diff
@@ -12,79 +12,107 @@ Moreover, React Cosmos tries to answer the question: **How to design components
...
-React Cosmos is a [monorepo](packages) with many small packages. This makes it possible to design isolated units, but also to build and test all packages together (inside one of the [examples](examples)).
+### Monorepo
+
+React Cosmos is a [monorepo](packages) powered by [Lerna](https://github.com/lerna/lerna). The monorepo structure makes it possible to publish independent packages while being able to have end-to-end tests for the whole project, and to easily link examples to unpublished packages.
+
+While the monorepo structure is powerful, it also creates limitations. Important to note:
+
+- **Test and build tools are installed globally in the root node_modules.** This includes Jest, Babel, webpack, and their corresponding plugins and loaders. ESLint is also applied globally.
+
+- **React and webpack deps are *hoisted* in the root node_modules.** The linked packages and examples are sibling directories, so they each have a separate node_modules dir. In real life, however, when Cosmos packages are installed in a user codebase, they all share a common node_modules. We simulate the real life scenario by deduplicate these dependencies.
```
jlc467: Noob question here -- I'm not at all familiar with Lerna, so I checked out the [docs for hoist](https://github.com/lerna/lerna/blob/master/doc/hoist.md) which describe it as a flag (`--hoist`). However,  I can't see that flag set anywhere in cosmos. Not sure why the flag would be needed anyway, since, by default, node looks in the root folder's node_modules for packages?

jlc467:
I tried the dll example and it seems to work.

CONTRIBUTING.md
skidding: > Noob question here -- I'm not at all familiar with Lerna, so I checked out the [docs for hoist](https://github.com/lerna/lerna/blob/master/doc/hoist.md) which describe it as a flag (`--hoist`). However, I can't see that flag set anywhere in cosmos.

We're not using the hoist option because we hoisted these deps manually by just adding them to the root [package.json](https://github.com/react-cosmos/react-cosmos/blob/a2f5a2f3ab1ae1d82675c3ab663103883d118405/package.json#L49-L50). I think this is simpler than adding them to every package and then relying on Lerna to hoist them in the root. I wonder if it helps to specify this as others might have the same question.

> Not sure why the flag would be needed anyway, since, by default, node looks in the root folder's node_modules for packages? I guess it depends where you are running the example from.

The flag is needed if you prefer `react` to be in `packages/foo/package.json` but not in root `package.json`. But yeah, this is why I didn't complicate matters by using that option and just installed it once, globally. Otherwise whenever we want to upgrade React we go in every package do to so.

flaviusone:
Also tested DllPlugin in our project and worked. LGTM.

skidding:
@jlc467 removed the word "hoisted" to avoid confusion with the `hoist` Lerna option: https://github.com/react-cosmos/react-cosmos/pull/391/commits/2242ab3cac5fb82abae3bf175c0eb5925381cf7c

PR #392, opened 2017-07-11
Mention NODE_ENV in README

skidding:


skidding:
cc @maxsalven

jlc467:
👍

Somewhat related... Even though we use module.exports in CRA cosmos.config example, do you think [this line](https://github.com/react-cosmos/react-cosmos/pull/392/files?diff=unified#diff-04c6e90faac2675aa89e2176d2eec7d8R236) would confuse someone?
**Currently:**
Define `.babelrc` to enable ES6 in config & fixtures, using the CRA preset.
**Proposing:**
Define `.babelrc` with the CRA preset to enable ES6 for fixtures and components.

I meant to address this in https://github.com/react-cosmos/react-cosmos/pull/379. CRA is so plug and play that I'm not sure people are taking note of the plugins included in the preset (myself included).

skidding:
Good point!

But now I wonder why do we need `.babelrc` altogether. [This](https://github.com/react-cosmos/react-cosmos#using-babel-node)  is connected. `react-cosmos-webpack` (server side) benefits from it. But it only reads cosmos.config.js and optionally the user's webpack config. The components and fixtures are ran using the user's webpack loaders. 🤔

skidding:
@jlc467 [just tested this](https://github.com/skidding/flatris/pull/24/files) in the Flatris repo and it seems like `.babelrc` was only needed for `cosmos.config.js` to use ES modules, which we already stopped encouraging. I also removed the entire section on webpack 2 because it doesn't seem relevant to Cosmos. Wdyt?

jlc467:
@skidding Think I understand. CRA's webpack config [already defines the babel preset to use](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.dev.js#L177), so the .babelrc isn't required. 

As for removing webpack 2 section -- it wasn't saying much beyond making sure to set `{ "modules": false }` if you are using the es2015 preset, but users can decide on their own which preset they want to use.

PR #394, opened 2017-07-13
Don't include HotModuleReplacement plugin twice

### LINKED ISSUE #393: Prevent duplicate HotModuleReplacementPlugin
### What's up?

Webpack HMR is enabled by default in Cosmos, but users usually already have `HotModuleReplacementPlugin` added to their webpack config. Unless we pass `hmrPlugin: false` in cosmos.config, Cosmos blindly duplicates it, causing a build error with an unclear error message. I've seen this many times.

### Mkay, tell me more...

- [x] We can use [merge.unique](https://github.com/survivejs/webpack-merge#mergeuniquefield-fields-field--field) from webpack-merge! Or just check for `plugin.constructor && plugin.constructor.name` manually.

Bonus:
- [x] Set `output.filename` to `'[name].js'` (makes Cosmos webpack config work when user uses CommonsChunkPlugin)
- [x] Create example with CommonsChunkPlugin using [manifest](https://webpack.js.org/plugins/commons-chunk-plugin/#manifest-file)
- [ ] Update React Redux Starter Kit docs (integration is much more smooth now)

```js
// Plugins
new webpack.optimize.CommonsChunkPlugin({ names: ['manifest'] })
```

Inspiration: [react-styleguidist](https://github.com/styleguidist/react-styleguidist/blob/f7876503e5a070c3de08c032727107e7802298f6/scripts/utils/mergeWebpackConfig.js#L17-L33)

skidding:
Also:
- [x] Don't break when user config has CommonsChunkPlugin

Closes #393


skidding:
- No more duplicate HMR plugin @NiGhTTraX 
- The instructions for integrating with React Redux Starter Kit are much smaller @Yustynn

Issue #346, opened 2017-04-28
Source map doesn't work


dfedynich:
Hello, guys. I feel pleasure while using cosmos to debug my React components like in real sand-box. Great DX tool indeed!

However I found one issue, which decreases the speed of my debugging session a lot. I use devtool **'source-map'** in my webpack.config.js and it seems that everything goes as expected.

![image](https://cloud.githubusercontent.com/assets/7403902/25530756/11680c26-2c30-11e7-9ae0-cb8e1aab8ce6.png)

However when I mark a debug point in my component's source code individual file, it doesn't fire. Could you give me any tips, why the source map doesn't work?

skidding:
Did you by any chance check out [this thread](https://github.com/webpack/webpack/issues/2145#issuecomment-294361203)? Source maps with Webpack can be problematic, but it seems like `cheap-module-source-map` & `cheap-module-inline-source-map` are better alternatives so maybe you should give them a try. Please report your findings if you do. Thanks!

dfedynich:
Hey, @skidding, thank you for the link with comparison of devtools for webpack. If we combine these knowledge with [webpack official](https://webpack.github.io/docs/configuration.html#devtool), we will get a real combo. However, I didn't manage to achieve a success - still can't fire breakpoints.

Here is my relevant webpack.config.js file
```
'use strict';

var merge = require('webpack-merge');
var devConfig = require('./webpack.dev');
var HtmlWebpackPlugin = require('html-webpack-plugin');

var config = merge(devConfig, {
    devtool: '#cheap-module-source-map',
    plugins: [
        new HtmlWebpackPlugin({
            filename: './index.html'
        })
    ]
});

module.exports = config;
```

Here is my cosmos.config.js file
```
module.exports = {
    componentPaths: [
           //my path to the components
    ],

    getComponentName: function (componentPath) {
        return componentPath.match(/components\/(.+)\.js$/)[1]
    },

    globalImports: [
        './vendor',
        './main'
    ],

    proxies: [
        'react-cosmos-normalize-props-proxy',
        './redux-proxy'
    ],

    hot: true,
    hmrPlugin: true,

    hostname: 'localhost',
    port: 8989,
    webpackConfigPath: './webpack.config.js'
};
```

Do you have any thoughts, why breakpoints are not fired?

david-ogalego:
I am having the same problem. Is it possible that as the bundle generated, is being loaded twice, the browser does not know how to link the sourcemap?

[...]

PR #395, opened 2017-07-15
Update and add new examples

skidding:
@jlc467 this is a tiny one. I just shortened the name of the example you created (it was way longer than the others) and update some package.json fields.


skidding:
Also added an example with source maps configured, to test if #346 is still a problem. All works well on my end
image

cc @dfedynich @david-ogalego

jlc467:
Source maps example working on my end too. Was looking for the `.map` file in the network tab... ended up finding it in `chrome://net-internals/#events`

PR #396, opened 2017-07-18
Refresh docs

skidding:


jlc467:
On the topic of exporting, the playground seems to use root-relative links, resulting in:
screen shot 2017-07-18 at 6 11 46 pm

If I change to relative, e.g. 
screen shot 2017-07-18 at 6 16 27 pm

everything "just works" again.

If this needs a code fix, I can open an issue and address is separately. Otherwise, it might be worth adding the need for a web server to the documentation. What do you think?

skidding:
> If this needs a code fix, I can open an issue and address is separately.

Definitely. That would be great. I think it's good UX to make it work without a web server, if possible.

skidding:
**See new README & CONTRIBUTING guide on [refresh-docs branch](https://github.com/react-cosmos/react-cosmos/tree/refresh-docs).** Any feedback appreciated!

cc @jlc467 @flaviusone @ovidiubute @RadValentin @catalinmiron @NiGhTTraX @bogdanjsx @maciej-ka @sapegin @tkjone @maxsalven

skidding:
Merging this to be able to build on the new docs in #402. Feel free to add comments later.

Issue #382, opened 2017-06-29
should read nested components fixtures


amrdraz:
It seems to me that no fixture files are read beyond the first component level.

I have a project architecture that looks like this
<img width="288" alt="screen shot 2017-06-30 at 00 26 55" src="https://user-images.githubusercontent.com/1708559/27713221-dda7ed62-5d2a-11e7-8492-41ef4c836bfb.png">
But while cosmos can detect the components in the project the playground looks like this
<img width="383" alt="screen shot 2017-06-30 at 00 24 32" src="https://user-images.githubusercontent.com/1708559/27713191-b3014e28-5d2a-11e7-93c8-c189873065ad.png">

it seems to me that the fixture files are not included

I tried modifying the cosmos.config.js file to no avail

```
const glob = require('glob')
module.exports = {
  componentPaths: ['src/components'],
  // componentPaths: glob.sync('src/{,**/}components'),
  // getComponentName: path => path.match(/src\/components\/(?:.+\/components\/)(.+)\.jsx?$/),
  fixturePaths: glob.sync('src/{,**/}components/{,**/}__fixtures__/'),
  ...
}
```

When I tried `fixturePaths: glob.sync('src/{,**/}components'),` I got index.js (the component index not the fixtures)

skidding:
Hey @amrdraz. Can you please try this?

```
components/Board/index.js
components/Board/__fixtures__/initial.js
components/Board/components/BoardSquare/index.js
components/Board/components/BoardSquare/__fixtures__/even.js
components/Board/components/BoardSquare/__fixtures__/odd.js
components/Board/components/BoardSquare/components/Square/index.js
components/Board/components/BoardSquare/components/Square/__fixtures__/foo.js
components/Board/components/Knight/index.js
components/Board/components/Knight/__fixtures__/bar.js
```

amrdraz:
Greate that works, (for some reason I couldn't start the project so I started a new one)

But I found a small problem

It seems as though all files under __fixtures__ in the same component get loaded, so in this example the slide fixture is being read twice under `BannerCarousel`

<img width="284" alt="screen shot 2017-07-21 at 13 37 47" src="https://user-images.githubusercontent.com/1708559/28462098-f8195406-6e19-11e7-9c29-d9ccf8f53c30.png">

Inspecting the code I realized the reason is that `getMatchingFixtures`matches the `componentName` against all available fixture paths

and since the nested component by definition included the parent component
[...]

PR #398, opened 2017-07-21
only consider the first __fixture__ directory under a component folder

amrdraz:
referencing issue #382

### DIFF
--- a/packages/react-cosmos-voyager/src/__tests__/index.js
+++ b/packages/react-cosmos-voyager/src/__tests__/index.js
@@ -195,6 +195,7 @@ testUseCase({
   }
 });
 
+// should throw depreication warning
 testUseCase({
   useCase: 'separate-packages',
   input: {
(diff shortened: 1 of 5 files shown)


packages/react-cosmos-voyager/src/index.js
```diff
@@ -67,6 +67,7 @@ const getFilePaths = ({
 
       const componentDir = path.dirname(componentPath);
       const componentName = getComponentName(componentPath);
+      console.log(componentName);
```
skidding: Forgot this

packages/react-cosmos-voyager/src/match-fixture-path.js
```diff
@@ -2,7 +2,7 @@ import { FIXTURE_EXTENSIONS_REGEX } from './fixture-extensions';
 
 const buildPathMatchers = (componentName, fixturesDir) => [
   `/${fixturesDir}/${componentName}/([^/]+)\\.(${FIXTURE_EXTENSIONS_REGEX})$`,
-  `/${componentName}/(?:.+/)?${fixturesDir}/([^/]+)\\.(${FIXTURE_EXTENSIONS_REGEX})$`,
+  `/${componentName}/${fixturesDir}/([^/]+)\\.(${FIXTURE_EXTENSIONS_REGEX})$`,
```
skidding: This wouldn't be very pretty, but what if we also added this rule?
```js
`/${componentName}/src/${fixturesDir}/([^/]+)\\.(${FIXTURE_EXTENSIONS_REGEX})$`,
```
This would keep the previous test passing (which is a real life scenario used by large code base using Cosmos), and still fix the problem. The only way this can backfire is if somebody names a component as lowercase "src", and nests it under another component. It's an edge case we can live with until we find a better solution to refactor this part.

amrdraz:
@skidding done

packages/react-cosmos-voyager/src/__tests__/index.js
```diff
@@ -168,6 +195,7 @@ testUseCase({
   }
 });
 
+// should throw depreication warning
```
skidding: @amrdraz one more 😂

amrdraz: 😆

skidding:
Thanks @amrdraz! 👍 

Published under `2.0.0-beta.19`, please check if the new version works as expected.

Now that you are a contributor, you might want to take a look at [the docs rewrite PR](https://github.com/react-cosmos/react-cosmos/pull/396#issuecomment-317355617).

PR #402, opened 2017-07-29
XHR and fetch proxies 👯

### LINKED ISSUE #253: Backend mock middleware
### Problem
- If you have a component based on Redux and, for example, your component contains a Button that when clicked it triggers a Redux action that performs a backend request,
- Then you cannot reliably use the Playground to test your component
- Two scenarios will happen:
  - Either your code contains checks for a network error and your component is fine
  - Or not 😢 

### Proposed Solution
- Introduce a `mock-backend` middleware (proxy in `react-cosmos` terms)
- Write what you need inside the fixture, I'm thinking something very close to [`superagent-mock`](https://github.com/M6Web/superagent-mock#usage)
- When Playground starts, it starts a mock server, that serves the data configured in the fixture

### Issues
- User code would need to be configurable to perform the network requests towards Playground (i.e. no hard-coded hostnames, no authentication, and be able to perform requests locally)
- Adds complexity to the fixture

skidding:
Fixes #253


skidding:
This is a long time request. Please review!

skidding:
@zaboco you might also be interested. Turns out `fetch-mock` was the way to go after all :)

packages/react-cosmos-xhr-proxy/src/__tests__/mocking.jsx
```diff
@@ -0,0 +1,162 @@
...
+        fixture={{
+          foo: 'bar',
+          xhr: [
+            {
+              url: '/users',
+              response: (req, res) =>
+                res.status(200).body([{ name: 'John' }, { name: 'Jessica' }]),
+            },
+            {
+              url: '/user',
```
ovidiubute: Does this match `/user*` or just `/user`?

skidding: Just `/user` I think. It just passes it to [xhr-mock](https://github.com/jameslnewell/xhr-mock#api). 

fetch-mock seems more flexible in this regard:
image

Ideally both xhr and fetch requests would be mocked in the same way, but merging the APIs would've been risky business so I went for the simple approach–just use their own API and avoid maintenance costs.

skidding:
Published under `2.0.0-beta.20`

PR #403, opened 2017-07-31
Jumps back to fixture in CP

adelinaenache:


packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -33,28 +34,42 @@ const getFilteredFixtures = (fixtures, searchText) => {
...
+      window.removeEventListener('keydown', this.onWindowKey);
+    }
+
+    onWindowKey = e => {
+      const isFocused = this.searchInput === document.activeElement;
 
-    if (e.keyCode === KEY_S && !isFocused) {
-      // Prevent entering `s` in the search field along with focusing
-      e.preventDefault();
+      if (e.keyCode === KEY_S && !isFocused) {
```
skidding: You must've indented these lines by mistake. Please remove one level of indentation and **make sure `yarn run lint` doesn't fail** – you can see the build is failing because of this:

image
image

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -33,28 +34,42 @@ const getFilteredFixtures = (fixtures, searchText) => {
   }, {});
 };
 
+const isFixtureSelected = (fixtures, component, fixture) => {
```
skidding: `isExistingFixtureSelected` would be a better name because we check if a fixture is selected *and* if it exists.

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -73,6 +88,7 @@ export default class FixtureList extends Component {
     e.preventDefault();
 
     this.props.onUrlChange(e.currentTarget.href);
+
```
skidding: No need for this line.

skidding:
@adelinaenache it seemed like you're not going to finish this, so I recreated your commit on a different branch (under your name and email address) and added the finishing touches. https://github.com/react-cosmos/react-cosmos/pull/439

Issue #397, opened 2017-07-19
Cosmos export navigation should work without webserver


jlc467:
### What's up?
The playground seems to use root-relative links, resulting in:
![screen shot 2017-07-18 at 6 11 46 pm](https://user-images.githubusercontent.com/6405557/28342216-29aba820-6be5-11e7-9021-aa972eeaedeb.png)
when loading an exported playground.

### Mkay, tell me more...

If I change to relative, e.g. 
![screen shot 2017-07-18 at 6 16 27 pm](https://user-images.githubusercontent.com/6405557/28342234-4057de9a-6be5-11e7-9220-abf827edf150.png)
navigation works again.

PR #404, opened 2017-08-07
Support custom URL for static assets

### LINKED ISSUE #297: Improve publicPath

skidding:
Closes #297

Also on this branch:
- [x] Export static assets as well
- [x] Relative paths inside CP #397
- [x] Fix CP UI in Firefox
- [x] ~~Fix CP in Safari~~ It's not perfect (fixture editor sometimes overlaps with Loader frame), but it's outside of scope and *much* better than before

### DIFF
--- a/packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
+++ b/packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
@@ -308,7 +308,7 @@ export default class ComponentPlayground extends Component {
               <a
                 ref="homeButton"
                 className={homeClassNames}
-                href="/"
+                href="?"
                 onClick={router.routeLink}
               >
                 <HomeIcon />
@@ -319,7 +319,7 @@ export default class ComponentPlayground extends Component {
                 <a
                   ref="fixtureEditorButton"
                   className={fixtureEditorClassNames}
-                  href={`/${fixtureEditorUrl}`}
+                  href={`${fixtureEditorUrl}`}
                   onClick={router.routeLink}
                 >
                   <CodeIcon />
@@ -328,7 +328,7 @@ export default class ComponentPlayground extends Component {
                 <a
                   ref="fullScreenButton"
                   className={styles.button}
-                  href={`/${fullScreenUrl}`}
+                  href={`${fullScreenUrl}`}
                   onClick={router.routeLink}
                 >
                   <FullScreenIcon />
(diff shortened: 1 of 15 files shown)


skidding:
@jlc467 bundled a fix for #397 as well, let me know if this is what you also had in mind

packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
```diff
@@ -319,7 +319,7 @@ export default class ComponentPlayground extends Component {
                 <a
                   ref="fixtureEditorButton"
                   className={fixtureEditorClassNames}
-                  href={`/${fixtureEditorUrl}`}
+                  href={`${fixtureEditorUrl}`}
```
ovidiubute: How about `href={fixtureEditorUrl}`

packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
```diff
@@ -328,7 +328,7 @@ export default class ComponentPlayground extends Component {
                 <a
                   ref="fullScreenButton"
                   className={styles.button}
-                  href={`/${fullScreenUrl}`}
+                  href={`${fullScreenUrl}`}
```
ovidiubute: `href={fullScreenUrl}`

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
```diff
@@ -120,7 +122,7 @@ export default class FixtureList extends Component {
                       <a
                         key={j}
                         className={fixtureClassNames}
-                        href={`/${uri.stringifyParams(nextUrlParams)}`}
+                        href={`${uri.stringifyParams(nextUrlParams)}`}
```
ovidiubute: Unnecessary String interpolation.

packages/react-cosmos-component-playground/src/components/ComponentPlayground/index.jsx
skidding: 😆 I just removed the slash without thinking 🐑🐑🐑 

Thanks!

packages/react-cosmos-component-playground/src/components/FixtureList/index.jsx
skidding: Fixed!

PR #414, opened 2017-08-24
Fix create-react-app example cosmos.config.js

tkjone:
In the `create-react-app` section of the documentation, there is an example config that looks like this:

```
module.exports = {
   componentPaths: ['src/components'],
  // ... 
}
```
When I ran the above in a new `create-react-app` I found that webpack had a compilation error due to the fact that new `create-react-app`s do not have a `components` sub directory.  The fix for me was this:

```
module.exports = {
   componentPaths: ['src'],
  // ... 
}
```

The next issue was the console complaining that `it` was not defined, so also needed to add an ignore:

```
module.exports = {
   componentPaths: ['src'],
   ignore: [/App.test.js/]
  // ... 
}
```

Another thing I noticed is that the the above is going to show `registerServiceWorker` and `index` as component folders...a step further would be to update the config with an ignore like this:

```
module.exports = {
  componentPaths: ['src'],
  containerQuerySelector: '#root',
  webpackConfigPath: 'react-scripts/config/webpack.config.dev',
  publicPath: 'public',
  ignore: [/App.test.js/, /registerServiceWorker.js/, /index.js/]
};
```

The issue I have with the above is that without context it could be confusing for new users using `create-react-app` to test `cosmos` for the first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ It's preferred to use CRA's own webpack config (instead of duplicating it).
 ```js
 // cosmos.config.js
 module.exports = {
-  componentPaths: ['src/components'],
+  componentPaths: ['src'],
   containerQuerySelector: '#root',
   webpackConfigPath: 'react-scripts/config/webpack.config.dev',
   publicPath: 'public'
--- a/README.md
+++ b/README.md
@@ -305,11 +305,10 @@ It's preferred to use CRA's own webpack config (instead of duplicating it).
 ```js
 // cosmos.config.js
 module.exports = {
-  componentPaths: ['src'],
+  componentPaths: ['src/components'],
   containerQuerySelector: '#root',
   webpackConfigPath: 'react-scripts/config/webpack.config.dev',
-  publicPath: 'public',
-  ignore: [/App.test.js/]
+  publicPath: 'public'
 };
 ```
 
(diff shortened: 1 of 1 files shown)


skidding:
Good catch!

I don't think setting the componentPaths to `src` is the best approach, because you're targeting *all* the user source (utils, constants, reducers and whatnot)–which is why you have to start ignoring stuff like `index.js` and `registerServiceWorker.js`.

How about adding another bullet to "Also make sure to", saying something like

> Create the `src/components` directory and place your components there, or change `componentPaths` option to match your existing structure

It's not ideal, but I still wouldn't include the entire app source in `componentPaths` as it's meant to target only component files. 

Beyond documentation, we should make sure the Cosmos isn't so fragile and it doesn't crash when component paths are missing. A check would be useful (probably [here](https://github.com/react-cosmos/react-cosmos/blob/299fc740426ac08b8dcdb31017fb84a1b0e3ec40/packages/react-cosmos-voyager/src/index.js#L75)) warning the user that a path isn't valid, allowing the rest of the program to move on. If in the end Cosmos will find 0 components the UI onboarding will take if from there. We already have a "you have no components yet" screen.

tkjone:
Sounds good.  Updated the `also make sure to` section with your suggestions.

skidding:
Thank you, sir! 🙇

Issue #415, opened 2017-08-25
Error: ENAMETOOLONG with cosmos-export


NiGhTTraX:
### What's up?

```
node_modules/.bin/cosmos-export --config playground/cosmos.config.js
[Cosmos] Using webpack config found at src/webpack.config.dev.js
node_modules/fs-extra/lib/mkdirs/mkdirs-sync.js:49
          throw err0
          ^

[...]

PR #416, opened 2017-08-25
Prevent infinite loop when exporting public path #415

skidding:


packages/react-cosmos-webpack/src/export.js
```diff
@@ -77,8 +77,16 @@ module.exports = function startExport() {
   // Copy static files first, so that the built index.html overrides the its
   // template file (in case the static assets are served from the root path)
   if (publicPath) {
-    const exportPublicPath = path.join(outputPath, publicUrl);
-    fs.copySync(publicPath, exportPublicPath);
+    if (outputPath.indexOf(publicPath) === -1) {
```
NiGhTTraX: I'm not sure if this will solve the case where `publicPath` is `./`. Might need to use `path.resolve` first, or `path.join`.

Anyway, check my [comment](https://github.com/react-cosmos/react-cosmos/issues/415#issuecomment-325118975) surrounding `publicPath`. Maybe clearer docs would be better than this guard.

skidding: The paths are resolved and absolute at this point, so it works. This is the output you've gotten in react-test-buffet with this patch:

image

skidding: The docs already describe `publicPath` as

> Where to serve static files from. Like --content-base in webpack-dev-server.

But any suggestion is welcome!

NiGhTTraX: Ah, sweet.

PR #417, opened 2017-08-25
Create a new Apollo Proxy

xavcz:
> Executed pairing with @skidding in Berlin with smiles & 🍻

What's up?
---
- Add support for a mocked Apollo Client and its React provider component as a proxy 🚀 
- Add an example using this proxy
- Babel switched from `es2015` to `env` cause `graphql` need generators function (transpiled from async/await)


xavcz:
Due for 3 weeks, docs added & code updated to fit `beta-28` 🌮

packages/react-cosmos-apollo-proxy/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "react-cosmos-apollo-proxy",
+  "version": "2.0.0-beta.23",
```
skidding: I think you need to make this `2.0.0-beta.28` for the tests to pass. The example references 2.0.0-beta.28 so it tries to download it from npm instead of linking from the monorepo.

xavcz: Oh, yes! Thanks for the heads up!

examples/apollo/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "2.0.0-beta.28",
+  "description": "Example of mocking the Apollo Client",
+  "repository": "https://github.com/react-cosmos/react-cosmos/tree/master/examples/apollo",
+  "license": "MIT",
+  "scripts": {
+    "start": "NODE_ENV=development cosmos"
+  },
+  "dependencies": {
+    "react-apollo": "^1.4.15",
+    "react-cosmos-utils": "^2.0.0-beta.28",
```
skidding: Sorry, mate! This should also be `2.0.0-beta.24`. Only packages that have changes get bumped on publish, so while latest is `2.0.0-beta.28`, lower level packages sometimes stay behind.

xavcz: Haha, no probs!

PR #422, opened 2017-09-04
Improved proxy config API

### LINKED ISSUE #420: Improved proxy config API
### What's up?

Based on feedback from a number of people, I've found existing way to configure proxies to be confusing and not pleasant to use. 

The confusion stems from the fact that all other options are configured in `cosmos.config.js` yet proxies need to be placed in their own file, even if they only have 2-3 lines of code. There's a good explanation for this, but the reason isn't obvious and the current solution isn't great.

After some brainstorming with @xavcz, he proposed an alternative way of configuring proxies which makes more sense (that's also backwards compatible).

### Mkay, tell me more...

First of all, why can't proxy code be inlined in `cosmos.config.js`? 

Because `cosmos.config.js` is meant to be imported on the server, and proxy code is client side code (loaded in the user bundle). If we imported cosmos.config in the client bundle then any Node-specific import like `fs` or `path` would corrupt the bundle and overall make Cosmos fragile.

So, given that the proxy config needs to be in a separate file than the main config, the proposal is to **keep all proxies in the same file instead of having one file per proxy.** This makes sense because proxy config tends to be small, while the number of proxies can get large in mature projects.

### Meet `cosmos.proxies.js`

Instead of defining `proxies` in cosmos.config.js, we'll have `proxiesPath`, which defaults to `./cosmos.proxies.js` so in most cases it won't need to be defined.

Here's an example of how cosmos.proxies.js would look like:

```js
import createReduxProxy from 'react-cosmos-redux-proxy';
// We can import anything from our codebase in this file
import configureStore from '../configureStore';

const ReduxProxy = createReduxProxy({
  createStore: state => configureStore(state)
});

const MyCustomProxy = props => {
  const { value: NextProxy, next } = props.nextProxy;
  return (
    <FooProvider>
      <NextProxy {...props} nextProxy={next()} />
    </FooProvider>
  );
};

// We can still ensure a specific proxy order
export default [
  ReduxProxy,
  // No need to import proxies if we rely on defaults
  'react-cosmos-router-proxy',
  'react-cosmos-fetch-proxy',
  MyCustomProxy
];
```

Just one file to define/mock the entire component context!

This solution has two other side benefits:
- This file can be easily imported and reused in other places like unit/integration tests, where we also need proxies to create component context as part of the test setup.
- This format is a bit more flexible, no need to return each proxy as a function any more (e.g. `() => createReduxProxy({ ... })`)

What do you think? @tkjone @xavcz @ovidiubute

skidding:
Closes #420 

Notes:
- Docs for fixtures and proxies have been revamped, please [take a look](https://github.com/react-cosmos/react-cosmos/tree/420-improved-proxy-api#fixtures) and provide feedback
- Unfortunately I couldn't implement the exact API proposed originally. All proxies need to be imported inside `cosmos.config.js`, instead of just returning a module name in the proxy list. Doing the latter would've meant static manipulation and would've 2x or 3x the work. Importing actually makes more sense because the code can be statically read by bundlers and linters. Please see latest docs for the minor change in API, it's still pretty sleek.
- Also due to time constraints, I immediately deprecated `config.proxies` instead of implementing backwards compatibility. The reasoning was that it's less work to assist anyone in need then making both ways work and maintaining them going forward. Not a lot of Cosmos users use proxies (yet!) from what I know. I will make a big announcement on Slack before publishing and in case your package manager decides to upgrade automatically (tho lock files should prevent) I added a clear warning message with a link to the new docs.
- Internal stuff: The Loader used to normalize ES modules (component, fixtures, proxies) which didn't make a lot of sense so I lifted that logic up into the webpack entry point. 

Thanks @xavcz and @tkjone for all your help on this one!


README.md
```diff
@@ -116,7 +124,7 @@ Run `npm run cosmos` or `yarn cosmos` and go to [localhost:8989](http://localhos
 
 #### What's a fixture?
 
-A fixture is a component mock. An example of component input. The input can be props, state, or–with the help of [proxies](#proxies)–anything a component depends upon (e.g. API responses).
+A fixture a JS object used to mock component input and external dependencies. The input can be [props](#props), [children](#children), [state](#state) and [context](#context). With the help of [proxies](#proxies), fixtures can mock anything else a component depends on, from API responses to localStorage to window size.
```
xavcz: The verb jumped out of the sentence in the refactor! "A fixture is a JS object..." 

Love this explanation of fixture

README.md
```diff
@@ -21,30 +21,35 @@
...
 
-Cosmos scans your project for components and loads them inside [Component Playground](https://react-cosmos.github.io/),
-enabling you to:
+Cosmos scans your project for components and enables you to:
 
 1. Render components under any combination of props, context and state
-2. See state evolve in real-time while interacting with running
+2. Mock *every* external dependency (eg. API responses, localStorage, etc)
+3. See app state evolve in real-time while interacting with running
 instances
```
xavcz: Should it be on the same line as 3.?

skidding: Cheers!

examples/axios/cosmos.proxies.js
```diff
@@ -0,0 +1,3 @@
+import createXhrProxy from 'react-cosmos-xhr-proxy';
+
+export default [createXhrProxy()];
```
ovidiubute: Would it be too much trouble to support exporting a single proxy without wrapping it in an Array? I imagine someone will open an issue about trying that in the future.

packages/react-cosmos-webpack/src/normalize-modules.js
```diff
@@ -0,0 +1,30 @@
...
+  Object.keys(components).reduce(
+    (acc, next) => ({
+      ...acc,
+      [next]: importComponent(components[next], next)
+    }),
+    {}
+  );
+
+export const normalizeFixtures = fixtures =>
+  Object.keys(fixtures).reduce((acc, next) => {
```
ovidiubute: ❤️  how "functional" this function is 😄

examples/react-router/cosmos.proxies.js
```diff
@@ -0,0 +1,3 @@
+import createReactRouterProxy from 'react-cosmos-router-proxy';
+
+export default [createReactRouterProxy()];
```
ovidiubute: As a user I would love to see an example with multiple proxies.

examples/axios/cosmos.proxies.js
skidding: Should be fairly simple, something like `Array.isArray(proxies) ? proxies : [proxies]` somewhere in the codebase. My only worry if it doesn't add confusion when newcomes read the examples and in some places `cosmos.proxies.js` returns a list, while in others just a proxy. Maybe they'll think it's two different things. Also, my assumption is that any mature codebase will rely on more than one proxy. Any similar examples for other projects?

examples/react-router/cosmos.proxies.js
skidding: Good point, will add.

packages/react-cosmos-webpack/src/normalize-modules.js
skidding: Not really proud of `acc2` and `next2` tho 😅

examples/axios/cosmos.proxies.js
ovidiubute: Maybe it's just weird that all examples have only one proxy. You're right this is already a pretty advanced topic 👍

skidding: Hmm, what about the [first example](https://github.com/react-cosmos/react-cosmos/tree/420-improved-proxy-api#where-to-put-proxies)?

skidding: Rewrote the first example to have better annotations and to feature most popular proxies: Redux, React Router and Fetch.

skidding: Also with a link to the proxy boilerplate.

image

Issue #251, opened 2017-01-02
Collapsible and smarter component list


ovidiubute:
### Problem
If you have more than 10 components in your app, with at least 2 or 3 fixtures each, the component list is pretty daunting to look at. To be able to find something you need to use the search feature, but this should not be the only way to find something.

### Proposed solution
- Make the component list expandable, i.e. each item in the list that is a component name should have an icon next to it (or maybe the item itself should be clickable) that triggers the collapse/expand of its fixture list. The fixture list should animate when opening/closing.
- When you start the Playground for the first time, all components should be collapsed by default except the first one with at least 1 user-defined fixture (first component that has more fixtures than just (no-props auto))
- As you use Playground, the fixtures most often used could be tracked via local storage. Thus, when reloading the Playground, instead of having all of the fixtures collapsed, the most recently or most commonly used components (to be discussed) could have their fixture list expanded.

ovidiubute:
@skidding please review

skidding:
Yes!

> When you start the Playground for the first time, all components should be collapsed by default except the first one with at least 1 user-defined fixture

Not sure about this. Maybe if we do a good job with oboarding #252 this isn't needed.

> As you use Playground, the fixtures most often used could be tracked via local storage. Thus, when reloading the Playground, instead of having all of the fixtures collapsed, the most recently or most commonly used components (to be discussed) could have their fixture list expanded.

❤️ it. I think the easiest implementation is to simply remember the last expanded components (list of name strings) between visits.

Extra thoughts:
- The CP is known to be a monster component. Now is a good time to extract a Nav component out of it. Besides the isolation of concerns, writing tests in the old suite is painful (it was designed poorly) and this is an opportunity to rewrite the nav UI part and make it an ES6 class tested with Jest.
- One think I find misleading about the current search mechanism is how the selected component remains visible regardless of the search input. I think it should only show the matched fixtures and "no results" when there aren't any. Completely optional, but maybe we could bundle this improvement as well (unless you feel otherwise).

skidding:
One more thing: It would be cool if we would separate components with fixtures from components without fixtures, in this order. The benefits are twofold:
- You see the components you have created fixtures for first
- The list buttons can diff between sections. E.g. components without fixtures shouldn't be collapsable and instead of having a "no props (auto)" fixture the component name itself could be the button.

jlc467:
[...]

PR #423, opened 2017-09-07
added tree view and integrated it into component playground

amrdraz:
This pull requests settles issue #251 

currently I'm writing tests for the FixtureList component, but I would like some feedback on the current design

there were things I could have done more like remembering the state of each node in the tree propr to search but after trying it out it felt unnecessary 

@skidding  please review ux and any design recomendations

as requested instead of using a tree-view with lots of dependencies I built a simple light tree view and integrated under the FixtureList component

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "xo.enable": true
+}
--- a/packages/react-cosmos-component-playground/package.json
+++ b/packages/react-cosmos-component-playground/package.json
@@ -8,6 +8,9 @@
     "lib"
   ],
   "main": "lib/index.js",
+  "scripts": {
+    "cosmos": "NODE_ENV=development cosmos"
+  },
   "devDependencies": {
     "@skidding/react-codemirror": "^1.0.1",
     "classnames": "^2.2.5",
@@ -21,8 +24,10 @@
     "react": "^15.5.4",
     "react-cosmos-loader": "^2.0.0-beta.24",
     "react-cosmos-state-proxy": "^2.0.0-beta.24",
+    "react-cosmos-webpack": "^2.0.0-beta.25",
     "react-dom": "^15.5.4",
-    "react-querystring-router": "^2.0.0-beta.23"
+    "react-querystring-router": "^2.0.0-beta.23",
+    "react-test-renderer": "^15.6.1"
   },
   "xo": false
 }
--- a/packages/react-cosmos-component-playground/cosmos.config.js
+++ b/packages/react-cosmos-component-playground/cosmos.config.js
@@ -1,4 +1,5 @@
 export default {
   componentPaths: ['src/components'],
+  ignore: [/__test__/, 'icons', 'service'],
   globalImports: ['src/utils/global.less']
 };
--- a/packages/react-cosmos-component-playground/src/components/FixtureList/__tests__/__snapshots__/index.js.snap
+++ b/packages/react-cosmos-component-playground/src/components/FixtureList/__tests__/__snapshots__/index.js.snap
@@ -1,0 +1,637 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Tree Fixtrue View should load Tree with propulated items  1`] = `
+<div
+  className="root"
+>
+  <div
+    className="searchInputContainer"
+  >
+    <input
+      className="searchInput"
+      onChange={[Function]}
+      placeholder="Search..."
+      value=""
+    />
+    <svg
+      viewBox="0 0 24 24"
+    >
+      <path
+        d="M15.5 14h-.79l-.28-.27C15.41 12.59 16 11.11 16 9.5 16 5.91 13.09 3 9.5 3S3 5.91 3 9.5 5.91 16 9.5 16c1.61 0 3.09-.59 4.23-1.57l.27.28v.79l5 4.99L20.49 19l-4.99-5zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"
+      />
+    </svg>
+  </div>
+  <div
+    className="nodeTree"
+  >
+    <div
+      className="node"
+    >
+      <button
+        className="nodeNameContainer isNode "
+        data-path="ComponentA"
+        onClick={[Function]}
+        title="ComponentA"
+      >
+        <div
+          className="nodeArrow"
+        >
+          <svg
+            height="10px"
+            viewBox="0 0 10 10"
+            width="10px"
+          >
+            <path
+              d="M 5,4.5 l0, 4.5, l-4,0z"
+              fill="#eeeeee"
+              stroke="#eeeeee"
+            />
+          </svg>
+        </div>
+        <div
+          className="nodeIcon"
+        >
+          <svg
+            height="15px"
+            viewBox="0 0 20 20"
+            width="18px"
+          >
+            <path
+              d="M 0,2 l2,0 l 6,0 l2,2 l10,0 l0,15 l-20,0z"
+              fill="#aaaaaa"
+            />
+          </svg>
+        </div>
+        <div
+          className="nodeName"
+        >
+          ComponentA
+        </div>
+      </button>
+      <div
+        className="nodeChildren"
+      >
+        <div
+          className="node"
+        >
+          <button
+            className="nodeNameContainer isLeaf "
+            data-path="ComponentA/foo"
+            onClick={[Function]}
+            title="ComponentA/foo"
+          >
+            <div
+              className="nodeArrow"
+            />
+            <div
+              className="nodeIcon"
+            >
+              <div
+                style={
+                  Object {
+                    "backgroundColor": "rgb(243, 204, 0)",
+                    "color": "#333333",
+                    "fontSize": "9px",
+                    "height": "16px",
+                    "left": "1px",
+                    "position": "relative",
+                    "width": "16px",
+                  }
+                }
+              >
+                <span
+                  style={
+                    Object {
+                      "bottom": 0,
+                      "position": "absolute",
+                      "right": 1,
+                    }
+                  }
+                >
+                  JS
+                </span>
+              </div>
+            </div>
+            <div
+              className="nodeName"
+            >
+              foo
+            </div>
+          </button>
+        </div>
+        <div
+          className="node"
+        >
+          <button
+            className="nodeNameContainer isLeaf "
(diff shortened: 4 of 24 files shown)


skidding:
First of all, playing with this is fun! Good job getting this to work!

Here are my thoughts:

Is `/components/ComponentA/components/ComponentB/` a known pattern? I find it repetitive because we keep adding "components" dirs. Doesn't look that good in the the playground either. How about putting the Tree component at the same level with FixtureList? Or is it possible to get rid of the "components" altogether and have `components/ComponentA/ComponentB`? I'm starting to see the value of nesting them, especially in the tree view...

I couldn't read all the code yet, so I'll have to come back to it. I like that you didn't introduce a bloated lib and kept it minimal, but it's quite a lot of code for me to parse so can you please write a small description of the new abstractions (components, classes, etc) you added and what each does? It will be very useful in guiding me when reading the code (but also useful to add as documentation via comments).

One thing I noticed is that you keep derived data into state, like the `filteredFixtures` and `selectedPath`, which is an anti pattern. A core tenet of React is to keep state minimal, and anything that can be derived should be computed on render. If it ever becomes an optimization problem we can memoize. So, unless you have a reason I'm not aware of I'd like to go back to using [this pattern](https://facebook.github.io/react/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state).

Regarding the design. Can we use the existing colors, spacings and hover please? The folder icon as well, it looks friendlier.
image

The new design is nice to play with, but has a bit of Win98 feel to it :)
image

I know the tree view changes the markup, but here's an example from Atom where the selection bg goes all the way from left to right even in a nested structure:
image
Let's use the same.

The yellow `js` icons looks cool on the Cosmos blue, but it gets redundant after a while. In Atom you have many types of files, so it makes more sense (although I hid the icons anyway), but here *all* non-folder items are fixtures, so I say we keep them clean.

Let me know if there's anything I can help with. I'll probably add some inline comments as well after this.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "xo.enable": true
```
skidding: No editor specific config files please. `.vscode` can go in your global gitignore.

packages/react-cosmos-component-playground/package.json
```diff
@@ -8,6 +8,9 @@
     "lib"
   ],
   "main": "lib/index.js",
+  "scripts": {
+    "cosmos": "NODE_ENV=development cosmos"
```
skidding: No need for this, you can already run `yarn run start:playground` from the root project dir.

packages/react-cosmos-component-playground/package.json
```diff
@@ -21,8 +24,10 @@
     "react": "^15.5.4",
     "react-cosmos-loader": "^2.0.0-beta.24",
     "react-cosmos-state-proxy": "^2.0.0-beta.24",
+    "react-cosmos-webpack": "^2.0.0-beta.25",
     "react-dom": "^15.5.4",
-    "react-querystring-router": "^2.0.0-beta.23"
+    "react-querystring-router": "^2.0.0-beta.23",
+    "react-test-renderer": "^15.6.1"
```
skidding: No need for this, test dependencies are installed globally in the root package.json to not have to add them in every package.

amrdraz:
`/components/ComponentA/components/ComponentB/` is pattern I picked up [from this post](https://medium.com/@alexmngn/how-to-better-organize-your-react-applications-2fd3ea1920f1). But I agree that while it helps keep things clean when I have other type of files that are specific to the component (reducer, actions, services etc..), it gets repetitive but. I'm not too attached to it, plus I try to conform to the projects conventions, I mainly kept it to keep a deeper tree when testing. (was even thinking of making the tree ignore it XD but then that would have invited all sorts of trouble)

----

The code is mostly tests actually there are 3 files that were introduced and a modification to the FixtureList to adapt the change.

We have the `Tree` component and its `TreeService` and we have a subcomponent of Tree called `Node` which renders each node in the tree. 

**Node**

The Node component takes props

```js
{
  name: string, // the name that appears next to the folder
  path: string, // the path up until this point in the tree used in title and to trigger actions on this node
 open: boolean, // if this node is not a lead then if true it will show the children and the down arrow
 active: boolean, // is true when the node is selected
 isLeaf: boolean, // is true when node path matches the path used to construct this node and the path ends with a value
 value: any, // a value that is set for the leaf at the end of the path can be anything
 onLeafClick: (node)=>{}, // a function that is fired when a leaf is clicked
 onNodeClick: (node)=>{}, // a function that is fired when a non leaf is clicked
 treeService: TreeService, // an instance of the TreeService used to select and toggleOpen
}
```
instance methods

```js
toggleOpen() // opens the node ats props.path
selectNode() // selects the node at props.path
```

**Tree**

initializes the TreeService and subscribes to it's state
updates the TreeService with changes occur to props that may change the state

can take the following props
```js
{
  fixtures: object //  an object map of a path to the components with and array of fixture names used
  paths: object // a processed paths object with path to fixture and value to be set on the leaf
  tree: object // an initial tree state
  selectedPath: string // path to be selected after the tree has been constructed
  onLeafClick: (node)=>{} // a function that is fired when a leaf is clicked passed the node as argument
}
```

When rendering the tree will sort leafs at the beginning in order to place the fixtures right under the folder then subcomponents can show

**TreeService**

This is where the tree's state is managed, it can be initialized with a tree state, a paths object, or a fixtures object

The tree is constructed using a paths object of the form `{ [path]: value }`
for each path in paths, a node is created for every fragment of the path
The node with the last fragment in the path has the value attached to it.

all nodes are initially open

> a path fragment array is the results of splitting the path on `/` (with this be a cross platform issue?)

if initialized with fixtures, fixtures are converted to the paths object form

```js
{
  `componentA`: [`fixture1`, `fixture2`],
  `componentA/B`: [`fixture1`],
  `componentC`: []
}
```
becomes
```
{
  `componentA/fixture1`: {
     component: 'componentA',
     fixture: 'fixture1'
   },
  `componentA/fixture2`: {
     component: 'componentA',
     fixture: 'fixture2'
   },
  `componentA/B/fixture1`: {
     component: 'componentA/B',
     fixture: 'fixture1'
   },
   `componentA/B`:  undefined
}
```

the tree is immutable and always updates the path the changed

**TreeService.select(path, multi)**

set the node at the end of the path to `active:true`

multi is a toggle option for multi path select, it is not needed, I was mostly playing around :P 

**TreeService.toggleOpen(path)**

toggle the `open` property of the node at the end of the path

I will update the files to reflect documentation
------

regarding the state upon reflection I agree there is no need to retain the fixture or selectedPath in the state I will move them back to be computed in render.

-----

I thought of a means to highlight the background and can change the folder icon
I personally prefer having something to anchor my eye on regarding the fixture icon, but I'm not attached to it.
I do not agree completely with regards to the spacing between elements, the font felt a bit large when in the tree view.

Will apply updates on Monday night I think

skidding:
> /components/ComponentA/components/ComponentB/ is pattern I picked up from this post.

Thanks for sharing the post. I read it and it I like it. And thanks for being flexible. I prefer using a flat structure in the CP for now to focus on the tree view functionality itself in this PR. You can open a subsequent PR to convert the entire CP package structure to the component/service pattern and link to the article in the contributing guide.

CP doesn't have many files, but I'm more interested in how well #440 works with this pattern. It would make a great case study. Maybe even add a section to README/Usage, saying how the config should look like if you use that pattern, similar to _Integration with popular tools_. BUT, let's focus on this PR.

>  I mainly kept it to keep a deeper tree when testing.

Makes perfect sense! Feel free to keep it like this until we finish the review process.

> Node

- Given the above, how about naming this `TreeNode` for now?
- `active: boolean`, how about `selected`? It's more suggestive
- I only see `onNodeClick` passed in tests. If you inspect the components you see no node receives `onNodeClick` – Is this WIP?
  image

Besides components and the tree class, it seems like we introduce _types_ so let's define them. In the future it will allow us to use something like Flow or TS, but for now let's document them to organize data and keep a clear image in our heads.

Like you said, we start from `fixtures`:

```js
{
  `componentA`: [`fixture1`, `fixture2`],
  `componentA/B`: [`fixture1`],
  `componentC`: []
}
```

Then we create an intermediary `paths` object:

```js
{
  `componentA/fixture1`: {
     component: 'componentA',
     fixture: 'fixture1'
   },
  `componentA/fixture2`: {
     component: 'componentA',
     fixture: 'fixture2'
   },
  `componentA/B/fixture1`: {
     component: 'componentA/B',
     fixture: 'fixture1'
   },
   `componentA/C`:  undefined
}
```

We need to take #383 into account. If we don't prepare for it, we'll have to refactor soon. In the future we'll be able to select components as well. So we'll have non-leaf nodes that are also selectable. Maybe this?

```js
{
  `componentA/B`: {
    component: 'componentA/B'
  },
  `componentA/fixture1`: {
    component: 'componentA',
    fixture: 'fixture1'
  },
  `componentA/fixture2`: {
    component: 'componentA',
    fixture: 'fixture2'
  },
  `componentA/B`: {
    component: 'componentA/B'
  },
  `componentA/B/fixture1`: {
    component: 'componentA/B',
    fixture: 'fixture1'
  },
  `componentA/C`: {
    component: 'componentA/C'
  }
}
```

Then we have the `TreeNode` type, which I think is the most important one. We must prevent inconsistent and redundant data.
image
For example we might know a node is a leaf by checking if it has children. Please correct me if I'm wrong.

Proposal:
```js
// TreeNode type
{
  children: [] // list of TreeNodes, always Array
  open: true, // always boolean
  selected: false, // always boolean
  name: 'components',
  value: {}
}
// isLeaf = children.length > 0
```

There's a lot of state duplication. The Tree service stores paths and the tree, but then you also keep paths and the tree in the Tree component, and sync the two using event listeners. We can get rid of so much code! ⚔️ 

So here's my biggest ask for this review: Can you transform the Tree service from a class with event listeners into a functional service? I'm positive it's possible. I see one big function: `createTreeFromPaths(paths, selectedPath)` (or from `fixtures` directly, which ever is easier – it's not even clear yet to my why we need the intermediary `paths` but it probably makes sense to keep the tree service cleaner). This one returns a tree but you never really have to add it to state, as it can always be derived deterministically. Then you have `selectedPath`, which also doesn't need to be added to state, because you can derive it from `urlParams.component` and `urlParams.fixture`. So whenever a node is clicked, the event bubbles up to FixtureLoader, which updates urlParams from selected path, and then that will trickled down (Flux style), effectively rendering a different `selectedPath` and `tree` respectively.  What do you think, doable?

Thank you @amrdraz!

skidding:
Closing this in favor of #448 

Thanks @amrdraz!

Issue #440, opened 2017-09-09
Redesign filesystem API


skidding:
### What's up?

Now that we [made the proxy API great again](https://github.com/react-cosmos/react-cosmos/pull/422), it's time to target our next victim. 😈

The README starts with this proposition:
> Cosmos **scans your project for components** and enables you to...

Indeed Cosmos scans your project for components, but the API is terrible.  Namely, `componentPaths`, `fixturePaths`, `getComponentName` are rigid and confusing. These options work if A. your project structure is perfectly organized or B. you know how Cosmos works inside out. Not a #PainlessOnboarding!

### Mkay, tell me more...

I've worked with several large codebases lately and I've reached the conclusion that the filesystem API should primarily focus on giving users **full control** over which components to add to Cosmos where to import them from. This is not a matter of improving documentation, but making the API obvious and flexible in the first place.

> Note: My plan is to support the existing options indefinitely and add complementary ones on top. This only affects an isolated part of the project ([react-cosmos-voyager](https://github.com/react-cosmos/react-cosmos/tree/master/packages/react-cosmos-voyager)) so it won't add much maintenance cost.

The new API is meant to accommodate any level of control, from manually targeting and naming fixtures to letting Cosmos _guess_ where your components are.

## Level 0: *Control freak*

Manually target component and fixture files. Supports any file structure and allows manually naming components and fixtures.

```js
// cosmos.config.js
{
  components: {
    Foo: './src/components/Foo',
    Bar: './src/components/Foo'
  },
  fixtures: {
    Foo: {
      one: './src/components/Foo/__fixtures__/one',
      two: './src/components/Foo/__fixtures__/two'
    },
    Bar: {
      one: './src/components/Bar/__fixtures__/one',
    }
  }
}
```

Users with intricate codebases can automate on top of this by using `glob` and similar tools.

## Level 1: *I did it my way*

Manually target component files, place fixtures by convention. Supports any component file structure and allows manually naming components.

```js
// cosmos.config.js
{
  components: {
    Foo: './src/components/Foo',
    Bar: './src/components/Foo'
  }
}
```

This will work:

```
components/Foo/index.js
components/Foo/__fixtures__/one.js
components/Foo/__fixtures__/two.js
```

This will also work:

```
components/Foo.js
components/__fixtures__/Foo/one.js
components/__fixtures__/Foo/two.js
```

A nice feature of Level 0 and 1 is the ability to **whitelist which components to load**, instead of adding tons of ignore rules.

[...]

PR #448, opened 2017-09-22
Component Playground Tree View Proof of Concept [WIP]

maxsalven:
As a preface, I appreciate there is an existing PR (https://github.com/react-cosmos/react-cosmos/pull/423) for this feature. I don't want to step on anyone's toes, so if having two 'competing' PRs causes stress, I'm happy to delete this.

This is a very early stage PR for a lightweight implementation of a tree view in the component playground.

Animated GIF:
cosmos

Initial state:
screen shot 2017-09-22 at 1 25 58 pm

After collapsing some folders:
screen shot 2017-09-22 at 1 26 27 pm

Searches match directories:
screen shot 2017-09-22 at 1 27 33 pm

Searches match fixtures:
screen shot 2017-09-22 at 1 27 28 pm

I stuck to the original styling as much as possible, as it already looks great.

The code basically works as follows:

1. When `FixtureList` mounts, we parse the `fixtures` prop into a tree structure, and store two copies of it in state (`originalData` and `filteredData`). There are two copies so that if we change the search filter, we can run our pruning against `originalData`, so we don't need to re-parse the `fixtures` again.

The tree needs to be in state so we can maintain the state of whether a folder is expanded or not.

2. A separate, stateless functional component `TreeRenderer` accepts this as a prop, along with a few others (`selected`, `onSelect`, `searchText`). 

That's basically it. The `selected` prop is just taken from the `urlParams`, nothing complicated. Selecting a node just calls `onUrlChange`, just like in the non-tree implementation.

The shape of the tree structure can be seen here, it's very simple and we use a little bit of redundancy to keep the rest of the code really clean:
https://github.com/maxsalven/react-cosmos/commit/b32f1afd505fdb4d9aa377627f339fa71075b464#diff-d328d2dbd3c9dc9f69490af7715b18d1R41

Search filtering is done within the `FixtureList` component (we could make the `TreeRenderer` handle this based on the `searchText` prop, but I figured this way is a little cleaner). If you modify the filter, we prune the original tree and update `state.filteredData`.

The only thing controversial is that we *mutate state* to toggle whether a node is expanded or not: 
https://github.com/maxsalven/react-cosmos/commit/b32f1afd505fdb4d9aa377627f339fa71075b464#diff-0d068eb382c8ee1bd6dde837a3c66a10R87
If you're allergic to such a 'hack', then it's not a big deal to write a pure method to deal with this, but I think this is quite an elegant solution (borrowed from `react-treebeard`).

### Notes

I didn't really spend much time looking around the `cosmos` codebase, and just hacked this together quickly. It's not a production ready PR, there are no tests (or fixtures!) etc. I just wrote this for fun, and it turns out it's a much lighter alternative to the other PR. There are probably things I didn't understand or overlooked etc. I just wanted to present it as an option, and if there's desire to get something similar to this merged, then I can help bring it up to production quality.

### Questions / Issues / Todo

- I haven't used the fuzzy matcher here (although it'll be very easy to do so), as I really like the search highlighting feature, and I couldn't see how to extract from `fuzz-aldrin` which characters to highlight.
- I had some strange bug with setting the `ref` on the `component` div, probably because I don't actually have a proper development environment set up. So for now, the 'scroll to selected Component on page load' isn't working. **Update:** I think this is because I'm using a stateless functional component, should work if I refactor it to a `React.Component`.
- We could add animation to the expand/collapse
- The down/right arrows should be extracted to `SvgIcon`, they're currently just some poor CSS.
- We currently only build the tree upon mounting. I imagine the `fixtures` prop can change over time however, so we should probably look to have a `componentWillReceiveProps` hook and recalculate the tree state, taking care to maintain the expanded/collapsed states for any components that persist across this prop change.
- I think I saw somewhere that we want to be able to click on a component itself (soon), and not just a fixture, so I implemented this (but it's a one line change to disable). There'd need to be some UX/UI discussion around this, as currently it toggles the folder as well as remove the `&fixture` part of the query string.
- I don't know about the overall architecture, but it may make more sense to have the tree-shaped fixture data as an actual prop into `FixtureList` rather than deriving it within the `FixtureList`. Unsure here.
- Just realized there's a simple bug where you can't type a space into the search bar, but given the fuzzy matcher is disabled, and I don't think component names or fixture names can have spaces(?), then typing in a space would just mean no matches anyway, so I'll postpone a fix for this until we have a bigger discussion around the matcher.
- Maybe we want to have some sort of 'collapse all' button at the root level?


skidding:
This is great!

> As a preface, I appreciate there is an existing PR (#423) for this feature. I don't want to step on anyone's toes, so if having two 'competing' PRs causes stress, I'm happy to delete this.

@amrdraz I hope this doesn't cause frustration. The end goal is to deliver the best UX with the leanest code, so even though I have a lot of suggestions for improvement I like the simplicity of this PR and think it has a good change of getting merged. That said, I've learned a lot from #423 and I already used some insights from it to review this PR. I would love it we could all collaborate on this to create the best result.

> I stuck to the original styling as much as possible, as it already looks great.

Now that both components and fixtures are clickable it feels a bit weird for components to take up more space and have a bigger font than fixtures. I tried setting components to 14px font and 30px height and find it more natural (besides saving space):

image

> When FixtureList mounts, we parse the fixtures prop into a tree structure, and store two copies of it in state (originalData and filteredData). There are two copies so that if we change the search filter, we can run our pruning against originalData, so we don't need to re-parse the fixtures again.

Generating `originalData` and storing it in the state makes sense, tho I would probably call it something like `fixtureTree`. But I don't think we need `filteredData` stored as it's easily derived from originalData. Did you keep it to be able to mutate it on expand/collapse? I'm sure there are elegant ways to functionally update a single tree node. I think it's called "tree walk".

> The shape of the tree structure can be seen here, it's very simple and we use a little bit of redundancy to keep the rest of the code really clean:

Since we don't mind redundant data I think we would benefit more from adding explicit `urlParams: { component[, fixture] }` to the tree structure. This way we don't have to do any inference like knowing whether a node is a component or fixture based on it having children or not. We just use this property to set the URL and we construct it when building the tree structure. E.g.

```js
{
  name: 'dirA',
  expanded: true,
  children: [
    {
      name: 'Component1',
      // We won't add this for now tho since we don't have a component page yet
      urlParams: { component: 'dirA/Component1' },
      expanded: true,
      children: [
        {
          name: 'fixtureA',
          urlParams: { component: 'dirA/Component1', fixture: 'fixtureA' }
        },
        {
          name: 'fixtureB',
          urlParams: { component: 'dirA/Component1', fixture: 'fixtureB' }
        }
      ]
    }
  ]
};
```

> A separate, stateless functional component TreeRenderer accepts this as a prop, along with a few others (selected, onSelect, searchText).

I'm a bit confused by the onSelect handler. It seems to do two things: toggle tree node visibility and select a component (future) or fixture. These are two different actions and only some nodes have one while others have the other. One thing I dislike is that we send an `expanded` value for tree leaves, which don't have children. How about having two different actions?

- `onToggle`: Triggered only on nodes with children (aka folders). Bubbles up to tree root where tree nodes are recalculated and trickled back down.
- `onSelect`: Handled inside `TreeItem` by just merging `node.urlParams` (see my data proposal above) over the current url params. This means TreeItems receive `onUrlChange` via props and don't need to bubble up the event since they can handle the action themselves.

This model is future proof. Consider this scenario that I have in my git stash somewhere: Tree nodes have an additional `filePath` attribute. TreeItem gets a small *external link* icon button on the right side which fires a request to the node server with that file path, which opens up that file in your OS editor. Feels great when you play with it. Will adapt & publish that code after this refactoring is completed.

> I haven't used the fuzzy matcher here (although it'll be very easy to do so), as I really like the search highlighting feature, and I couldn't see how to extract from fuzz-aldrin which characters to highlight.

Didn't take a close look at the API yet, but it would be nice to preserve at least a part of the fuzzy matcher. Namely, I like that you can now type: "COMPONENT_NAMEFIXTURE_NAME", since fixtures are often named the same across components and you can't target them otherwise.

Nonetheless, the new highlighting is badass. 😎

> I had some strange bug with setting the ref on the component div, probably because I don't actually have a proper development environment set up. So for now, the 'scroll to selected Component on page load' isn't working. Update: I think this is because I'm using a stateless functional component, should work if I refactor it to a React.Component.

Well at the moment it's trying to use refs in FixtureList, which don't exist anymore as the elements have been moved in TreeRenderer/TreeFolder/TreeItem. The scrollIntoView logic should probably be placed in the component that renders the item we want to scroll to.

PS. Can we rename `TreeRenderer` to `Tree`? I feel like all components are "renderers" 😬

> We could add animation to the expand/collapse
The down/right arrows should be extracted to SvgIcon, they're currently just some poor CSS.

I find line-based arrows like Atom uses more pleasant, it'd be cool we could use something similar

image

What I did until now is find the material icons I like (https://material.io/icons/#ic_keyboard_arrow_right) and then copy the [SVG path](https://github.com/callemall/material-ui/blob/353f64d896eca71399c1e97a5c02d3ed5d78050c/packages/material-ui-icons/src/KeyboardArrowRight.js) into a SvgIcon.

> We currently only build the tree upon mounting. I imagine the fixtures prop can change over time however, so we should probably look to have a componentWillReceiveProps hook and recalculate the tree state, taking care to maintain the expanded/collapsed states for any components that persist across this prop change.

Indeed. New fixtures show up via webpack hot module reloading.

> I think I saw somewhere that we want to be able to click on a component itself (soon), and not just a fixture, so I implemented this (but it's a one line change to disable). There'd need to be some UX/UI discussion around this, as currently it toggles the folder as well as remove the &fixture part of the query string.

See my comments above. We must support this but not enable it for now, since we don't have anything to show on component pages atm.

> I don't know about the overall architecture, but it may make more sense to have the tree-shaped fixture data as an actual prop into FixtureList rather than deriving it within the FixtureList. Unsure here.

I wouldn't worry about this too much for now. Eventually we will merge this vision with the one from #440, since it's useful to circulate as few data structures as possible throughout the codebase. So once #440 is done we might even work with a nested structure (ie. `folder.folder.component` instead of `folder/folder/component`) from the get go, starting with user config. But I think it makes more sense to tackle this in #440 and carry on here as long as the new data structure makes sense for our rendering needs (including future ones).

> Just realized there's a simple bug where you can't type a space into the search bar, but given the fuzzy matcher is disabled, and I don't think component names or fixture names can have spaces(?), then typing in a space would just mean no matches anyway, so I'll postpone a fix for this until we have a bigger discussion around the matcher.

It's because we do `const searchText = e.target.value.trim();` before updating the state. As a UX rule of thumb, we shouldn't ever alter user input but derive stuff from it before sending it to wherever we use it (in this case the tree filtering algorithm). It feels crippling to a user when they type something in an input and only part if it registers. Solution is to save searchText in state unaltered and then trim the search text and derive the filtered data on the fly (already suggested this above).

> Maybe we want to have some sort of 'collapse all' button at the root level?

Definitely, but in a subsequent PR to not make this one bigger.

--

Other:
- I found a bug: Type something and then press ESC, the input is cleared but search results are kept (I think this automatically gets fixed if we move to deriving filtered tree on the fly instead of keeping it in state)
- Added some inline comments but then changed my strategy to review high level stuff first (mostly data structures and data flow). I might point to some other, lower impact stuff in a 2nd review, but I think this comment covers the essentials.

amrdraz:
@skidding  I'm always happy to learn, I was having trouble (when I had time) to think of a leaner way of reducing the code to handle the open state so I'm excited to see/learn from this PR

This already looks far more promising thank you @maxsalven for your proposal

packages/react-cosmos-component-playground/package.json
```diff
@@ -2,11 +2,10 @@
   "name": "react-cosmos-component-playground",
   "version": "2.0.0-rc.1",
   "description": "Dev tool for creating reusable React components",
-  "repository": "https://github.com/react-cosmos/react-cosmos/tree/master/packages/react-cosmos-component-playground",
+  "repository":
+    "https://github.com/react-cosmos/react-cosmos/tree/master/packages/react-cosmos-component-playground",
```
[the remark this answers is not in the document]
maxsalven: Not sure how to fix this on my end. Can this be specified via a .prettierrc file (which you should probably have anyway?)

packages/react-cosmos-component-playground/src/components/TreeRenderer/index.js
```diff
@@ -0,0 +1,129 @@
...
+        searchWords={[searchText]}
+        textToHighlight={node.name}
+        highlightClassName={styles.searchHighlight}
+      />
+    </a>
+  );
+};
+
+const TreeRenderer = ({
+  nodeArray,
```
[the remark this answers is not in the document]
maxsalven: Looking at other implementations of tree renderers, I saw that some actually expect an object as their root level input `{ name: "Root Directory", children: [..] }` so I wanted to be explicit that we don't have a single parent root directory, we have an array of them. Happy to change to `nodes` if you prefer not to be so explicit however.

maxsalven:
I've pushed updates that deal with everything you raised, albeit a couple items could use a bit of improvement (the last two in this list). I think the code is clearer now, thanks for the great suggestions!

- [X] Renamed TreeRenderer to Tree
- [X] Better Icons
- [X] Adjust heights
- [X] Added fixtures
- [X] Make filteredFixtureTree derived data rather than state
- [X] Include urlParams in tree shape and don't include them for non-fixtures for now
- [X] Split out `onToggle` and `onSelect` methods
- [X] Highlighting/search improvements, deals with fuzzy search, spaces and fixes escape bug [WIP]
- [X] Basic support for receiving updated Fixtures props [WIP]

> Generating originalData and storing it in the state makes sense, tho I would probably call it something like fixtureTree. But I don't think we need filteredData stored as it's easily derived from originalData. Did you keep it to be able to mutate it on expand/collapse? I'm sure there are elegant ways to functionally update a single tree node. I think it's called "tree walk".

I've updated it to make the filteredFixtureTree as derived data. I originally kept the filteredFixtureTree in state because I was concerned mainly about:

A) Re-calculating the filtered tree on every render. I assumed at the time that if you edit a component props in the Editor, that the filtered tree would recalculate per keystroke and may cause lag with large trees. However it doesn't seem to be the case that it re-renders per keystroke, so this is no longer a concern.

B) Issues with expanding/collapsed folder state. Now that the filteredTree is derived, you can't actually toggle the folders when a filter is applied, they are all fixed in the open position. If this is an issue, I can clean this up, but it'll mean a more complicated `onToggle` method (we can't just mutate the node anymore, we have to find the matching node in the fixtureTree state and update that).

> Since we don't mind redundant data I think we would benefit more from adding explicit urlParams: { component[, fixture] } to the tree structure. This way we don't have to do any inference like knowing whether a node is a component or fixture based on it having children or not. We just use this property to set the URL and we construct it when building the tree structure. E.g.

Updated. I feel it makes Tree a little bit less independent (urlParams is a bit specific to `react-cosmos` rather than a generic concern of tree rendering) but it's not a big deal. If you wanted to be really clean about it, you'd probably want to pass in a `isNodeSelected` function into the `Tree` component, but I don't think it's worth the extra complexity.

>Didn't take a close look at the API yet, but it would be nice to preserve at least a part of the fuzzy matcher. Namely, I like that you can now type: "COMPONENT_NAMEFIXTURE_NAME", since fixtures are often named the same across components and you can't target them otherwise.

I managed to get the fuzzy matcher to work, along with highlighting, kind of. 

You can search with spaces and fuzzy matches now, and we highlight correctly:
screen shot 2017-09-26 at 11 05 09 am

However, if the search spans two different things, then we lose highlighting at the moment:
screen shot 2017-09-26 at 11 06 58 am

> We currently only build the tree upon mounting. I imagine the fixtures prop can change over time however, so we should probably look to have a componentWillReceiveProps hook and recalculate the tree state, taking care to maintain the expanded/collapsed states for any components that persist across this prop change.
> Indeed. New fixtures show up via webpack hot module reloading.

I added very basic support, which seems to work fine:
```
  componentWillReceiveProps(nextProps) {
    if (
      JSON.stringify(nextProps.fixtures) !== JSON.stringify(this.props.fixtures)
    ) {
      this.setState({ fixtureTree: fixturesToTreeData(nextProps.fixtures) });
    }
  }
```

The caveat is that you currently lose the expanded/collapsed state if you modify a fixture. We eventually want to walk the new tree, checking at every node if that node exists in the original tree, and if so, using the `expanded` value from the original tree. Don't think it's critical, but a TODO. We might want to consider this feature alongside a broader "save expanded/collapsed state to local storage".

maxsalven:
I pushed a proof of concept for storing the expansion state locally:
https://github.com/maxsalven/react-cosmos/tree/playground-tree-view-local-storage

I was thinking that it's a bit pointless having collapsible folders if every time I load the window, it'll all be super long and expanded again. So may as well build in this feature.

It's pretty naive, we just store an object in local storage that looks like:
```
{
  dirB/Component2: false,
  dirB/subdirA/Component4: true
}
```

We also have a corresponding `localStorageKey` in the tree structure which we use to look up into the local storage object.

Very simple, works well. Issues:
- If you are working on two different projects, and have `dirA` in both of them, then collapsing in one `cosmos` will mean it's collapsed when you switch to the other `cosmos`. I'm not sure how we can namespace a project. Probably not a big concern.
- Because we don't have namespaced projects, we don't know when a fixture has been deleted, so we can't delete the appropriate items from the local storage object. Not a huge deal, just means the object might get a little 'big' but the total amount of `cosmos` UI folders a given user will click on in their lifetime is small enough to never be a real performance or storage space concern. We could make the object above thinner by only storing 'false' states, at the cost of some small amount of clarity.

packages/react-cosmos-component-playground/package.json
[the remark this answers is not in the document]
skidding: TIL `prettierrc` exists :)

packages/react-cosmos-component-playground/src/components/Tree/index.js
```diff
@@ -0,0 +1,207 @@
...
+      onClick={e => {
+        e.preventDefault();
+        e.stopPropagation();
+        onToggle(node, !node.expanded);
+      }}
+    >
+      <div
+        className={styles.componentName}
+        style={{
+          paddingLeft: CONTAINER_LEFT_PADDING + nestingLevel * INDENT_PADDING
```
skidding: Looks like we need to add `"no-mixed-operators": 0,` to XO rules in package.json, because it's conflicting with Prettier which doesn't like superfluous brackets.

packages/react-cosmos-component-playground/src/components/Tree/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+const nodeShape = shape({
+  name: string.isRequired,
+  expanded: bool,
+  urlParams: shape({
+    component: string.isRequired,
+    fixture: string.isRequired
+  })
+});
+nodeShape.children = arrayOf(nodeShape);
```
skidding: Took me a few seconds to realize why you put this here 😅

packages/react-cosmos-component-playground/src/components/FixtureList/filter.js
```diff
@@ -0,0 +1,38 @@
...
+export const findNode = (node, filter, matcher) => {
+  return (
+    matcher(filter, node) || // i match
+    (node.children && // or i have decendents and one of them match
+      node.children.length &&
+      Boolean(node.children.find(child => findNode(child, filter, matcher))))
+  );
+};
+
+export const filterTree = (node, filter, matcher = defaultMatcher) => {
```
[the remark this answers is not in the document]
maxsalven: Done

packages/react-cosmos-component-playground/src/components/FixtureList/filter.js
```diff
@@ -0,0 +1,38 @@
...
+  // The root object can be an array
+  if (node.constructor === Array) {
```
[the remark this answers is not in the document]
maxsalven: Done, except we now have an XO issue due to circular function calls

packages/react-cosmos-component-playground/src/components/FixtureList/index.js
```diff
@@ -83,23 +52,30 @@ export default class FixtureList extends Component {
...
-      searchText: e.target.value
-    });
+  onSearchChange = e => {
+    this.setState({ searchText: e.target.value });
   };
 
-  onFixtureClick = e => {
-    e.preventDefault();
+  onToggle = (node, expanded) => {
+    // Mutates state. The world won't explode, just be aware. Hugely simplifies things.
```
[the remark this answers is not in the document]
maxsalven: Done

packages/react-cosmos-component-playground/src/components/FixtureList/filter.js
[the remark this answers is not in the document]
maxsalven: I guess you don't want a global 'no-use-before-define' rule override, so I added `// eslint-disable-next-line no-use-before-define`. Is there an XO specific syntax I should use instead?

packages/react-cosmos-component-playground/src/components/Tree/index.js
maxsalven: Done. Unfortunately there's another Prettier formatting issue. Not sure how you want to deal wit h this (now and going forward).

packages/react-cosmos-component-playground/src/components/Tree/index.js
maxsalven: 👍

maxsalven:
I added a new fixture under Tree with an active search, and added the styling improvements. Also took care of cleaning up the filter related code, and fixed a bug with the text highlighter (if you searched for 'a', it would render e.g. 'A' instead of 'DirA').

>Keep everything folder rendered and hide it via CSS, this will only trigger componentDidMount once. This might also enable nice CSS transitions if we just change the height of a container div from 0 to full and vice versa.

Have implemented this. The scrollIntoView bug is now fixed. However, I couldn't get the CSS transition to work. I did a bit of google, which suggested using max-height, but it doesn't work with `auto`. So then we'd have to dynamically traverse the subtree at each folder to find out the height of the children (a naive implementation would have to check all sub-branches and count the number of visible items). Is there a much cleaner CSS only solution you think we can use for this, rather than having to do dynamic height calcs?

packages/react-cosmos-component-playground/src/components/FixtureList/filter.js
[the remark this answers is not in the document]
skidding: > Done, except we now have an XO issue due to circular function calls

This is because `const foo =` is a var declaration, and the right way to do it is to declare `function foo` for root level functions in a module. I've not be consistent with this myself as it's nicer to use arrow and to always write functions in a single way, but I'm planning to use `function` declarations for module exports from now on. So make them function declarations if it's not a big deal, otherwise I can live with since I'm sure there are plenty other similar cases in the codebase.

> I guess you don't want a global 'no-use-before-define' rule override, so I added // eslint-disable-next-line no-use-before-define. Is there an XO specific syntax I should use instead?

Nope, this is fine.

packages/react-cosmos-component-playground/src/components/Tree/index.js
skidding: > Unfortunately there's another Prettier formatting issue.

Which one?

> Not sure how you want to deal with this (now and going forward).

Wouldn't want to bother you and this PR with it, but the long term plan is to cherry pick a few ESLint rules remove XO. Too many rules makes for a bad DX and Prettier already takes care of most of the linter's previous responsibilities.

skidding:
> Is there a much cleaner CSS only solution you think we can use for this, rather than having to do dynamic height calcs?

Honestly, no. I feel a bit guilty for suggesting it because I've done this many times in the past and every time I went through the same process as you and ended up using JS calc 😅. I was hoping maybe it became easier to do this nowadays (and maybe it has and neither of us sees it), but for now I say we keep it simple.

---

Looking thru the rest of the changes now.

packages/react-cosmos-component-playground/src/components/Tree/index.js
skidding: OH, I think you meant this commit 78fbe50c0a75d99150bfd536fafad664650606ed

Well, I think adding a file `.prettierignore` with the following fixes it once and for all...

```
package.json
packages/*/package.json
examples/*/package.json
```

packages/react-cosmos-component-playground/src/components/Tree/index.less
```diff
@@ -38,6 +38,15 @@
   }
 }
 
+.componentChildren {
+  overflow: hidden;
+  transition: max-height 1s ease;
```
skidding: This doesn't have any effect for the moment, right? Unless you want to give animations another try I say we simplify and rely on good ol `display: none/block`

skidding:
Finished review. @maxsalven this looks almost ready!

I left a few small comments (hope you can find them, Github PRs can be confusing), but I think the main part is to fix the FixtureList and any other failing tests. Otherwise this looks functional and pretty. Is there anything else on your mind?

packages/react-cosmos-component-playground/src/components/Tree/index.js
maxsalven: Did you want this to go in this PR, or will you create a separate one for it?

packages/react-cosmos-component-playground/src/components/Tree/index.less
maxsalven: Done

maxsalven:
I realized we had a bug in that the `fullScreen` and `editor` params weren't persisted when clicking on fixtures. 

I went through a few different stages fixing this, and I'm not sure what solution you like most.

https://github.com/react-cosmos/react-cosmos/pull/448/commits/0981099165701fcbc22a36368d28606a54fb12d2

Here, we ignore these global params in the actual fixture `href` attributes, so the status bar will show a URL without these (even though clicking on the link will correctly update the URL). Very simple code, but slightly wrong hrefs.

https://github.com/react-cosmos/react-cosmos/pull/448/commits/09364d67658599477d062e1e625955b2a7d72acc

Here we have the correct href attributes on the fixtures. Means we have to mess around a fair bit with the `Tree` code.

https://github.com/react-cosmos/react-cosmos/pull/448/commits/2052ee12c0eb63aa73ed88caaa529cd0315b0ffd

Here we make a slight change so that we pass the entire URL (including domain) to the `onUrlChange` function. This makes the last 2 failing tests pass, but perhaps is a bit too much leaning towards coding specifically to make tests pass, rather than 'the best code'.

Let me know which of these three 'levels' you like most, and then I can adapt the tests to the desired level.

Note these commits only make the existing tests green, there aren't any tests for the actual tree related code yet.

packages/react-cosmos-component-playground/src/components/Tree/index.js
skidding: No need. I was already working on something else and put it on that branch.

skidding:
I was good with the 2nd option but it seems like the 3rd is even better, so I'd say keep that especially if it's the latest. One small note, but maybe `baseUrlParams` should be called `currentUrlParams` (or just `urlParams` if you prefer brevity), since if I understand correctly they represent the current url params (including a different component/fixture pair that might be selected), as opposed to a subset of url params to extend.

maxsalven:
> One small note, but maybe baseUrlParams should be called currentUrlParams (or just urlParams if you prefer brevity), since if I understand correctly they represent the current url params (including a different component/fixture pair that might be selected), as opposed to a subset of url params to extend.

Updated to `currentUrlParams`.

maxsalven:
I added some tests to the `Tree` component:

[-] Renders component and fixture names
[-] Indents things appropriately
[-] Renders links and incorporates currentUrlParams
[-] Allows you to select on a fixture
[-] Allow you to toggle a component
[-] Highlights search text

Please let me know if there's anything else you'd like to see or if there's anything you'd change.

packages/react-cosmos-component-playground/src/components/FixtureList/index.js
```diff
@@ -90,23 +58,29 @@ export default class FixtureList extends Component {
     }
   };
 
-  onChange = e => {
-    this.setState({
-      searchText: e.target.value
-    });
+  onSearchChange = e => {
```
skidding: You have to update line 28 to `window.__changePlaygroundSearch = this.onSearchChange;` for the Cypress smoke tests to pass

maxsalven: Done

skidding:
Hello Tree View! 🌴🌳🎉

Thanks @maxsalven!

skidding:
Regarding [tree view persistence](https://github.com/react-cosmos/react-cosmos/pull/448#issuecomment-332090580). Feel free to start a new thread for this.

> I was thinking that it's a bit pointless having collapsible folders if every time I load the window, it'll all be super long and expanded again. So may as well build in this feature.

👍

> It's pretty naive, we just store an object in local storage that looks like:

Sounds good.

> If you are working on two different projects, and have dirA in both of them, then collapsing in one cosmos will mean it's collapsed when you switch to the other cosmos. I'm not sure how we can namespace a project. Probably not a big concern.

We have a `rootPath` on the server side (derived from cosmos.config or process.cwd when config will be optional). We can surface that to client to namespace each state per project.

> Because we don't have namespaced projects, we don't know when a fixture has been deleted, so we can't delete the appropriate items from the local storage object. Not a huge deal, just means the object might get a little 'big' but the total amount of cosmos UI folders a given user will click on in their lifetime is small enough to never be a real performance or storage space concern. We could make the object above thinner by only storing 'false' states, at the cost of some small amount of clarity.

I think my previous point partly responds to this. So I'd say we should do a cleanup call whenever we receive the component/fixture tree.

> I guess that's because localForage is using IndexedDB? Time for a new proxy? :)

That would be best. Hopefully IndexDB can be overridden globally. This might help: https://github.com/dumbmatter/fakeIndexedDB

PR #467, opened 2017-10-05
feat: React 16 compatibility

skidding:
While React 16 itself has an almost seamless upgrade path, making Cosmos compatible with React 16 required considerable effort due to how we use React and the other related tools (mainly Enzyme). 

> Disclaimer: This is an overview of the changes required to make this happen. Other projects might meet different challenges when upgrading, so make sure to carefully read each relevant change log beforehand.

#### Order of mount/unmount

> When replacing \<A /\> with \<B /\>, B.componentWillMount now always happens before A.componentWillUnmount. Previously, A.componentWillUnmount could fire first in some cases.

Under normal circumstances this shouldn't matter, but FetchProxy and XhrProxy are global. By global I mean there can only be one in page at the same time. These proxy mock a global dependency when created and unmock it on unmount. Because of the new order of events, changing the mocks would result in A.mock => B.mock => A.unmock. We fixed this by only unmocking mocks set by current instance.

#### Refs are no longer accepted on stateless components https://github.com/facebook/react/issues/10831

They weren't supported since 0.14 I think, but React never complained when they were set. This was convenient because the Loader treated all components the same. We fixed this by checking the type of component and only adding refs to React.createClass and ES Class. I don't think this is mentioned in any change log, though.

#### *requestAnimationFrame* shim required in Jest

```js
// jest.setup.js
global.requestAnimationFrame = cb => setTimeout(cb, 0);
```

#### New adapter required

Enzyme v3 comes with separate adapters we need to manually config. Reasonable. This is the final Jest config. Notice how the `requestAnimationFrame` shim is required before importing the React 16 adapter.

```js
// jest.setup.js
import { configure } from 'enzyme';

global.requestAnimationFrame = cb => setTimeout(cb, 0);

// We need to import this after requestAnimationFrame shim is added
const Adapter = require('enzyme-adapter-react-16');

configure({ adapter: new Adapter() });
```

#### Wrappers no longer auto update https://github.com/airbnb/enzyme/issues/1175

This is a big one and required changing a lot of tests. There's a bunch of people complaining and requesting the functionality back. I'm fairly confident the updated Cosmos tests are solid and in line with Enzyme v3, but one person makes a good point about this change:

> Yes, I can go and fix the failing tests by adding an explicit update() call. **But what about my existing tests that pass with v3. When I refactor components, existing tests may pass by coincidence, since I'm missing an update() call that I may now potentially need.** I don't want to go through 1000's of tests and check they still make sense.

#### *wrapper.node* is no longer available

> Attempted to access ReactWrapper::node, which was previously a private property on
            Enzyme ReactWrapper instances, but is no longer and should not be relied upon.
            Consider using the getElement() method instead.

`getElement()` might be useful in some cases, but it's not equivalent. `.node` returned the actual DOM node, whereas `getElement()` returns another wrapper and it doesn't seem to be documented at the moment.

#### Cypress events no longer fire React events https://github.com/cypress-io/cypress/issues/647

Until this is solved, I put a wild workaround in place to keep the smoke tests running https://github.com/react-cosmos/react-cosmos/commit/0331891ae4875088d6dc80f991ff3d07fb625942.

@ovidiubute @flaviusone @tkjone @xavcz you might find this interesting.


skidding:
In case it's not clear: This impacts the Cosmos repo. Cosmos will continue to be compatible with React 14 & 15.

skidding:
Done! `2.1.0` is React 16 compatible.

NiGhTTraX:
This looks scary to me. Don't you think the tests are too coupled to React/Enzyme details? I certainly can't compare the complexity of Cosmos to the complexity of my project, but upgrading React for me was a [breeze](https://github.com/NiGhTTraX/react-test-buffet/pull/158/files). I'm curious to hear your thoughts on whether the advantages brought by Enzyme and Cypress are worth the tight coupling observed in this PR.

skidding:
> I'm curious to hear your thoughts on whether the advantages brought by Enzyme and Cypress are worth the tight coupling observed in this PR.

Cypress: Yeah, sure. It's a disposable piece. I only added one test file on purpose and it offers enough value as smoke tests: If somehow a bug slips through the cracks of unit tests, at least we know that the UI is standing are core actions work. It took me some time to debug the bug I mentioned, but the cost of maintenance is low since as I said it's just one test file. I was considering adding one end to end test for each example, which would make an awesome CI, but I don't plan to invest more in Cypress until it stabilizes.

Enzyme: This is a tricky one. Tbh I asked myself a couple of times if I could do without. I don't even use `shallow` anymore*, so I guess I might as well render normally at this point. The selectors are the remaining valuable part in Enzyme. It's very easy to find children, filter through them, get props, etc. Are there any other libs for doing this? @NiGhTTraX what do you use, bare ReactTestUtils?

\* I gave up on `shallow` because it doesn't call lifecycle events which I often want to test organically. So instead of thinking whether to use `shallow` or whether to use `mount`, I just use `mount` all the time. It's fast enough and Cosmos fixture mock everything for children to work.

NiGhTTraX:
> I don't plan to invest more in Cypress until it stabilizes.

Have you considered another tool, or does the roadmap for Cypress look orange? :tangerine: 

In the brief time I used Enzyme I too didn't use `shallow` because it didn't exercise the entire component e.g. lifecycle methods and state and refs at one point (is this still the case?). For selectors I wrap the components in jQuery and [use its API](https://github.com/NiGhTTraX/react-test-buffet/blob/6562f9a3624ccaf8983de90f3ffcdde09e89dc24/tests/unit/components/todo-filter.spec.js#L24). It's probably the only place where I still enjoy using jQuery :smile: 

When it comes to querying children in tests I took a different approach.

At first it seems nice that Enzyme lets you easily check the props sent to a child component, going so far as to not render them to make the tests atomic. However, once I made it clear in tests that I'm interested in the data passing between a parent component and its children I also wanted to invert that dependency and make it more explicit.

Therefore, I write components that accept other components through either `children` or props. Then I use [mock components](https://github.com/NiGhTTraX/react-test-buffet/blob/6562f9a3624ccaf8983de90f3ffcdde09e89dc24/tests/unit/helpers/chai-react.js#L67-L131) in tests and set [expectations](https://github.com/NiGhTTraX/react-test-buffet/blob/6562f9a3624ccaf8983de90f3ffcdde09e89dc24/tests/unit/components/list.spec.js#L27-L31) on them.

I guess the pressure this puts on good design for components is about the same as when checking the props with Enzyme, although with my approach there is more emphasis on [events coming from the children](https://github.com/NiGhTTraX/react-test-buffet/blob/6562f9a3624ccaf8983de90f3ffcdde09e89dc24/tests/unit/components/list.spec.js#L33-L37). The main advantage for me is that if I can easily pass a mock in a test then I can just as easily replace implementations in [production code](https://github.com/NiGhTTraX/react-test-buffet/blob/6562f9a3624ccaf8983de90f3ffcdde09e89dc24/src/app-factory.js#L21). My components become generic and composable and I can do more with less of them.

It seems verbose at first, but in the end I enjoyed being able to compose components in endless ways e.g. `<Modal><Dialog><Survey Toggleable={Checkbox} /></Dialog></Modal>` and having unit tests for each of them (and one screenshot for the cluster) instead of writing tests for a `<ModalDialogSurveyWithCheckboxes />` component.

P.S.: A/B tests become a delight - you just change 1 line of code.

ovidiubute:
FWIW you can still run lifecycle methods in Enzyme shallow rendering. Pass
{ lifecycleExperimental: true } to shallow as 2nd argument. This will be
default behavior in a future Enzyme release but we've been using this with
success for some time already.

PR #469, opened 2017-10-12
New fixture format and file discovery

skidding:
From: #440


skidding:
I don't expect anyone to review the 9000 LOC diff, but I did review it entirely myself and I wanted to point out any change I found interesting (besides the obvious API change discussed in length at #440 😀):

- Enabled async/await and already used it in the newer modules.
- Added first **Flow types**! Added some notes on this in the Monorepo section from the Contributing Guide.
- Started enforcing a case convention for file names, found under "Keep coding style consistent" in the Contributing Guide.
- Renamed react-cosmos-component-playground to react-cosmos-playground for brevity.
- Removed the `component` prop from Loader and all proxies, since component is now found under `fixture.component`.
- Refactored Cypress tests to make each test independent from the others
- **Made `cosmos.config.js` optional** (since `componentPaths` is no longer a hard requirement).
- Updated all examples to new fixture format and removed cosmos.config from all, except for a new example meant exactly for old style fixtures.
- Moved from Lerna bootstraping to **Yarn Workspaces!** The Yarn team is still working on fixing a couple of bugs related to Workspaces, but this cleans up the repo a lot: A single `yarn.lock` root file. There are many other advantages, one of them being `yarn upgrade-interactive` working across packages.
- Added `component` to list of "special props" in normalize-props-proxy cc @flaviusone
- Removed all global Jest mocks from `packages/__mocks__` to make packages less coupled. Mocking per package, better yet per test file, is preferred.
- A bunch of other bug fixes, package upgrades and refactors not worth going into detail

I'm merging this branch because reviewing additional commits is impossible and it will keep creating conflicts with other PRs. Here are the next steps:

- One or two extra PRs with small changes
- Moving from v3 alpha to v3 beta
- I'll prepare a README update to reflect the new API, but will refrain from merging until v3 RC goes out, to avoid confusing new users.
- **Please continue to test the `next` tag and come up with feedback and suggestions.**

This was massive effort but I couldn't have done it without you @amrdraz @jlc467 @maxsalven @flaviusone @tkjone!

PR #470, opened 2017-10-17
Tree View enhancements

skidding:
Just tiny things I had in my stash that weren't worth the back and forth. Biggest thing is I made the Tree styles look good by itself so we no longer need the `background: black` hack. @maxsalven please review. I think every review is self explanatory but let me know if you see any issue.

image

### DIFF
--- a/packages/react-cosmos-component-playground/src/components/FixtureList/dataMapper.js
+++ b/packages/react-cosmos-component-playground/src/components/FixtureList/dataMapper.js
@@ -10,6 +10,8 @@ const dataObjectToNestedArray = (base, path = '') => {
         name: key,
         expanded: true,
         children
+        // TODO: Enable this when we'll have component pages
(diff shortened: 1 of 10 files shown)


skidding:
Merging this so I can update #469. Feel free to comment later.

packages/react-cosmos-component-playground/src/components/FixtureList/dataMapper.js
```diff
@@ -10,6 +10,8 @@ const dataObjectToNestedArray = (base, path = '') => {
         name: key,
         expanded: true,
         children
+        // TODO: Enable this when we'll have component pages
+        // https://github.com/react-cosmos/react-cosmos/issues/314
```
maxsalven: For this feature, we'll also want to make sure we don't accidentally copy over fixture from currentUrlParams when generating hrefs and handling clicks in the Tree component.

maxsalven:
Looks good, thanks.

Issue #476, opened 2017-10-25
React Router proxy querystring issue


maxsalven:
When clicking on a `Link` in a component that uses react-router, if the link has a query string, then the component will render appropriately but the `fixture.url` value will not include the query string in the editor.

As a basic fix, we should render the url correctly in the editor. I haven't looked into it, but this may be as simple as changing:
https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-router-proxy/src/LocationInterceptor.js#L5
`const getPathFromProps = props => props.location.pathname;`
to
``const getPathFromProps = ({location: { pathname, search, hash }}) => `${pathname}${search}${hash}`;``

As a more advanced fix, we should support the `location` object from react-router, which includes a state field that is not exposed by the url but can impact how your component renders:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md

I'm not sure what the best way to proceed here is.

One option is to allow specifying only either `fixture.url` or `fixture.location`. This is probably the simplest solution, but isn't that nice.

I'd prefer to see both when I use the react-router proxy and for the two to stay in sync when I navigate the site. But when I set up my fixture, I'd prefer to just specify something like:

```
{
  url: "/my/url?foo=true#somehash",
  location: {
    state: {
      from: "dashboard"
    }
  }
}
```

and then when I open the fixture in cosmos, I get a fully populated location:
```
{
  url: "/my/url?foo=true#somehash",
  location: {
    pathname: "/my/url",
    search: "?foo=true",
    hash: "#somehash",
    state: {
      from: "dashboard"
    }
  }
}
```

It's a little bit of magic, but I don't think it will confuse anyone.

The main question here is what if you edit the fixture in the react-cosmos fixture editor, and you put in contradictory information (i.e. location.pathname doesn't match what's in location.url). Ideally the change you make would get reflected in the appropriate other field.

I'll try to find time to play around with the router and see what works.

skidding:
Thanks for opening this. My first thought is that synchronizing `fixture.url` and `fixture.location` is a world of pain. The only clean option for keeping `pathname`, `search` and `hash` together while preserving the location state is something like this:

```js
{
  url: "/my/url?foo=true#somehash",
  locationState: {
    from: "dashboard"
  }
}
```

The router proxy can map `url` and `locationState` deterministically to the Router's `location` structure and vice versa.

skidding:
Closed by #477

PR #477, opened 2017-10-25
React router proxy improvements

maxsalven:
Addresses https://github.com/react-cosmos/react-cosmos/issues/476

reactroutercosmos

Fixes `fixture.url` not reflecting navigated to query strings and hashes.

Also reflects the [react-router location object](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md) in the editor. This lets you edit the `location.state` field which is not exposed in the URL as well, and you can much more clearly see which props your component is receiving from the router.

If you provide both a url and a location fixture which don't match (i.e. you might specify url="/a" but location.pathname = "/b"), it will use the location fixture.

The only thing I wasn't able to achieve is keeping the fixture "in sync" when editing it (navigating in the component works fine, I think). By this I mean if you change the fixture.url, then location.pathname does not update, and vice versa. The component also only updates when you edit location.pathname.

@skidding any ideas on how to fix this?


maxsalven:
An alternative is to only ever show location.state, and only do so if it has a value.

This has the benefit of not having two competing 'sources of truth', but the downside of not being able to nicely see `location.{pathname,search,hash}`.

skidding:
> @skidding any ideas on how to fix this?

How about [this](https://github.com/react-cosmos/react-cosmos/issues/476#issuecomment-339264764)?

> An alternative is to only ever show url and location.state, and only do so if it has a value.

This has the benefit of not having two competing 'sources of truth', but the downside of not being able to nicely see location.{pathname,search,hash}.

Ideally I think we get the sync problem fixed instead. Or alternatively, we have 'read only' parts of the fixture (since location.pathname etc isn't really a prop, it's a function of the url).

Actually it seems like you reached the same conclusion before me. 👍

maxsalven:
I've updated this to just use the `url` and (optional) `locationState` fixtures. Have added some basic tests as well.

Note that we rely on the 'url' npm package. However if I add this as a dependency to https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-router-proxy/package.json

then I get an error when running `yarn install` in the root react-cosmos directory:

```
Build failed { Error: spawn webpack ENOENT
    at _errnoException (util.js:1021:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:192:19)
    at onErrorNT (internal/child_process.js:374:16)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  code: 'ENOENT',
  errno: 'ENOENT',
  syscall: 'spawn webpack',
  path: 'webpack',
  spawnargs: 
   [ '--config',
     'packages/react-cosmos-playground/webpack.config.js' ] }
```

What should I do?

skidding:
> then I get an error when running yarn install in the root react-cosmos directory:

This is due to a [Yarn bug](https://github.com/yarnpkg/yarn/issues/4706) that was fixed but a new version hasn't been published yet. Check out [this comment](https://github.com/yarnpkg/yarn/issues/4706#issuecomment-336567094) for a temporary solution. I recommend the 2nd option because removing `node_modules` and re-installing every dep takes a lot of time.

packages/react-cosmos-router-proxy/src/index.js
```diff
@@ -1,23 +1,51 @@
 import React from 'react';
-import { string } from 'prop-types';
+import { string, any } from 'prop-types';
 import { MemoryRouter, Route } from 'react-router';
 import { proxyPropTypes } from 'react-cosmos-shared/lib/react';
 import LocationInterceptor from './LocationInterceptor';
+import urlParser from 'url';
+
+function buildLocation(url, locationState) {
+  const parsedUrl = urlParser.parse(url);
```
skidding: Nitpicking: Destructuring `const { pathname, search, hash } = urlParser.parse(url)` would make for clean shorthand keys in the returned object.

packages/react-cosmos-router-proxy/src/index.js
```diff
@@ -31,7 +59,8 @@ export default () => {
   RouterProxy.propTypes = {
     ...proxyPropTypes,
     route: string,
-    url: string
+    url: string,
+    locationState: any
```
skidding: Shouldn't this be an object type?

packages/react-cosmos-router-proxy/src/index.js
maxsalven: Good idea, thanks.

packages/react-cosmos-router-proxy/src/index.js
maxsalven: Whilst `location.state` would typically be an object, I think it can actually be anything. There's nothing in the official docs, but it works perfectly fine with e.g. a string. The flow-typed libdefs also indicate any:
https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-router_v4.x.x/flow_v0.53.x-/react-router_v4.x.x.js#L9

skidding:
Excellent! 👌 

Found a Playground bug while debugging this: #493

PR #478, opened 2017-10-26
Pin Prettier

skidding:


skidding:
@maxsalven does this look right? Thanks for pointing it out.

maxsalven:
Looks good! :+1:

PR #481, opened 2017-10-28
Error catch proxy

skidding:
TODO:
- [x] Use Flow types
- [x] Style error message


skidding:
**Playground no longer needs a hard reload after unhandled error in Loader!**

cc @maxsalven 

error-catch-proxy

I also added some extra Flow typing. This is the first proxy with Flow-typed props instead of React propTypes. I'm pretty excited about the typing of the Proxy linked list.

First there's the generic linked list type:

```js
export type LinkedItem<Item> = {
  value: Item,
  next: () => LinkedItem<Item>
};
```

Then the proxy props which use a `LinkedItem` type with a `ComponentType` argument, which also expects the same proxy props recursively...

```js
export type ProxyProps = {
  nextProxy: LinkedItem<ComponentType<ProxyProps>>,
  fixture: Object,
  onComponentRef: Function,
  onFixtureUpdate: Function
};
```

This structure statically ensures that `props.nextProxy.value` is a Component (with the same props) and `props.nextProxy.next` is a function which returns the same structure recursively. @xavcz you might like this 😎

skidding:
Almost forgot, @tkjone I think you'll also find this interesting. First `componentDidCatch` implementation and Flow stuff.

xavcz:
Dude!! I'm STOKED, that is awesome developer experience!! 🔥

Can't wait to try it 🎉

skidding:
Published under `react-cosmos-webpack@3.0.0-beta.3`

PR #482, opened 2017-10-29
[WIP] Treeview Persist Rebased

maxsalven:
There were so many changes since the previous PR (https://github.com/react-cosmos/react-cosmos/pull/471) that it was easier for me to just port everything over manually to a fresh branch rather than try to merge stuff in. Hope that's ok! Unfortunately that means I'll need to port the discussion over to here:

https://github.com/react-cosmos/react-cosmos/pull/471#discussion_r146375229
> If we return || {} here we can always expect an Object and never have to do savedExpansionState || {} when consuming the expansion state.

The problem is that `localForage.getItem` always returns a promise, so `|| {}` will never get reached. It doesn't look like localForage supports fallback/default values either. Is there anyway to wrap the localForage.getItem method so that their promise returns `{}` instead of null?

https://github.com/react-cosmos/react-cosmos/pull/471#discussion_r146376796
> Wouldn't it make sense to prune here as well? Fixtures can change via HMR. In which case it might make sense to bundle this logic and hide the entire expansion state business under a function like this:

The only thing pruning does is clear out the expansion state for directory paths that *no longer exist*, i.e. the user has moved the component or deleted the fixture. I think both these actions are fairly rare. Pruning doesn't have any impact/benefit on when a user adds a new fixture via HMR (it'll essentially be a noop in this scenario). The only benefit of pruning is clearing out a few bytes of storage in the user's indexedDB cache, and a microscopic speedup in our code that checks if a folder is expanded or not (as there will be less items to look through). Given this, I would say we don't even **need** to prune at all, and fairly infrequently would be fine. So I think pruning only on mount is plenty often enough, and there's no benefit to pruning every single time we load the fixture tree. Thoughts?

> Does a prop with default value need to be marked as required? 

I view these as two separate things. If we ever remove the default prop, we still want to know that this is a required prop. So personally I think it's better coding hygiene to leave the `required` in there.

TODO:
- [x] Existing Tests: I wasn't able to get your suggestion regarding waiting for the localForage.getItem promise to return to work (https://github.com/react-cosmos/react-cosmos/compare/treeview-persist-2?expand=1#diff-14e33cea23c7b5a14667d77e126875d9R20). Please let me know if you can take a look, thanks.
- [x] New Tests: To test the `persistence.js` file, do you think I should just mock out `getSavedExpansionState` and `setSavedExpansionState`? How would I test the actual `FixtureList` component, given the persistence relies on localForage/indexedDB?
- [x] `projectKey`. If we don't have a projectKey working before this gets merged in, then we should probably disable the pruning code for now, as otherwise everytime we switch projects, we'll lose all our saved folder expansion state. As someone who has multiple react-cosmos projects, this will be frustrating.


skidding:
> The problem is that localForage.getItem always returns a promise, so || {} will never get reached. It doesn't look like localForage supports fallback/default values either. Is there anyway to wrap the localForage.getItem method so that their promise returns {} instead of null?

My bad for not paying attention. This should do it:

```js
export async function getSavedExpansionState(projectKey) {
  const value = await localForage.getItem(`${TREE_EXPANSION_STATE}-${projectKey}`);
  return value !== null ? value : {};
}
```

FYI, async/await is now enabled so you can use it anywhere you want to make code more readable, including component lifecycle methods.

> The only thing pruning does is clear out the expansion state for directory paths that no longer exist, i.e. the user has moved the component or deleted the fixture. I think both these actions are fairly rare. Pruning doesn't have any impact/benefit on when a user adds a new fixture via HMR (it'll essentially be a noop in this scenario). The only benefit of pruning is clearing out a few bytes of storage in the user's indexedDB cache, and a microscopic speedup in our code that checks if a folder is expanded or not (as there will be less items to look through). Given this, I would say we don't even need to prune at all, and fairly infrequently would be fine. So I think pruning only on mount is plenty often enough, and there's no benefit to pruning every single time we load the fixture tree. Thoughts?

Webpack sometimes fails on this, but HMR can be triggered by removing files as well, not just adding/changing. So in theory the fixture list can reduce while browsing the playground (updated via componentWillReceiveProps). That said, I get your point that pruning is the least of our concerns and could be just as well done once a month. I suggested we do it in both places because I thought it makes the code more modular with no extra cost. So my suggestion is to do whatever makes more sense from a code readability/maintainability standpoint.

> I view these as two separate things. If we ever remove the default prop, we still want to know that this is a required prop. So personally I think it's better coding hygiene to leave the required in there.

Good point. I'll do the same in the future.

> Existing Tests: I wasn't able to get your suggestion regarding waiting for the localForage.getItem promise to return to work (react-cosmos/react-cosmos/compare/treeview-persist-2?expand=1#diff-14e33cea23c7b5a14667d77e126875d9R20). Please let me know if you can take a look, thanks.

I'll check out the branch locally and report back what I find.

> New Tests: To test the persistence.js file, do you think I should just mock out getSavedExpansionState and setSavedExpansionState? How would I test the actual FixtureList component, given the persistence relies on localForage/indexedDB?

By mocking `localForage`. 
- The mock itself [already exists](https://github.com/react-cosmos/react-cosmos/blob/8948dbc18bcad1c526056f35251d5c49c0458c6e/packages/react-cosmos-playground/src/__mocks__/localforage.js)
- Here's an example of [activating the mock](https://github.com/react-cosmos/react-cosmos/blob/8948dbc18bcad1c526056f35251d5c49c0458c6e/packages/react-cosmos-playground/src/components/ComponentPlayground/__tests__/fixture-editor/pane-portrait.js#L8-L10)
- Here's an example of [setting the mock value](https://github.com/react-cosmos/react-cosmos/blob/8948dbc18bcad1c526056f35251d5c49c0458c6e/packages/react-cosmos-playground/src/components/ComponentPlayground/__tests__/fixture-editor/pane-portrait.js#L108-L110)

Does this answer your question?

> projectKey. If we don't have a projectKey working before this gets merged in, then we should probably disable the pruning code for now, as otherwise everytime we switch projects, we'll lose all our saved folder expansion state. As someone who has multiple react-cosmos projects, this will be frustrating.

Assume projectKey exists. I will add it quickly after you finish this and only publish next version after both are merged.

skidding:
Found the issue with the test. It didn't have anything to do with async promises, but with the new Enzyme 3 API. We need to call `wrapper.update` after each action which changes the render tree (in this case the action is the setState call from componentDidMount which occurs while we wait "afterOngoingPromises"). This is how the passing beforeEach block looks like:

```js
beforeEach(async () => {
  wrapper = mount(<Loader fixture={populatedFixture} />);
  await afterOngoingPromises();
  wrapper.update();
});
```

This can be done without async/await...

```js
beforeEach(() => {
  wrapper = mount(<Loader fixture={populatedFixture} />);
  return afterOngoingPromises().then(() => {
    wrapper.update();
  });
});
```

... but who would want to go back to nesting callbacks? Ugh. 😄

maxsalven:
Thanks for the feedback, have pushed some updates. We're getting close!

I updated the localForage mock to return null instead of undefined if a key doesn't exist; this matches the library behaviour.
https://github.com/react-cosmos/react-cosmos/pull/482/commits/79b348b350cc976078f36c22d02cd18b68d45676#diff-2b32aef9f8b1bf9ef8191509a3bcccadR7

I got the existing tests to work using the method you suggested. However it's not something I'm experienced in, so please take a close look to make sure you are happy:
https://github.com/react-cosmos/react-cosmos/pull/482/commits/5011f189b8976248c571f3f3475c988e9ad07111

I have a failing test because we don't currently mock `localForage.setItem`. Is there a reason why you didn't mock it (and instead went for an internal method `__setItemMocks`? (I don't want to waste time trying to write a mock if you already spent some time and realized it wouldn't work for some reason).

In the interim, we can't test the `updateLocalToggleState` method.
https://github.com/react-cosmos/react-cosmos/pull/482/commits/e73c1a62583c84043584efbda19589d2658f3a24#diff-98dd3fe452d5da6f7ac8f0f22e79d300R15

skidding:
Tests are looking great!

> I updated the localForage mock to return null instead of undefined if a key doesn't exist; this matches the library behaviour.

Good idea!

> Is there a reason why you didn't mock it (and instead went for an internal method __setItemMocks? (I don't want to waste time trying to write a mock if you already spent some time and realized it wouldn't work for some reason).

I don't think I had any good reason besides blindly applying the `__setMocks` pattern I've been using for mocking other modules which don't have a setter like localForage does. So unless you encountered any issue with it I say implementing the mock's `setItem` method to update `itemMocks` is a good idea. That said, I would keep `__setItemMocks` for two reasons:
- Unlike `setItem`, it resets any existing mock data
- Not having to update existing tests

Is this it or did I miss any other outstanding issue?

maxsalven:
I updated the localForage mock and upgraded the persistence tests.

I also merged master and implemented persistence for the fixture folders.

I think we're good to go here(?)!

skidding:
Another great contribution from @maxsalven! 🤘

skidding:
PS. I predict we'll want to default to nothing expanded as the default state.

maxsalven:
> Another great contribution from @maxsalven! 🤘

:) Thanks for merging!

> PS. I predict we'll want to default to nothing expanded as the default state.

Yeah it's a tricky one. I think for first time users, it's nice to see your first ever fixture without having to expand any folders etc. For advanced users, you probably don't want 20 open folders. But now that we persist things, it's not a huge deal for an advanced user to close some folders as a one-off task. 

Either way, a very simple change to make if needed!

PR #483, opened 2017-10-30
Allow namespacing of fixtures in tree view

maxsalven:
If users have the `namespace` attribute on a fixture set, then instead of rendering a fixture name as "my namespace/fixture name", we render fixture directories instead.

Only allows a single level of nesting.

I added a different fixture directory icon, didn't really know which one to pick. This can probably be improved, perhaps we want something with some angled brackets instead of a star.

screen shot 2017-10-30 at 11 52 58 am

### DIFF
--- a/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/nested-data.js
+++ b/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/nested-data.js
@@ -7,7 +7,12 @@ export default {
     fixtures: {
       'dirA/Component1': ['fixtureA', 'fixtureB'],
       'dirB/Component2': ['fixtureA', 'fixtureB'],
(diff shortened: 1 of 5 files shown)


maxsalven:
Perhaps we want to make the fixture directory color the same as the fixture color (i.e. grayish)?

Mockup:
screen shot 2017-10-30 at 12 00 07 pm

skidding:
Cool!

Didn't check the code yet, but regarding design: The starred folder is interesting, but makes me wonder if it's some sort of bookmark. I like the lighter/more transparent version because the folder matches the text color, which makes perfect sense. So I'd would go with the 2nd version without the star.

skidding:
No other comments aside from the icon design. Code looks great!

maxsalven:
Updated the styling and added some tests.

skidding:
Works & looks great! 👌

PR #484, opened 2017-10-30
Add `react-error-overlay`

Timer:
I'm going to bed, but this should get you started.

If you'd like, you can tell me what adjustments to make and I'll see what I can do. 😄

I'm not familiar with `react-cosmos`, so I added build and runtime error reporting to where I felt it made most sense.

http://g.recordit.co/LGb1qX9yJw.gif

Enables all `react-error-overlay` features: runtime errors, click-to-open source, and build time errors.


skidding:
@Timer what a wonderful surprise! 😍

I have a few questions:
- The runtime error overlay doesn't auto close after a HMR update that fixes the problem. Is this the expected behavior? It probably is because of the X close button and it makes sense for runtime errors to stick because they might have corrupted something global, but I wanted to make sure this is how it works in CRA as well. Y/n?
- CRA's runtime handler seems to render (pun intended) the root `componentDidCatch` handler [I just merged](https://github.com/react-cosmos/react-cosmos/pull/481) useless. I don't mind of course because this is next level stuff, but I'm curious what's React's vision in general for this. In this case it seems `componentDidCatch` overlaps 100% CRA's runtime handler. Is `componentDidCatch` meant more for graceful error handling in production and CRA will always overlap custom error states in development? cc @gaearon 
- It seems like `react-error-overlay` publishes ES modules, which is at the moment incompatible with Cosmos' packaging (`react-dev-utils` seems all CJS which is great). `loader-entry` is included in the user bundle and because Cosmos is meant to work with a wide range of setups we always publish CJS modules and make no assumptions about the user's loaders. Not sure how to handle this... 🤔 It's the only real obstacle in merging this. Ideas welcome!

Timer:
> The runtime error overlay doesn't auto close after a HMR update that fixes the problem. Is this the expected behavior? It probably is because of the X close button and it makes sense for runtime errors to stick because they might have corrupted something global, but I wanted to make sure this is how it works in CRA as well.

It is the expected but not the best behavior.
In CRA we force a page reload on next save after a runtime error because of global corrupted state, so we should make that the default behavior here (doesn't do that currently).

> ... `componentDidCatch` ...

I'm not on the React team so take what I say with a grain of salt.

IMO, `componentDidCatch` is meant for a production use so that your users aren't left with a broken application.
In development, using `react-error-overlay` is going to give a much better development experience -- you could always close the overlay and look at your `componentDidCatch` page for testing purposes too.

> It seems like `react-error-overlay` publishes ES modules, which is at the moment ... It's the only real obstacle in merging this. Ideas welcome!

We can dual publish a CJS and ESM version (based on `package.json` toggle). Can you open an issue on CRA for this please?

gaearon:
>We can dual publish a CJS and ESM version (based on package.json toggle). Can you open an issue on CRA for this please?

We should.

skidding:
> It is the expected but not the best behavior.
In CRA we force a page reload on next save after a runtime error because of global corrupted state, so we should make that the default behavior here (doesn't do that currently).

Actually I think it's best not to hard reload in this case. The nice thing about Cosmos is that it keeps state outside of the user's render tree, so regardless of what crashes in the user code we can re-render from scratch after a HMR change and still preserve previous state.

Even if it's Redux state, we create the store from scratch and flush all component instances on every render. It's a different take on hot reloading...

runtime-recover

> We can dual publish a CJS and ESM version (based on package.json toggle). Can you open an issue on CRA for this please?

Done: https://github.com/facebookincubator/create-react-app/issues/3354

Timer:
> Actually I think it's best not to hard reload in this case. 

Totally your call. 😄 

If you'd like, you can open another issue for support for something like dismiss-runtime-error that way it can be automatically closed.

skidding:
> If you'd like, you can open another issue for support for something like dismiss-runtime-error that way it can be automatically closed.

Sounds good, but we can definitely merge without that. 

If you're interested in finishing the remaining bits, there are [four outstanding ESLint errors](https://travis-ci.org/react-cosmos/react-cosmos/jobs/294704243#L785-L789) and 3-4 falling Jest tests—partly due to react-error-overlay ES syntax that Jest doesn't transform and thus understand, so best to tackle tests after facebookincubator/create-react-app#3354 is resolved.

Thanks again for your interest in this!

Timer:
I've rebased this so we don't fall behind -- did you want to tackle https://github.com/facebookincubator/create-react-app/issues/3354 or do you want me to take care of it? 😄 No rush.

skidding:
> I've rebased this so we don't fall behind -- did you want to tackle facebookincubator/create-react-app#3354 or do you want me to take care of it? 😄 No rush.

I'll take a stab at it 🔪

Timer:
Whoops, sorry! We just released `react-error-overlay@3` yesterday which changed all of this. 😄

I'll update this PR tonight and see where it puts us. Sorry for the run-around.

skidding:
I was eager to make the change but this is even better! 

> I'll update this PR tonight and see where it puts us. Sorry for the run-around.

Excellent. Let me know if you run into any trouble. Besides the 3 falling tests I have one minor ask: Can you put the CRO logic inside a separate file to keep loader-entry clean? [Like `./react-devtools-hook`](https://github.com/Timer/react-cosmos/blob/0f648fce8f95db28b81dbd531187671ee289883a/packages/react-cosmos-webpack/src/client/loader-entry.js#L1)

Timer:
:shipit:

skidding:
Thank you @Timer, this is a great improvement in Cosmos' DX!

Quick question: Is there any difference/advantage in using `import * as ErrorOverlay from` instead of just `import ErrorOverlay from`?

Timer:
@skidding in `react-error-overlay@2` doing `import ErrorOverlay from` was actually different than `import * as ErrorOverlay from` because we were working with real ES modules and not compiled ones.

I'm not sure what happens with webpack's UMD build, but it's more future proof this way (we may ship two versions instead of UMD).

So doing `import ErrorOverlay` resulted in strictly the default, when we really wanted the named exports (which requires `import * as ErrorOverlay`).

packages/react-cosmos-webpack/src/server/server.js
```diff
@@ -3,9 +3,11 @@ import { silent as silentImport } from 'import-from';
 import express from 'express';
 import webpackDevMiddleware from 'webpack-dev-middleware';
 import webpackHotMiddleware from 'webpack-hot-middleware';
+import launchEditor from 'react-dev-utils/launchEditor';
 import getCosmosConfig from 'react-cosmos-config';
 import extendWebpackConfig from './extend-webpack-config';
 import getUserWebpackConfig from './user-webpack-config';
+
```
Timer: Sorry about this new line, I didn't notice it.

skidding:
@Timer Got it. Thanks again!

Timer:
My pleasure! Let me know if you ever need help upgrading in the future.

skidding:
@Timer thinking of programmatically closing runtime error overlay in some cases. I noticed there's a `dismissRuntimeErrors` function but [it isn't exported](https://github.com/facebookincubator/create-react-app/blob/9ce144ed90c37f116b8457821d36dba2ceb0490a/packages/react-error-overlay/src/index.js#L90). Do you see any reasons why it shouldn't be called? Otherwise I'd like to submit a PR for exporting it.

Timer:
Mainly because in our use case it doesn't make sense to dismiss. Happy to accept a pull request exporting it.

PR #487, opened 2017-10-30
Find fixture files with .ts and .tsx extensions by default

mbohgard:
Less config for TypeScript users!


skidding:
Thank you @mbohgard! 

Published under `3.0.0-beta.4`. Please try it out and see if it works without having to define `fileMatch` in your config.

mbohgard:
Yep, works great :)

PR #498, opened 2017-11-05
Handle missing index.html gracefully

skidding:
No index.html with default webpack config
image

No index.html with default custom config
image

Bonus: Loading screen
image


skidding:
TODO:
- [ ] Add "Loading..." screen while bundling

jlc467:
Hi @skidding,

In the first screenshot, is `html-webpack-config` the intended package name?

Also, what is the advantage of making the user install it separately when using default webpack config, versus installing it with `react-cosmos-webpack`?

skidding:
Thanks for pointing it out, that was a typo! 😊

PR #500, opened 2017-11-07
HTTP Proxy API requests

maxsalven:
Create-react-app provides a `proxy` option in `package.json` (not to be confused with react-cosmos proxies):
https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#proxying-api-requests-in-development

React-cosmos does not respect this setting, so API requests will fail in this scenario.

This PR shows a working proof of concept for proxying API requests in react-cosmos.

@skidding currently I've got it hard-coded to a specific host, just as a proof of concept.

Would you want a new `httpProxy` option to go in to the cosmos config, for generic use by anyone? 

Would you want us to set the httpProxy option automatically for create-react-app users? What's the appropriate way to do this in react-cosmos?

Any idea on how to write a test for this?


maxsalven:
Also note if we want to support the full range of proxy options that create-react-app supports, then I think we may need to match the config done in webpack-dev-server? Not sure if that's correct or not.

https://github.com/webpack/webpack-dev-server/blob/d2fef5803a771a7de7e0f8f218d9c232b7754c05/lib/Server.js#L166

If this is the case, I'd suggest we support a basic "/api" -> "http://somwhere:1234/api" using a config option instead until someone requests something more complicated.

skidding:
This is tricky. Your solution is fair, but it looks like using webpack-dev-server is the best long term strategy. The problem is there is at least one large webpack 1 codebase depending on Cosmos and WDS depends on webpack 2.2+. While it's inevitable to move on, I would preserve this compatibility for Cosmos 3.0. Does CRA support a subset of WDS proxy options via package.json? Maybe we can mirror that for now.

maxsalven:
It looks like CRA has two options from the end-user perspective:

1. Specify a proxy string (`"proxy": "http://localhost:4000",`), and *all* unmatched requests will get sent to the proxy, with a single caveat: "The development server will only attempt to send requests without `text/html` in its `Accept` header to the proxy." I don't know what impact having all unmatched requests be proxied will have on `react-cosmos` (do you currently use a catch-all route in express for anything?)

2. You can specify a full-blown proxy config object: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#configuring-the-proxy-manually

The code to parse the proxy in create-react-app is complex:
https://github.com/facebookincubator/create-react-app/blob/2e82ebb3371731a5c4e346f310848ddb23fd0976/packages/react-dev-utils/WebpackDevServerUtils.js#L266

Personally I don't think it's worth spending time on flexible proxy setup options now given a future switch to webpack-dev-server. 

Would you be ok with moving forward with this PR and use a config setting for the really simple "common use case" whilst we await the move to webpack-dev-server? I.e. we only allow users to proxy from a given path to a given url. Given it's an explicit opt-in feature in CRA, I don't think it's a big deal to not autodetect it and instead ask users to opt-in here as well. I imagine anyone using a proxy is not a beginner to development and won't have major issues to figure out they need to tweak something in cosmos.

Just to re-iterate my proposal, one extra optional setting in the config:

Option 1. `httpProxy: "http://localhost:1234/api"`
Option 2. `httpProxy: { context: "/api", target "http://localhost:1234" }`

and a few extra lines in the source code.

Very happy to look into (auto-)matching CRA config once we're on webpack-dev-server. Thanks!

skidding:
> Would you be ok with moving forward with this PR and use a config setting for the really simple "common use case" whilst we await the move to webpack-dev-server?

Yes. Agreed! 👍

We'll also add an excerpt of this discussion in the option's docs for full transparency.

maxsalven:
Have updated the PR. Any thoughts on the best way to test this?

[Note: there are test failures but they're failing on `master` as well]

packages/react-cosmos/src/server/server.js
```diff
@@ -42,6 +43,11 @@ export default function startServer() {
   const loaderCompiler = webpack(loaderWebpackConfig);
   const app = express();
 
+  if (httpProxy) {
+    const { context, target } = httpProxy;
```
skidding: How about an informative log like we do for `publicPath` at line 65?

maxsalven: This is already built in to the middleware library:
```
[Cosmos] No webpack config found, using defaults
[HPM] Proxy created: /api  ->  http://localhost:4001
[Cosmos] See you at http://localhost:8989/
```

maxsalven:
- [x] Added a brief example in the config options section of the docs.
- [x] Added a test, but I'm not confident about this so please check.
- [x] Added flow type in the appropriate place.

skidding:
Flawless!

Issue #430, opened 2017-09-07
Mocking server responses blocks HMR to reload modules


dfedynich:
### What's up?

Mocking server responses allows us to completely isolate components. And that's a great idea. But it blocks HMR to load update modules in the current implementation.

### Mkay, tell me more...

Xhr and fetch proxies use _xhr-mock_ library to mock server responses. This library makes monkey patching of XmlHttpRequest under the hood. This library catches all requests by default and provides mocks only for specified URLs. So, when HRM sends the request for the update module, it will get the response with status '0' and dive in endless checking cycle. To sort this issue out I suggest one of the next solutions:
- Update xhr-mock library to the new version. This library has got some new commits recently, where we can use the factory to mock only specific requests and pass the rest out. However, the current version **2.0.0-preview.5** is rather raw and got some other glitches. So, I don't recommend to update to this version until it becomes robust.
- Fix xhr and fetch proxies to teardown mocks for HMR requests. If you are interested in this fix and choose this way I can create a pull request in the nearest time. Please, let me know if you want me to do that.

skidding:
Good catch! Flagging this as a bug.

A few notes:
- Fetch proxy doesn't use xhr-mock, but [fetch-mock](http://www.wheresrhys.co.uk/fetch-mock/), which doesn't interfere with XHR requests so HMR works just fine with Fetch proxy.
- I agree we should wait until xhr-mock@2.0 becomes stable before upgrading. After that, it would be ideal if we can rely mainly on it to do the job and not have to maintain intricate customizations
- It would be great if you could contribute to fix this! My only ask is to please describe the technical solution before implementing, to make sure we're on the same page.

dfedynich:
Actually, you can add the next check to your xhr-proxy in the constructor

```
if (module.hot) {
  module.hot.status((status) => {
    if (status === 'check') {
      xhrMock.teardown();
    }
  });
}
```

`module.hot.status` is the part of [management API of HMR](https://webpack.github.io/docs/hot-module-replacement.html)

skidding:
Interesting. And then the constructor will run again and the mocks will continue to work?

dfedynich:
Yes, exactly.

skidding:
Interested in making the patch yourself? Contributions are highly appreciated.

dfedynich:
I'd be delighted to make the patch.

skidding:
[...]

PR #503, opened 2017-11-09
Enable HMR in xhr-proxy

maxsalven:
Fixes https://github.com/react-cosmos/react-cosmos/issues/430

Looks like xhr-mock 2 is [still in preview](https://github.com/jameslnewell/xhr-mock/blob/master/CHANGELOG.md), so I implemented the solution proposed in the issue.

Credit and thanks to @dfedynich


skidding:
👍

PR #504, opened 2017-11-09
Capture runtime errors that occur before Loader initializes

skidding:
global-runtime

### DIFF
--- a/packages/react-cosmos-playground/src/components/ComponentPlayground/index.js
+++ b/packages/react-cosmos-playground/src/components/ComponentPlayground/index.js
@@ -1,5 +1,6 @@
+// @flow
+
 import React, { Component } from 'react';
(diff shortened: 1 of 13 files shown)


packages/react-cosmos-playground/src/components/ComponentPlayground/index.js
```diff
@@ -15,18 +16,49 @@ import DragHandle from '../DragHandle';
...
+  fixture?: string,
+  editor?: boolean,
+  fullScreen?: boolean
+};
 
-const fixtureExists = (fixtures, component, fixture) =>
-  fixtures[component] && fixtures[component].indexOf(fixture) !== -1;
+type State = {
+  // There doesn't seem to be a way to use const values as types. Let me know
+  // if you can think of a better way to express this!
```
maxsalven: I think you can do `typeof PENDING | typeof MISSING`

Issue #505, opened 2017-11-09
File paths not resolved for components in .jsx files when importing them in fixtures


iw-dweaver:
### What's up?
In the latest beta (3.0.0-beta.9), when components are imported into fixture files from files with a .jsx extension (or any extension that is not .js for that matter), cosmos is unable to resolve that file path.

### Mkay, tell me more...

The issue is `require.resolve` throws and we end up with null in our component filePaths.

https://github.com/react-cosmos/react-cosmos/blob/385c1218a74fb3350e42fc41698d1c49924656b5/packages/react-cosmos-voyager2/src/server/extract-components-from-fixture-file.js#L103

skidding:
@dan-weaver please try `react-cosmos@3.0.0-beta.10`

dan-weaver:
@skidding thanks works nicely and thanks for the test case!

PR #506, opened 2017-11-09
use resolveUserPath instead of require.resolve for non .js extensions

dan-weaver:
addresses #505 
simply replace require.resolve with the shared util function so it no longer throws when no .js file is found at the path.


dan-weaver:
Oh @skidding i think you had mentioned I needed to bump the version of react-cosmos-shared.. I hadn't done that yet is it still necessary?

skidding:
Thanks!

I will publish a new beta later today.

PR #518, opened 2017-11-15
Expand example react-boilerplate config

TagoreS:
Add the appropriate containerQuerySelector and a globalImports that imports the default global stylesheet to the react-boilerplate example config.


skidding:
Thanks @TagoreS!

PR #523, opened 2017-11-18
Basic extraction of HOCs from component name

maxsalven:
Given:
```
fixtures: {
  ComponentA: ['fixtureA'],
  'withRouter(ComponentA)': ['fixtureA'],
  'withRouter(Connect(ComponentA))': ['fixtureA']
}
```
(Note: I realise it should be `connect` rather than `Connect`, apologies).

Instead of:

screen shot 2017-11-18 at 3 20 54 pm

We show:

screen shot 2017-11-18 at 3 19 29 pm

Please let me know thoughts on styling. I personally prefer to always have the HOC names shown, rather than a tooltip, as they are an integral part of the component I have created a fixture for.

This is just a simple PR, if one wanted to do something intelligent with e.g matching HOC names to proxies (or automatically showing a helpful error message 'we notice your component relies on redux but you don't have the redux proxy installed') then that can come at a later stage.

There's an outstanding issue with search; if you search for "comp with", you'll get no matches as the actual text we are searching is "withRouter(ComponentA)" [the ordering is different]. Not hard to fix but wanted to get approval for the PR before polishing.

Also note if the user has an opening parentheses in their actual component name, then there's not much we can do about that, it'll get parsed.

### DIFF
--- a/packages/react-cosmos-playground/src/utils/common.less
+++ b/packages/react-cosmos-playground/src/utils/common.less
@@ -1,13 +1,18 @@
 @default-spacing: 10px;
-@default-font-family: 'BlinkMacSystemFont', 'Lucida Grande', 'Segoe UI', Ubuntu,
-  Cantarell, sans-serif;
+@default-font-family: 'BlinkMacSystemFont',
+  'Lucida Grande',
+  'Segoe UI',
+  Ubuntu,
+  Cantarell,
+  sans-serif;
 @default-font-size: 14px;
 
 @left-nav-bgcolor: #0c293e;
 @left-nav-bgcolor-darker: rgb(10, 33, 50);
 @left-nav-bgcolor-selected: rgba(255, 255, 255, 0.1);
 @left-nav-color: #a5bed2;
-@left-nav-color-selected: #f8fafc;
+@left-nav-color-selected: rgb(248,250,252);
+@left-nav-color-selected-lighter: rgba(248, 250, 252, 0.6);
 
 @drag-handle-size: 4px;
 @editor-font-size: 14px;
--- a/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
+++ b/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
@@ -8,8 +8,8 @@ export default {
       projectKey: 'test'
     },
     fixtures: {
-      ComponentA: ['fixtureA'],
-      'withRouter(ComponentA)': ['fixtureA'],
+      ThisIsAVeryLongComponentName: ['fixtureA'],
+      'withRouter(ThisIsAVeryLongComponentName)': ['fixtureA'],
       'withRouter(connect(ComponentA))': ['fixtureA']
     },
     urlParams: {},
--- a/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
+++ b/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
@@ -1,0 +1,18 @@
+import FixtureList from '../index';
+
+export default {
+  component: FixtureList,
+
+  props: {
+    options: {
+      projectKey: 'test'
+    },
+    fixtures: {
+      ComponentA: ['fixtureA'],
+      'withRouter(ComponentA)': ['fixtureA'],
+      'withRouter(Connect(ComponentA))': ['fixtureA']
+    },
+    urlParams: {},
+    onUrlChange: () => {}
+  }
+};
--- a/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
+++ b/packages/react-cosmos-playground/src/components/FixtureList/__fixtures__/hoc-names.js
@@ -10,7 +10,7 @@ export default {
     fixtures: {
       ComponentA: ['fixtureA'],
       'withRouter(ComponentA)': ['fixtureA'],
-      'withRouter(Connect(ComponentA))': ['fixtureA']
+      'withRouter(connect(ComponentA))': ['fixtureA']
     },
     urlParams: {},
     onUrlChange: () => {}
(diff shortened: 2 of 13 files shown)


alp82:
Love it, looks great for a useful start.

skidding:
> or automatically showing a helpful error message 'we notice your component relies on redux but you don't have the redux proxy installed'

Definitely a different PR, but this sounds amazing ❤️

> if one wanted to do something intelligent with e.g matching HOC names to proxies

This should be fairly easy to add from the get go, by maintaining a small dictionary. Eg.

```
const knownHocNames = {
  connect: 'Redux',
  withRouter: 'Router',
};
const displayName = knownHocNames[name] || name;
```

Unless it has some downsides I'm not aware of. Like maybe there are other libs that decorate the component with "connect".

---

PR looks great. Comments:
- I think the HoC names would look better with some transparancy rather than being gray (similar to fixture names).
- No sure about this one, but how about reversing the HoCs order? 🤔 The component name is on the left, so it starts from the "inside".
- The ellipsis dots look a bit ugly image
I don't know if this can be changed, but I imagine most components will have "..." in a big app with many HoCs.

maxsalven:
Have pushed some updates.

> I think the HoC names would look better with some transparancy rather than being gray (similar to fixture names).

Done.

> No sure about this one, but how about reversing the HoCs order? 🤔 The component name is on the left, so it starts from the "inside".

Personally I strongly disagree with this. Given the styling, I mentally read it as "The component is Component A, and the HOCs are withRouter, connect". Having the HOCs in reverse would be very confusing to me personally. My brain may be backwards from the majority however!

> I don't know if this can be changed, but I imagine most components will have "..." in a big app with many HoCs.

The closest I could get with simple CSS was to have the ellipsis be styled with the HOC style if a HOC exists. I'm not super familiar with Less syntax, so it's possible that there's a nicer way to write the CSS. It's not quite perfect:

screen shot 2017-11-18 at 9 45 35 pm
screen shot 2017-11-18 at 9 45 45 pm

The first and the third are correct. The second is 'wrong', as we are using the HOC ellipsis style when the ellipsis belongs to the component name. I don't think it looks too bad, and I don't know how to do a better solution without Javascript. Open to PRs ;).

Happy to hear any feedback on the above.

I'll try wrap up the PR this weekend by fixing the search functionality and adding some tests.

packages/react-cosmos-playground/src/utils/common.less
```diff
@@ -1,13 +1,18 @@
...
+  Cantarell,
+  sans-serif;
 @default-font-size: 14px;
 
 @left-nav-bgcolor: #0c293e;
 @left-nav-bgcolor-darker: rgb(10, 33, 50);
 @left-nav-bgcolor-selected: rgba(255, 255, 255, 0.1);
 @left-nav-color: #a5bed2;
-@left-nav-color-selected: #f8fafc;
+@left-nav-color-selected: rgb(248,250,252);
```
skidding: Hmm, Prettier doesn't add spaces after commas?

maxsalven:
My IDE may not be set up for less, and your lint-staged only processes .js files.

________________________________

skidding:
> Personally I strongly disagree with this. Given the styling, I mentally read it as "The component is Component A, and the HOCs are withRouter, connect". Having the HOCs in reverse would be very confusing to me personally. My brain may be backwards from the majority however!

👍

> The closest I could get with simple CSS was to have the ellipsis be styled with the HOC style if a HOC exists.

Nice. I wanted to suggest this but thought it's too crazy for something so small. Can't think of anything simpler, either.

> The first and the third are correct. The second is 'wrong', as we are using the HOC ellipsis style when the ellipsis belongs to the component name.

It's good enough. If we want to rationalize it and make it "a feature not a bug" we can say the light "..." indicates that the component has HoC.

Update: You reached the same conclusion :)

skidding:
> lint-staged only processes .js files

Thanks, will fix.

maxsalven:
I decided to move the extraction of HOC names into the data-mapper layer, so it only runs on mount (or if the fixture props change). Having it dynamically calculated in the Tree component meant it had to recalculate whenever you search or toggle anything, and given the data is static, I think it makes more sense to do it 'once' in the data-mapper.

I wasn't quite sure of a key name, so I used `displayData`. Thoughts?
```
displayData: shape({
    componentName: string.isRequired,
    hocs: arrayOf(string).isRequired,
    search: string.isRequired
  }),
```

I also generated a search string for Components, which is  `` `${componentName} ${hocs.join(', ')}`.trim()``. This way if you search for `component withRouter`, then we will correctly match a component whose actual name is `withRouter(Component)` (which wouldn't normally match).

The matched text **highlighting** isn't ideal, as we are separately rendering the componentName and the hoc names (as they need to be in different spans), and hence the displayed hoc names (`withRouter`) won't match the search string (`component withRouter`) for highlighting purposes, so no highlights show. **The searching still works fine**, just the matches don't always highlight properly. This is similar to existing issues with fixture names not always highlighting when you included part of the component name in your search. I don't think it's a show-stopper for the PR, and it's not something I currently have much time to investigate.

@skidding if you're happy with the code as it is, then I'll put some tests together.

maxsalven:
Realized I never addressed this:

@skidding said:
```
const knownHocNames = {
  connect: 'Redux',
  withRouter: 'Router',
};
const displayName = knownHocNames[name] || name;
```

What I actually was talking about was identifying HOC names in order to automatically enable proxies (which may not even be practical, given most of them need some sort of inputs, e.g. `url`). I think for displaying the HOC names in the UX, we should stick with the actual wrapper names. This is a lot clearer in my opinion.

skidding:
> @skidding if you're happy with the code as it is, then I'll put some tests together.

Yes. Looking good!

maxsalven:
Thanks @skidding. I think we're ready for merging. Test coverage includes checking that the data structure generated by the data mapper correctly deals with HOCs, and that the Tree renders this data correctly.

skidding:
@maxsalven good stuff! I'll make for a neat `3.1` release.

PR #527, opened 2017-11-22
Fix Context Proxy example in README

jozsi:
Hello,

I beleive the context items should be under the `context` key.

P.S.: Great job with v3.0 👍


skidding:
Thanks @jozsi!

PR #531, opened 2017-11-23
Avoid re-starting Cosmos when adding new component / fixtures files

xavcz:
While @skidding was walking me through the Cosmos booting process, we realized that finding the common directory for fixtures and components could be done to avoid re-starting Cosmos when you create a new component or fixtures file.

This PR solves this problem by watching the common directory of fixtures and component:

```
/components/__fixtures__/Foo/blank.js
/components/__fixtures__/Bar/one.js
/components/Foo.js
/components/Bar.js

-> common directory to watch for changes: /components
```

See [screencast](https://d3vv6lp55qjaqc.cloudfront.net/items/1D2m0b2C2a1S3i0O2O30/Screen%20Recording%202017-11-23%20at%2002.59%20PM.gif?X-CloudApp-Visitor-Id=2680119&v=7507541a)

### DIFF
--- a/packages/react-cosmos/src/server/embed-modules-webpack-loader.js
+++ b/packages/react-cosmos/src/server/embed-modules-webpack-loader.js
@@ -32,11 +32,19 @@ module.exports = async function embedModules(source: string) {
   const componentModuleCallls = convertPathsToRequireCalls(
     keys(deprecatedComponentModules).map(c => deprecatedComponentModules[c])
   );
-  const { componentsCommonDir, contextCall } = getContextCall(fixtureFiles);
+
+  // get a flat list of all components paths
+  const paths = fixtureFiles
+    .map(file => file.components.map(component => component.filePath))
+    .reduce((list, current) => [...list, ...current]);
+
+  const componentsCommonDir = commondir(paths);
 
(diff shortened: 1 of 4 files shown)


packages/react-cosmos/src/server/embed-modules-webpack-loader.js
```diff
@@ -32,14 +32,11 @@ module.exports = async function embedModules(source: string) {
   const componentModuleCallls = convertPathsToRequireCalls(
     keys(deprecatedComponentModules).map(c => deprecatedComponentModules[c])
   );
+  const { componentsCommonDir, contextCall } = getContextCall(fixtureFiles);
```
skidding: I hope you don't mind the nitpicking, but can you put `componentsCommonDir = commondir(paths)` here and then just return the `contextCall` from `getContextCall`, as the function name suggests. This way you also just need to send the context path to `getContextCall` instead of all the fixture files.

xavcz: I don't mind at all, that makes sense! Thought about it but didn't do anything. Thanks for the comment!

skidding:
@xavcz thanks!

PR #534, opened 2017-11-25
Test API

skidding:


xavcz:
omg can't wait for it

skidding:
OK, here we go. This is a big one so I'll try to be concise while also not missing something important. A compact part of this will get into the README, so please help me communicate this better if I don't get it right in the first try (most likely).

### Quick recap on the problem this is trying to solve

Fixtures already provide a way to reproduce a component's state thru state injection and mocking via proxies. This is useful in the Playground UI. So far so good. But when writing headless tests for components (eg. via Jest) the hardest part is the setup—setting up the component's state. So why not reuse the fixtures here as well? Great. We've already been doing that by using the `Loader` component inside tests, but this solution came short in many ways. Most notably:
- Proxies had to be imported in each test file
- It didn't play so well with Enzyme, because both React's ref callback (which we sometimes use) and rendering Cosmos fixtures in general is inherently async, so we had to get our hands dirty with Promises regardless of how simple the test was

Besides solving these two problems, writing a dedicated test API opened up the possibility for various other helpers I've found useful to reuse while writing component tests. A large part of these came from my experience consulting for @tkjone's company, where I'm grateful to have been able to test this solution on a large number of use cases). Additionally, all component tests from the Cosmos monorepo have been upgraded to the API I'm about to present, so you can explore the `test-api` branch to see over 100 working examples.

### Almost there... just one more thing

Before we get into meat of it, I also want to mention the underlying refactor behind this branch. I used this opportunity to shift some parts from Cosmos' architecture more towards the long term goal of making Cosmos a general purpose UI dev tool.

Why is this relevant? The new test API means we now have two ways for rendering fixtures, so I wanted to make sure that the Playground and the Test API share as much code as possible. So I created a generic `createContext` abstraction that interfaces with a pluggable `renderer` which currently has two implementations (ReactDOM and Enzyme), but is on track to go beyond React or even the DOM. Long term vision aside, this also makes the new test API easy to use for people who don't use and don't want to use Enzyme.

### Alright, what is this thing?

Let's say for now that we're using Enzyme and Jest, though neither is required. Here's is how a component test looks like:

```js
import createContext from 'react-cosmos-test/enzyme';
import Button from './Button';
import fixture from './fixture';

const { mount, getWrapper } = createContext({ fixture });

beforeEach(mount);

test('renders hello', () => {
  expect(getWrapper().text()).toContain('Hello World');
});
```

Let's see what we have here:
- `createContext` Receives the fixture (and other possible params irrelevant for now) and returns a "context"—a set of functions that enable us to mount the fixture and retrieve information from it later. Calling it has not side effects, so it can be called at any time and in any function scope to suit our needs.
- `mount` Async function that needs to be called for component to render. This is not done implicitly because usually we want to render from scratch before every test. In special cases you can call this once for more tests.
- `getWrapper` Many things to mention here:
  - It prevents us from having to hoist the wrapper var and populate it whenever we call _mount_.
  - It has the extra benefit of automatically updating the wrapper before returning it, solving [an annoying issue with Enzyme v3](https://github.com/airbnb/enzyme/issues/1175).
  - It returns the component wrapper, not the wrapper of the first proxy. Essentially, it does `wrapper.find(fixture.component)`.
  - It accepts an optional selector argument (eg. `getWrapper('.btn')` or `getWrapper(ChildComponent)`)

Let's look at a more complex example. I'll put the fixture inline to get a better picture.

```js
import async from 'async-until';
import createContext from 'react-cosmos-test/enzyme';
import Form from './Form';

const fixture = {
  component: Form,

  fetch: [{
    matcher: '/form-template',
    response: {
      formTitle: 'Dynamic Form'
      // ...
    }
  }]
};

const { mount, getWrapper } = createContext({ fixture });

beforeEach(async () => {
  await mount();

  // We know the component will query /form-template and will
  // render a form with response.formTitle, but due to async
  // methods (like fetch or redux-thunk actions) it might take
  // a couple of event loops for the component to get in the
  // state we're interested in
  await until(() => getWrapper().text().includes('Dynamic Form'))

  // Interested in testing stuff that happens after form is submitted
  getWrapper('form').simulate('submit')
});

test('something happened on submit', () => {
  // ...
});
```

`until` is an additional utility I find useful in this combination, but is not part of the Cosmos test API. Instead of running some assertions after one or more `setImmediate` or `setTimeout(fn, 0)` calls, we give up expectations regarding synchronicity and just focus on the order of events.

Here are a bunch of other _context_ methods:

```js
import createContext from 'react-cosmos-test/enzyme';
import Form from './Form';
import fixture from './fixture';

const { mount, unmount, getWrapper, getRef, get } = createContext({ fixture });

beforeEach(mount);

test('check something on component instance', () => {
  expect(getRef().something).to(/* ... */);
});

test('check something in component state', () => {
  expect(getRef().state.something).to(/* ... */);
  // or
  expect(get('state').something).to(/* ... */);
});

test('something happened in Redux state', () => {
  expect(get('reduxState').something).to(/* ... */);
});

test('something happened in URL location', () => {
  expect(get('url')).to(/* ... */);
});

test('something happens after un-mounting', () => {
  unmount();
  // expect...
});
```

In case it's not obvious, `get` returns any fixture field, (which proxies keep up to date as component changes state).

#### Auto mocking prop callbacks in Jest

When using `react-cosmos-test` inside the Jest env, it automatically decorates prop callbacks. Eg:

```js
import createContext from 'react-cosmos-test/enzyme';
import Button from './Button';

const fixture = {
  component: Button,
  label: 'Click me, yo!',
  onClick: () => console.log('Alright, I was clicked.')
};
const { mount, getWrapper, get } = createContext({ fixture });

beforeEach(mount);

test('renders hello', () => {
  getWrapper('button').simulate('click');

  // props.onClick is automatically wrapped in jest.fn()!
  expect(get('props').onClick).toHaveBeenCalled();
});
```

#### Testing fixtures without Enzyme

Here's an example using `react-test-renderer` instead of Enzyme:

```js
import { create as renderer } from 'react-test-renderer';
import createContext from 'react-cosmos-test/generic';
import fixture from './fixture';

const { mount, getWrapper } = createContext({ renderer, fixture });

beforeEach(mount);

test('renders hello', () => {
  // Careful, this is no longer an Enzyme wrapper, but a react-test-renderer wrapper
  expect(getWrapper().toJSON()).toMatchSnapshot();
});
```

In short, the "generic" `createContext` function accepts a `renderer` option which is a function that, given a React element, has to return a wrapper with an `unmount` method.

### Can I try this?

**Yes you can! Install `react-cosmos-test@next`.**

> You might have to also upgrade existing Cosmos packages to `next` release.

### Feedback, please!

Thank you for taking the time to read this and I appreciate any feedback. Special thanks to @tkjone and @xavcz who already contributed a lot to this.

Other people I believe this is of interest to: @jozsi @maxsalven @dan-weaver @alp82 @flaviusone @NiGhTTraX

alp82:
Looks great. Is there a typo in section "Auto mocking prop callbacks in Jest":
```js
const fixture = {
  component: Button,
  label: 'Click me, yo!',
  onClick: () => console.log('Alright, I was clicked.')
};
```

Shouldn't be that:
```js
const fixture = {
  component: Button,
  props: {
    label: 'Click me, yo!',
    onClick: () => console.log('Alright, I was clicked.')
  }
};
```

alp82:
Some feedback regarding the API:

`createContext(fixture)` is cool. Not sure about the name "context" but got no better alternative right now.

`mount` and `unmount` are obvious and easy to understand. Still one question though: What if i want to mount with slight modifications to any fixture field? For example some change in the state, url or prop. Am I required to have a fixture for each variation or can i use a fixture and manipulate it when mounting it?

` getWrapper()`: is the name maybe a bit misleading? In fact we get the component inside all the HOC wrappers. Isn't `getComponent()` more obvious? Not sure though.

`getRef()` is obvious too, nice

`get()`: i only understood what it does after your explanation "`get` returns any fixture field". I get that short function names are desired, but wouldn't `getField()` or even `getFixtureField()` be more understandable?

skidding:
> Looks great. Is there a typo in section "Auto mocking prop callbacks in Jest"?

You're right. I corrected the typo. Thanks.

> createContext(fixture) is cool. Not sure about the name "context" but got no better alternative right now.

Right. "Context" is one of *those* words, but I feel like in this case is really makes sense — as everyone else in this situation probably 😄. Open to suggestions.

> mount and unmount are obvious and easy to understand. Still one question though: What if i want to mount with slight modifications to any fixture field? For example some change in the state, url or prop. Am I required to have a fixture for each variation or can i use a fixture and manipulate it when mounting it?

You can run `createContext` more than once in the same test file. Eg. once in every describe block. And you can extend the fixture like this:

```js
const { mount /* ... */ } = createContext({
  fixture: {
    ...myFixture,
    url: '/different'
  }
});
```

You can use something like `lodash.merge` for overriding more nested fixture attributes.

> getWrapper(): is the name maybe a bit misleading? In fact we get the component inside all the HOC wrappers. Isn't getComponent() more obvious? Not sure though.

Interesting. Wrapper doesn't refer to component wrapper (eg. an outer \<div\>), but rather to an API wrapper that adds methods on top of the component instance. Eg. `.find`, `.unmount`, etc. It's heavily inspired by how Enzyme works, but since this API is meant to be generic I'm open to other ideas as well.

> get(): i only understood what it does after your explanation "get returns any fixture field". I get that short function names are desired, but wouldn't getField() or even getFixtureField() be more understandable?

I agree, "get" is misleading. The reason why I'm leaning towards something so short is that tests get very repetitive, and after you write "getFixtureField" the 199th time you might wish it was called "get". But maybe we could document the verbose one while also exposing a shorter alias. FYI in previous iterations I used specific getters for specific fixture fields. Eg. `getUrl()`, `getReduxState()`. I like these, but they are coupled with Proxies, so I don't know if it's a good idea to expose these methods by default. Curious what other people thing about this as well!

skidding:
OK, looking to make this official. I wrote this documentation piece and would appreciate your feedback before incorporating it into the README.

Disclaimer: This is meant to be an accurate representation of what the API does, without dwelling much on the motivation behind this and affiliated techniques. I plan to write a Medium article to expand more on that. But in the meantime I'd like to merge this (if everything works as expected) because it contains a rich diff and want to unblock future development.

OK, here it goes. Let me know what you think!

---

## Headless testing

> Add `react-cosmos-test` to your dev dependencies for this API.

Besides showing up in the Playground UI, fixtures can also be used independently with the `createContext` API.

### Using Enzyme

The test API exposes an entry point specifically designed for [Enzyme](http://airbnb.io/enzyme/).

```js
import createContext from 'react-cosmos-test/enzyme';
import fixture from './fixture';

const { mount, getWrapper } = createContext({ fixture });

beforeEach(mount);

test('renders hello', () => {
  expect(getWrapper().text()).toContain('Hello World');
});
```

But this is the only way. As we'll see below, *createContext* can also be used with a custom renderer.

### Using a custom renderer

Here's how to render a fixture with good ol' [react-test-renderer](https://reactjs.org/docs/test-renderer.html).

```js
import { create as renderer } from 'react-test-renderer';
import createContext from 'react-cosmos-test/generic';
import fixture from './fixture';

const { mount, getWrapper } = createContext({ renderer, fixture });

beforeEach(mount);

test('renders hello', () => {
  // Careful, this is no longer an Enzyme wrapper, but a react-test-renderer wrapper!
  expect(getWrapper().toJSON()).toMatchSnapshot();
});
```

### Capturing state changes

The fixture does more than just defining component input. Like a sticky fly trap, it captures state changes that occur during the component's lifecycle, which we can then inspect. For example:

- If Redux state changes, the latest state can be read via `get('reduxState')`
- If Router URL changes, the latest URL can be read via `get('url')`

Instead of polluting our tests with various store and provider initialization, we let the [Proxies](#proxies) take care of it and then collect state changes from the fixture context. Hence the *createContext* name.

### createContext API

The createContext API makes use of already configured proxies, which can be included in more ways.

```js
// Detect proxies automatically by reading cosmos config from cwd
// or via --config CLI arg
const { mount } = createContext({ fixture });

// Or point to a custom config path
const { mount } = createContext({
  fixture,
  cosmosConfigPath: '/path/to/my/special/config';
});

// Or pass proxies directly
const { mount } = createContext({ proxies, fixture });
```

#### Context methods

- *async* `mount` Mounts component via renderer (usually called in `beforeEach`)
- `unmount` Calls unmount method of wrapper returned by renderer
- `getWrapper` Returns wrapper returned by renderer
- `getRef` Get component ref (exclusively for Class components)
- `getField` (or `get` for brevity) Returns updated fixture field

alp82:
> But this is the only way.
You forgot a "not"

skidding:
@alp82 corrected, thanks! 

Let me know if you have any other ideas. I know the name `createContext` raises eyebrows because "context" is a generic term and thus considered bad practice. @maxsalven suggested `createMockedComponent`, and while I'm not particularly a fan because it sounds like we're returning a decorated component (when in fact we're returning a set of wrapper function around the component and fixture data). But I do appreciate the attempt to make it more concrete, so I welcome any other thoughts on this.

skidding:
Also added an example in the "Capturing state changes" section.

alp82:
`createTestWrapper`?
`createComponentApi`?
`componentApiCreator`?

skidding:
> createTestWrapper

🤔

- Is *wrapper* more suggestive than *context*? Honest question, this is a learning experience for me...
- Is `createTestContext` better than `createContext`?

> createComponentApi

This is interesting because `createContext` is actually a lower level abstraction in Cosmos. It is not only used in `react-cosmos-test`, but also in `react-cosmos-loader`, where the renderer is effectively `react-dom`. And the vision is to make eventually extend the renderer interface to support other rendering engines besides React.

Other variations:
- `createComponentWrapper`
- `createComponentContext`

And maybe it would be useful to prefix each entry point separately:
- `createComponentContext` for the lowest level method (not of public interest atm)
- `createTestContext` for `react-cosmos-test/generic`
- `createEnzymeContext` for `react-cosmos-test/enzyme`

Alternatively:
- `createComponentWrapper`
- `createTestWrapper`
- `createEnzymeWrapper`

The distinction ought to also make the examples and codebase easier to follow.

skidding:
> `createTestWrapper` feels like a parent implementation of `createEnzymeWrapper`, but both are on the same level

Not quite. [The Enzyme implementation calls the generic one](https://github.com/react-cosmos/react-cosmos/blob/fc6d4fbd1d4fed3dc5989bcec38f271b7bafad7e/packages/react-cosmos-test/src/enzyme.js#L4-L30), so `createTestWrapper` *is* a parent, which is why you have to pass the renderer yourself.

You're right about `wrapper.getWrapper`.

So to recap one small variation from the current naming would be:

```js
// You'll only use this if you work with Cosmos internals
// - Most logic is here, yet renderer-agnostic
// - Works in any environment (eg browser), no Node deps
import { createComponentContext } from 'react-cosmos-loader';

// You'll use this with react-test-renderer or to build a wrapper for an Enzyme-competitor
// - Expected to be called in Node env, uses modules like `fs` or `path`
// - Looks for cosmos config and tries to automatically include proxies
// - Automatically wraps prop callbacks in jest.fn() when used in Jest env
import createTestContext for `react-cosmos-test/generic`;

// You'll use this with Enzyme
// - Includes Enzyme renderer
// - Returns inner wrapper of fixture component instead of root wrapper of first proxy
// - getWrapper can receive a selector as arg (eg. getWrapper('.btn'))
import createEnzymeContext for `react-cosmos-test/enzyme`;
```

They all build on each other, from bottom to top.

skidding:
I actually realized that `createEnzymeContext` was a bad idea. It's not an Enzyme context, it's the same Cosmos context in all cases, Enzyme is just the type of the returned wrapper. So anyway, I settled on `createTestContext` for now because it does some test-oriented heuristics and it's part of the test API. But, it the end we're talking about the default export which we can always rename later without breaking the API. It's just a matter of how we present examples, it can be named anything.

Did the following:
- [ ] Added the README section
- [ ] Added `getField` alias to `get`
- [ ] Updated examples to use `createTestContext`
- [ ] Published latest to `3.1.0-3` (`next` dist)

Will merge later today and release official `3.1`. Thanks everyone!

maxsalven:
Sorry a little bit last minute; is there a possibility to specify a different `cosmos.proxies.js` file for the test pipeline?

jozsi:
@maxsalven - as far as I know, you can specify a different config file (that sets a different proxy file).
```
// Or point to a custom config path
const { mount } = createContext({
  fixture,
  cosmosConfigPath: '/path/to/my/special/config';
});
```

maxsalven:
> as far as I know, you can specify a different config file (that sets a different proxy file).

I'd like to not have to manually specify a special config path each time, but rather just 'all tests use this proxy config'

skidding:
> I'd like to not have to manually specify a special config path each time, but rather just 'all tests use this proxy config'

Option 1: Create a thin wrapper where you pass the `customConfigPath` or import proxies and pass them directly via `proxies`. Drawback: Imports to this wrapper will be relative instead of `react-cosmos-test/*`. If you use aliases it's not a problem.

Option 2: You could run the tests with `--config /path/to/other.cosmos.config.js`. Drawback: Maybe `--config` conflicts with Jest? I'd be happy to accept a PR where `--cosmos-config` is also supported as a CLI arg.

skidding:
FYI, I wrote a post about the idea behind / into to the test API. You can read the draft and provide feedback here: https://medium.com/@skidding/testing-react-components-30516bc6a1b3

Hint: If you leave any comments using Medium your name will appear under credits. Thanks!

Issue #521, opened 2017-11-16
Dismiss run time error overlay on fixture change


skidding:
### What's up?

After the CRA runtime error overlay is triggered it stays open indefinitely until we click on the X button. This makes sense in a normal app because runtime errors can corrupt the App state. However, in Cosmos we can move from one component to another and the error overlay becomes irrelevant.

### Mkay, tell me more...

To close the error overlay programmatically we can import and call the `dismissRuntimeErrors` function from the `react-error-overlay` package. **Note: [this PR](https://github.com/facebookincubator/create-react-app/pull/3414) needs to be published for this method to be available.**

I see two possibilities here:
1. Automatically close error overlay when component changes
2. Automatically close error overlay when fixture changes

The 1st is more conservative, but I think we can safely go with the 2nd as well. The only risk when dismissing the runtime overlay and opening a fixture of the same component is that we might get the same error and the overlay will be triggered again.

Suggestion for where to implement this: In the [Router's onChange hander](https://github.com/react-cosmos/react-cosmos/blob/2c71da7f489080837229eff70e86bf5020d3db5c/packages/react-cosmos-playground/src/index.js#L40), which gets triggered when the URL changes and receives `component` and `fixture` as params.

maxsalven:
A related item, if you click on a fixture and get the overlay, then pressing escape doesn't clear the overlay until you click on the iframe. Would be nice to wire up an escape click in the FixtureList to close the overlay as well.

jozsi:
Done via #535 - still depends on the new version of `create-react-app`

jozsi:
Updated `react-error-overlay` via #574

PR #535, opened 2017-11-26
Dismiss runtime error on fixture change - #521

jozsi:
Beginning of the implementation for [Dismiss run time error overlay on fixture change #521](https://github.com/react-cosmos/react-cosmos/issues/521)

Work in progress.

Notes/questions:
- [ ] `react-error-overlay` still needs to be published with [@skidding's PR](https://github.com/facebookincubator/create-react-app/pull/3414), until then we are assuming it will be versioned `v3.1.0`
- [x] `react-error-overlay` is imported in `react-cosmos-loader` but is in the dependency list of `react-cosmos` - should it be moved to `react-cosmos-shared`?
- [x] preferably `react-cosmos-loader` shouldn't even know about `react-error-overlay` - how to tackle this?
- [x] tests?

Big thanks to @skidding for his 24/7 availability to help!

### DIFF
--- a/packages/react-cosmos-loader/src/__tests__/mount.js
+++ b/packages/react-cosmos-loader/src/__tests__/mount.js
@@ -7,6 +7,7 @@ const mockFixture = {};
 const mockProxy = () => {};
 const mockStateProxy = () => {};
 const mockErrorCatchProxy = () => {};
+const mockDismissRuntimeErrors = () => {};
 
 jest.mock('react', () => ({
   Component: jest.fn(),
@@ -32,7 +33,8 @@ describe('without container query selector', () => {
         foo: {
           bar: mockFixture
         }
-      }
+      },
+      dismissRuntimeErrors: mockDismissRuntimeErrors
     });
   });
 
@@ -70,6 +72,11 @@ describe('without container query selector', () => {
     const container = render.mock.calls[0][1];
     expect(container.parentNode).toBe(document.body);
   });
+
+  it('passes dismissRuntimeErrors to loader element', () => {
+    const { dismissRuntimeErrors } = createElement.mock.calls[0][1];
+    expect(dismissRuntimeErrors).toBe(mockDismissRuntimeErrors);
+  });
 });
 
 describe('with container query selector and class name', () => {
@@ -87,7 +94,8 @@ describe('with container query selector and class name', () => {
           bar: mockFixture
         }
       },
-      containerQuerySelector: '#app123'
+      containerQuerySelector: '#app123',
+      dismissRuntimeErrors: mockDismissRuntimeErrors
     });
   });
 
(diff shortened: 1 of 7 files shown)


skidding:
> react-error-overlay is imported in react-cosmos-loader but is in the dependency list of react-cosmos - should it be moved to react-cosmos-shared?

I think the cleanest way would be to keep `react-error-overlay` as a `react-cosmos` dependency, and to pass the `dismissRuntimeErrors` function as a prop to RemoteLoader. [Here](https://github.com/react-cosmos/react-cosmos/blob/9b93fd80525c38c570b4e3824932881198ab562c/packages/react-cosmos/src/client/mount.js#L45) and then [here](https://github.com/react-cosmos/react-cosmos/blob/9b93fd80525c38c570b4e3824932881198ab562c/packages/react-cosmos-loader/src/mount.js#L39).

>  preferably react-cosmos-loader shouldn't even know about react-error-overlay - how to tackle this?

I agree, see above.

> tests?

- In the RemoteLoader, you can add a test [here](https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-loader/src/components/RemoteLoader/__tests__/fixture-select.js#L113). Basically you add a new prop that is the dismissErrorOverlay callback (eg. `dismissErrorOverlayMock = jest.fn()`) in the setup part and then `expect(dismissErrorOverlayMock).toHaveBeenCalled()`
- In `react-cosmos` package it's even easier. Like [this](https://github.com/react-cosmos/react-cosmos/blob/9b93fd80525c38c570b4e3824932881198ab562c/packages/react-cosmos/src/client/__tests__/mount/new-fixtures.js#L81-L86), but you check that the prop value is equal to the `dismissRuntimeErrors`, which you import normally from `react-error-overlay` in the test file.

jozsi:
@skidding - done. Looking for a quick PR.

Tests will fail until `react-error-overlay` gets published.

skidding:
Looks great. 

Unorthodox idea: You can replace the REO import with a dummy function in `react-cosmos/src/client/mount.js` and mark the test as `test.skip` so we can merge as is. I say this because I massacred `RemoteLoader` in #534 and it's best for me to adapt your code than to completely invalidate this PR if I complete my work before REO 3.1 is published. If you want to be more dramatic you can put a console.log in that dummy function saying that we're waiting for https://github.com/facebookincubator/create-react-app/pull/3414 to be published 😅. But no rush, it will likely take me a few more days to get there.

jozsi:
Done. Interestingly, some tests (unrelated to this PR) don't pass neither on my machine, nor Travis.

Note: when REO gets published (and will be versioned as v3.1.0), we need to revert commit 57e5feb

skidding:
Looks good. 

> Interestingly, some tests (unrelated to this PR) don't pass neither on my machine, nor Travis.

I was confused at first as well, but those tests *are* related to this PR. You're not passing `dismissRuntimeErrors` to RemoteLoader in some of its test files. The confusing part is that it times out instead of showing the runtime errors. This is because it crashes inside a `beforeEach` Promise, which apparently doesn't do anything on the `reject` path, except to stall the test indefinitely. Sucks, but good to know.

Solution: Either add an empty function as the default value for `dismissRuntimeErrors` in RemoteLoader, or pass a mock in every test file.

jozsi:
Nice catch! I've set `dismissRuntimeErrors` as `defaultProps`, passes now 😁

skidding:
@jozsi congrats, you have just shipped a future feature! 😄

PR #549, opened 2017-12-11
Respond to fixture changes

skidding:


skidding:
OK, it seems like something was fishy with hot reloading since I merged the Test API, so I put in some extra time to iron out every use case. Here are 4 hot reloading scenarios covered by this PR:

- Fixture source changes apply immediately initially
- **After component state changed** fixture source changes are kept until fixture is re-selected
- After fixture is re-selected fixture source changes apply immediately again

fixture-update-change

- Fixture source changes apply immediately initially
- **After fixture has been edited** fixture source changes are kept until fixture is re-selected
- After fixture is re-selected fixture source changes apply immediately again

fixture-edit-change

- Component source changes are always applied immediately, even when state updates are active

fixture-component-change

- Component source changes are always applied immediately, even when edits are active

fixture-component-change-edit

---

@maxsalven can you please test this via `react-cosmos@3.2.0-3`?

It includes a fix for the rename fixture error. 🙏

Issue #151, opened 2015-08-14
Relay-ification


KyleAMathews:
With Relay and the brave new world of colocated component query fragments, I was thinking Cosmos could get an interesting Relay extension.

Basically, for many/most Relay-ified components, a Cosmo fixture wouldn't be necessary. Instead you could wire it directly to your GraphQL endpoint and use live production data.

What would be fun/useful is if by default, Cosmo would load a random entity to populate the component. This would be a real-world stress test for components you're developing/tweaking.

Even cooler is if there was a button that would trigger a new fetch of a random entity. With that you could quickly toggle through live production data and see how your component responds to each. I'd want a list of each component that you'd loaded to maintained so you could quickly toggle back to previous entries as say you found errors in several of them and are cycling through the entities with errors fixing problems.

A stretch goal would be to add an autocomplete field to search (in some flexibly defined way) for entities to view. E.g. when developing a user profile widget, search for specific users by name.

Thoughts?

dzannotti:
+1

KyleAMathews:
Another idea. Be able to star and label particular objects that you could come back to e.g. an entity that's prone to bugginess. These could show up where fixtures are now.

KyleAMathews:
Oooooo... and what if could create a grid wall of components. Say assign each component to a 100x200 space and each one pulling in a random entity. Or test a component's responsive design by doing some sort of masonary-style layout where components are assigned random heights and widths (well... random within defined constraints). On a large screen you could have dozens or even hundreds of instances of a component loaded that you're hot-reload editing code and styles. That'd be pretty epic. Something with sliders like I'm doing on my demo here: http://kyleamathews.github.io/react-component-gallery/

skidding:
> With Relay and the brave new world of colocated component query fragments, I was thinking Cosmos could get an interesting Relay extension...

I like the Relay extension. Note that an advantage of the component playground is the offline/isolated dev env. Still, having both offline fixtures and the option to hook up to Relay endpoints would be double useful.

That being said, the easiest way to add this extension at the moment would be in [the lib that generates the function tree for ComponentPlayground](https://github.com/skidding/cosmos/blob/60bc9fa893b79cbb404e0b028916a4ec8aa09287/component-playground/lib/get-component-fixture-tree.js). There's already a [PR](https://github.com/skidding/cosmos/pull/136) to automatically generate an empty fixture if no fixtures are found for a component. So the trend to make previously defined fixtures optional is already emerging!

[...]

PR #552, opened 2017-12-14
Apollo Client finer integration

### LINKED ISSUE #445: Fine-tune the Apollo Proxy with custom responses
Awesome work on the apollo proxy!

As an idea for an improvement, can we locate the mock response in the fixture file instead of in the proxy config? I think this makes more sense and aligns better with the way the other proxies work. It would allow me to mock multiple different server responses by using different fixtures, and I don't have to worry about the proxy config and my fixtures getting out of sync (e.g. if I remove a component from cosmos, I now have to remember to modify the proxy config to remove the mock).

xavcz:
Fixes #445 

screencast

This PR adds the ability to:
- change locally the Apollo cache to mocks responses and re-run queries / mutations based on this cache
- pass your own link (ex: `HttpLink`) to connect to a GraphQL API and get real responses

The Apollo Proxy listens to cache changes (thanks to the Apollo DevTools hook provided in Apollo Client 😇) and sends it to the fixture editor under the key `apollo`. 

This makes the playground reloads and the Apollo Proxy re-instantiate a client with a restored cache from the fixture: this means you can also add the result of `cache.extract()` manually to your fixture. I'm using a method as similar as [the one recommended for server-side rendering](https://www.apollographql.com/docs/react/basics/caching.html#server) 💃  

It may not be the optimal solution to mock components fetching data through React Apollo, but that's a first step forward! 

If people really want to mock with a specific result, we may provide a [client like the one shown in this article](https://dev-blog.apollodata.com/query-components-with-apollo-ec603188c157?gi=a09002b53f12#d024)?

What are your thoughts? cc @maxsalven @dan-weaver 

Remaining: 
- [x] update README if we believe it's a valid step forward 🌮 
- [x] test it in a way that makes sense (not like now)


maxsalven:
This looks very cool, especially being able to get real API responses if you want to (I can see myself having a "LIVE" version of many fixtures, or perhaps having a visual UI toggle somewhere in another proxy that toggles all fixtures that depend on network requests to either get a real response or serve the fixture response).

Apologies for not having contributed much to the GQL discussion of late; I've only played with it as a toy concept so far, but in a few weeks will be starting to bring GQL into our production system. At that point I'll have more concrete feedback. My initial reaction is that I'd still like to mock the actual GQL server response, but let's see where my thinking goes once I've actually used this in a 'real world' scenario.

examples/apollo/package.json
```diff
@@ -8,7 +8,12 @@
     "start": "cosmos"
   },
   "dependencies": {
-    "react-apollo": "^1.4.15",
+    "apollo-cache-inmemory": "^1.1.4",
```
skidding: Does the consuming repo need this dependency?

packages/react-cosmos-apollo-proxy/src/index.js
```diff
@@ -1,45 +1,71 @@
...
+          Pass it a local schema built with graphql-tools or a link pointing to a GraphQL endpoint.
+          Read more at: https://github.com/react-cosmos/react-cosmos#react-apollo-graphql.`
+        );
+      }
+
+      if (schema && link) {
+        throw new Error(
+          `It looks like the Apollo Proxy is configured with both a schema & an Apollo Link!
+          You can  either just pass a local schema instance, or build your own link to connect to the schema you want to use.
+          Read more at: https://github.com/react-cosmos/react-cosmos#react-apollo-graphql.`
```
skidding: ❤️ing these verbose errors.

skidding:
This looks cool!

I'm not a real user so I can't give valuable user feedback. The raw cache seems a bit difficult to approach, so I don't see myself writing Apollo cache in fixture by hand, but in the end does the old (existing) proxy have any advantage over the newer one?

I expect the "live" version to be popular, though. To quote @KyleAMathews from an [ancient issue](https://github.com/react-cosmos/react-cosmos/issues/151):

> With Relay and the brave new world of colocated component query fragments, I was thinking Cosmos could get an interesting Relay extension.

> Basically, for many/most Relay-ified components, a Cosmo fixture wouldn't be necessary. Instead you could wire it directly to your GraphQL endpoint and use live production data.

> What would be fun/useful is if by default, Cosmos would load a random entity to populate the component. This would be a real-world stress test for components you're developing/tweaking.

---

> @maxsalven My initial reaction is that I'd still like to mock the actual GQL server response

What would be the difference between this and mocking via fetch or XHR proxy?

examples/apollo/package.json
xavcz: Surely not! Thanks! 🎩

xavcz:
> 🤗 **Thanks for all these feedbacks, folks! It gives a clearer picture!**

Recap
---
The goal of the Apollo proxy is to help developing components in association with a GraphQL API, backed by the Apollo Client and the React integration: 
- you can stress your components with _live data_, the results come from an external API;
- you can prototype your components with _local data_, you define the shape of the results you'll get.

Flow
---
Here is a diagram representing the API I think it'll end up with:

diagram

"Easy-peasy" config (`proxy.config.js`)
---
The most simple way to use the proxy is to pass a GraphQL `endpoint`. A client is created for you to send operations over HTTP (`HttpLink`) to this endpoint and it depends on a classic `InMemoryCache`.

```js
// proxy.config.js

export default [
	createApolloProxy({
	  endpoint: 'https://1jzxrj179.lp.gql.zone/graphql'
	}),
	// ...other proxies
]
```

"Real-world" config (`proxy.config.js`)
---
A client used in a production app often uses multiple links (polling, retry, context, whatever...) and may have some specific options for the cache. We allow to overwrite the default client by your own:
```js
// proxy.config.js
import { client } from './src/client.js';

export default [
  createApolloProxy({ 
	client
  }),
  // ...other proxies
]
```

"API Prototype" config (`proxy.config.js`)
---
Your API is no ready yet, no worries, prototype it directly in Cosmos! The proxy accepts a `schema` instance to run operations against. It uses the [`apollo-schema-link`](https://www.apollographql.com/docs/link/links/schema.html) under the hood: direct execution against `graphql-js`.

```js
// proxy.config.js
import { makeExecutableSchema, addMockFunctionsToSchema } from 'graphql-tools';

const typeDefs = `
  type Query {
    hello(who: String): String
  }
`;
 
const mocks = {
  Query: () => ({
    hello: (root, { who }) => `Hello ${who ? who : 'C O S M O S'}`
  })
};

const schema = makeExecutableSchema({ typeDefs });
 
addMockFunctionsToSchema({ schema, mocks });

export default [
  createApolloProxy({
    schema
  }),
  // ...other proxies
];
```

"Fine-tune" fixture (`xyz.fixture.js`)
---
Let's say that we have this component where we want to directly mock the GraphQL response (cc @maxsalven 😛). We can do that by writing fixtures, yeah!

Like we have this query:
```graphql
query PostsForAuthor($authorId: Int!) {
      author(id: $authorId) {
        id
        firstName
        posts {
          id
          title
          claps
        }
      }
    }
```

```js
// MyComponent.fixture.js
import MyComponent from './MyComponent';

export default {
	component: MyComponent,
	apollo: {
		resolveWith: {
			author: {
				id: 123,
				firstName: 'Ovidiu',
				posts: [
					{ id: 456, title: 'Testing React Components', claps: 1234 },
					{ id: 789, title: 'When to assert?', claps: 56 },
				]
			}	
		}
	}
}
```

How does these config live together?
---
`endpoint` is the highest-level.
`endpoint` is overwritten by `client`.
`client` is overwritten by `schema`.
`schema` overwrites any config.

maxsalven:
Great update, thanks @xavcz.

Questions:

- If you have a schema in your proxy config ("client is overwritten by schema"), does this mean you can't simultaneously have fixtures that access live data? 

- For the fixtures with directly mocked responses, perhaps it would be better to somehow use the query name as the key under `resolveWith` (or maybe we can just call the root key `apolloMocks` and then under that you have `{ PostsForAuthor: { author { ... } } }`). This means you could mock a component which performs multiple different queries.

- Is there a way to mock the 'loading' state? 

@skidding 
> What would be the difference between this and mocking via fetch or XHR proxy?

You could indeed directly mock the responses at this level, however all requests go to the same endpoint (with the query itself as part of the `POST` payload), so dealing with a component that performs multiple different queries would be a bit unwieldy (you'd have to write code to manually inspect the requests).

maxsalven:
In the 'direct mocking' approach, would also be cool if you could pass in a function that accepts the query's variables:

```
apolloMocks: {
  PostsForAuthor: (authorId) => ({
    id: authorId,
    firstName: "Max",
  })
}

**Comment by xavcz** [comment]
> If you have a schema in your proxy config ("client is overwritten by schema"), does this mean you can't simultaneously have fixtures that access live data?

👍

I guess we could stitch real data with the local schema, yes, good call!

> For the fixtures with directly mocked responses, perhaps it would be better to somehow use the query name as the key under resolveWith [...] This means you could mock a component which performs multiple different queries.

🤐

`resolveWith` would be the result you expect to get under the `data` props. If you need multiple queries, I'd recommend to can combine them like this:
```graphql
# definition, api-side
type Query {
	author: Author
	pizzas: [Pizza]
}

↓ ↓ ↓ ↓

# consumption, ui-side
query twoQueriesInOneNetworkRequest {
	author {
		# ...
	}
	pizzas {
		# ...
	}
}
```

You'll get an object `{ data: { author: { ... }, pizzas: [ ... ], ... } }` and have two queries sent in one network request. If however you really want to send two different queries as two network requests, well, you can test the HOC directly, the proxy wouldn't be useful for the composed component.

 > Is there a way to mock the 'loading' state?

🤐

A component broadcasting GraphQL operations with Apollo is basically a component with a HOC around it. To test the "loading state", you can test the loading _prop_ the inner component receives 🍡 I see your point, maybe in terms of DX it could be cool to do that directly from the fixture, however it won't be included out-of-the-box in this PR.

> In the 'direct mocking' approach, would also be cool if you could pass in a function that accepts the query's variables

🤠

Ooh, nice! That would be pretty neat, however, we would lose the ability to see it in the fixture editor, right?

xavcz:
Updated the API to match my initial comment with a slightly different priority order.

In the proxy config
1. Try to use `client` option, the real client you use in your app
2. If no `client`, looks for an `endpoint` that you may want to consume
3. If no `client` nor `endpoint`, create a link to fetch a local `schema`

If no `client || endpoint || schema`, the proxy throws: GraphQL is kind of a contract between your client and an API, and there is no way to know what is the contract.

On the way to send the operation, if `apollo.resolveWith/failWith` is defined in the component's fixture, replace the link by a "fixture" link [inspired from this article](https://dev-blog.apollodata.com/query-components-with-apollo-ec603188c157) that either give a result or an error. 

Could you give it a try and see how it feels? You can run the `examples/apollo` app.

> I guess we could stitch real data with the local schema, yes, good call!

We may run into problems when we have mock function in the local schema that are the same that the real one 🤔

skidding:
Look great! Looking forward to feedback from experience GraphQL devs.

examples/apollo/cosmos.proxies.js
```diff
@@ -1,20 +1,10 @@
...
-  }
-`;
-
-const mocks = {
-  Query: () => ({
-    hello: (root, { who }) => `Hello ${who ? who : 'C O S M O S'}`
-  })
-};
+// see more examples at:
+// https://github.com/react-cosmos/react-cosmos#react-apollo-graphql
```
skidding: How about having another example with local mocks? It should also make development easier when testing all types of mocking.

xavcz: Actually, I'm thinking about removing the local mocks part!

GraphQL is all about one source of truth, your schema, and creating local mocks in a separate client app feels like creating a second source of truth, that can easily drift off the alignment with the original API

skidding: > Actually, I'm thinking about removing the local mocks part!

Makes sense!

dan-weaver:
Hey there. How can I help? This looks good to me so far.. It's a lot like my hand rolled solution I'm currently using but yours is more robust :).  

question
>Actually, I'm thinking about removing the local mocks part!

does that mean your removing the ability to do something like this?
https://github.com/react-cosmos/react-cosmos/pull/552/files#diff-7062272aa1857f29f5d4365b65596ef5R9
 Maybe I'm not following, but I find it really valuable to be able to mock at the fixture level!

packages/react-cosmos-apollo-proxy/src/index.js
```diff
@@ -1,43 +1,50 @@
...
+Read more at: https://github.com/react-cosmos/react-cosmos#react-apollo-graphql.`
+        );
+      }
+
+      const { resolveWith, failWith } = this.props.fixture[fixtureKey] || {};
+
+      this.client =
+        client ||
+        new ApolloClient({
+          cache: new InMemoryCache(),
```
dan-weaver: Something I ran into recently in my "hand rolled" version of this.  You probably are going to need to paramaterize dataIdFromObject at some point.

maxsalven:
Not sure it helps anyone at all, but just to share what my current solution is.

I have a very simple proxy:
```
import React from "react";
import { ApolloProvider } from "react-apollo";
import apolloClient from "gql/apolloClient";

const apolloClientInstance = apolloClient();
parent.__APOLLO_CLIENT__ = apolloClientInstance; // eslint-disable-line

export default () => {
  const ApolloProxy = (props: ProxyProps) => {
    const { nextProxy, fixture } = props;
    const { value: NextProxy, next } = nextProxy;

    const nextProxyEl = <NextProxy {...props} nextProxy={next()} />;

    if (!fixture.apollo) {
      return nextProxyEl;
    }

    return (
      <ApolloProvider client={apolloClientInstance}>
        {nextProxyEl}
      </ApolloProvider>
    );
  };

  return ApolloProxy;
};
```

and then I use it in conjunction with the fetch proxy to mock the response:

```
{
    component: HomeScreenContainer,
    name: "Mocked Response",
    url: "/?date=2018-02-01",
    apollo: true,
    props: {},
    fetch: [
      {
        matcher: "/api/graphql",
        response: { data: mockResponse },
      },
    ],
  },
```

or I also have 'Live' fixtures that talk to the actual API:

```
  {
    component: HomeScreenContainer,
    name: "Live - June 01, 2017",
    url: "/?date=2017-06-01",
    apollo: true,
    usesAPI: true,
    props: {},
  },
```

I haven't experimented with any sort of advanced mocking, but if this proxy could satisfying these two simple use cases, I would be happy to switch to it.

xavcz:
Folks, thanks so much for your feedbacks! ❤️ 

 @dan-weaver By "removing local mocks" I'm thinking of getting rid of this part:

```
// proxy.config.js
import { makeExecutableSchema, addMockFunctionsToSchema } from 'graphql-tools';

const typeDefs = `
  type Query {
    hello(who: String): String
  }
`;
 
const mocks = {
  Query: () => ({
    hello: (root, { who }) => `Hello ${who ? who : 'C O S M O S'}`
  })
};

const schema = makeExecutableSchema({ typeDefs });
 
addMockFunctionsToSchema({ schema, mocks });
```

You shouldn't need to create an abstract layer above your existing API in your cosmos config. Mocking would happen only at the level of the fixture, like the code you highlighted 😊 

@maxsalven that's super interesting! Yeah, the goal is to support sending operations and either get a mocked response or a live response. I imagine that with the fetchproxy solution, it would be harder to get mocks based on query variables or mocks for failing operations 🤔

I'll come up with some updates tomorrow, let's get this PR merged asap! 😁

xavcz:
> 🚀 Ready to be merged! (_4 months later, haha_)

I believe it's good now:
- [added tests around the proxy / custom link](https://github.com/react-cosmos/react-cosmos/blob/341985f15628cef3484abfa52ef8905ef035b387/packages/react-cosmos-apollo-proxy/src/__tests__/client.js), 
- [updated the docs](https://github.com/react-cosmos/react-cosmos/blob/341985f15628cef3484abfa52ef8905ef035b387/README.md#react-apollo-graphql)
- [completed the example with live behavior / mocked (named) responses](https://github.com/react-cosmos/react-cosmos/tree/341985f15628cef3484abfa52ef8905ef035b387/examples/apollo/components/__fixtures__/Author)

What do you think?

cfnelson:
@xavcz Looks amazing! 🔥 Can't wait to see this merged & released!!! I gave a quick read through and loving the Docs & examples! Can I do anything to help move this along?

xavcz:
Oh mate, thanks a lot!

Maybe having a look at the failing build on node 6? It fails because of async await in a dependency? 😐

skidding:
@xavcz I think I solved the CI issues in https://github.com/react-cosmos/react-cosmos/pull/619!

Feel free to merge my patch or replicate the changes on your branch. But don't merge this PR just yet as I haven't reviewed all code yet. 🤓

packages/react-cosmos-apollo-proxy/src/__tests__/client.js
```diff
@@ -0,0 +1,179 @@
...
+      proxyConfig: {
+        endpoint: 'https://xyz'
+      }
+    });
+
+    expect(fetchMock.called('https://xyz', 'POST')).toBe(true);
+
+    // wait for the fake network request to complete
+    await until(() => {
+      wrapper.update();
```
skidding: Maybe put `wrapper.update();` inside `getWrappedComponent`, before returning?

packages/react-cosmos-apollo-proxy/src/__tests__/mocking.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
```
skidding: Seems like this test file is more related to rendering the other proxies and adding the ApolloProvider, so maybe rename to `render.js`?

README.md
```diff
@@ -465,46 +465,181 @@ Check out the [React Router example](examples/react-router) to see the proxy in
...
 #### React Apollo (GraphQL)
 
-If you use the [React integration](http://dev.apollodata.com/react/) of [Apollo Client](http://dev.apollodata.com/) to provide data in your app, you may want to provide mocks for isolated UI testing with GraphQL.
-Your components wrapped with the `graphql` higher-order component provided by `react-apollo` depends on the `ApolloProvider` defined at the top-level of your app. This proxy does that for you!
+If you use the [React integration](http://dev.apollodata.com/react/) of [Apollo Client](http://dev.apollodata.com/) to provide data in your app, you may want to:
+
+* work on your data components in isolation ;
+* provide static or dynamic mocks to prototype your components:
+
+This proxy wraps your components with the `ApolloProvider` so they can render in Cosmos like they would normally in your app, and it .
```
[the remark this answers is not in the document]
xavcz: Lol, I don't know 😱

README.md
```diff
@@ -465,46 +465,181 @@ Check out the [React Router example](examples/react-router) to see the proxy in
 
 #### React Apollo (GraphQL)
 
-If you use the [React integration](http://dev.apollodata.com/react/) of [Apollo Client](http://dev.apollodata.com/) to provide data in your app, you may want to provide mocks for isolated UI testing with GraphQL.
-Your components wrapped with the `graphql` higher-order component provided by `react-apollo` depends on the `ApolloProvider` defined at the top-level of your app. This proxy does that for you!
+If you use the [React integration](http://dev.apollodata.com/react/) of [Apollo Client](http://dev.apollodata.com/) to provide data in your app, you may want to:
+
+* work on your data components in isolation ;
```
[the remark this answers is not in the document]
xavcz: 👍 consistency!

README.md
```diff
@@ -465,46 +465,181 @@ Check out the [React Router example](examples/react-router) to see the proxy in
...
+            votes: post.votes + 10
+          }
+        };
+      }
+    }
+  }
+};
+```
+
+### Failing response
```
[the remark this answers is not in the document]
xavcz: 👍 consistency!

README.md
```diff
@@ -465,46 +465,181 @@ Check out the [React Router example](examples/react-router) to see the proxy in
...
+  apollo: {
+    failWith: {
+      message: 'Something went bad, please try again!'
+    }
+  }
+};
+```
 
-Check out the [Apollo example](examples/apollo) to see `react-cosmos-apollo-proxy` in action.
+Check out the [Apollo example](examples/apollo) to see `react-cosmos-apollo-proxy` in action! 🚀
```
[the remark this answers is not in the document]
xavcz: 🎩

examples/apollo/components/Author.js
```diff
@@ -0,0 +1,73 @@
...
+              )}
+            </li>
+          ))}
+        </ul>
+      </div>
+    );
+  }
+}
+
+export default compose(withData, withMutation)(Author);
```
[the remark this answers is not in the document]
xavcz: Yeah, it doesn't have a `displayName` by default, [however we can add it](https://github.com/apollographql/react-apollo/search?utf8=%E2%9C%93&q=displayName). It's actually one of the few (minor) feature I helped for!

examples/apollo/cosmos.proxies.js
```diff
@@ -1,20 +1,10 @@
...
 
 export default [
   createApolloProxy({
-    typeDefs,
-    mocks
+    endpoint: 'https://1jzxrj179.lp.gql.zone/graphql'
```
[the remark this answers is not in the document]
xavcz: Definitely!

examples/apollo/package.json
```diff
@@ -8,8 +8,11 @@
     "start": "cosmos"
   },
   "dependencies": {
-    "react-apollo": "^1.4.15",
-    "react-cosmos": "^3.7.1",
+    "graphql": "^0.13.1",
+    "graphql-tag": "^2.7.3",
+    "graphql-tools": "^2.14.1",
+    "react-apollo": "^2.0.4",
+    "react-cosmos": "^3.2.0",
```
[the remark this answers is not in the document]
xavcz: 👍 consistency

packages/react-cosmos-apollo-proxy/src/__tests__/mocking.js
xavcz: Nice one, thanks!

packages/react-cosmos-apollo-proxy/src/__tests__/client.js
xavcz: Oooh 🤯

examples/apollo/cosmos.proxies.js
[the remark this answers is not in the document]
xavcz: > Hmm, doing a live call in tests can lead to lots of pain down the road, because all of a sudden the test suit doesn't down offline nor if gql.zone is down.

You are the voice of the reason, excellent point. I agree with your approach.

I replicate the `fetch-proxy` behavior in the tests (they fetch `http://xyz` and no real call is made when running the tests).

In order to give the user a feeling of the real behavior, I'm going to download the live API to create a local GraphQL API palce in a `.local-graphql-api` directory, inside the example.

☺️ This ensures pure encapsulation for the tests as well as high-fidelity example

xavcz:
Many thanks for this review!

It should be good to go now. 

Happy to give you a walkthrough before we merge it, if you are available (especially concerning [this comment](https://github.com/react-cosmos/react-cosmos/pull/552#discussion_r174950485))!

skidding:
Thanks for making all the changes!

I'll look closely at the fetch mock-related flaky test and see if I can make it solid.

Any idea about the 1st mount vs 2nd mount inconsistency in the Cosmos UI with the Apollo proxy?
apollo-error

cfnelson:
I noticed that the `real-error` fixture exhibited the same behaviour where the thrown error being logged to console after switching between fixtures. 

See the GIF below:
screen recording 2018-03-16 at 10 32 am

skidding:
@cfnelson yes! I've been experiencing that as well. I tried to understand why that's happening but no luck. I appreciate any help getting to the bottom of this as I am not experienced with GraphQL. Thanks!

PR #563, opened 2017-12-20
fix: update fixture documentation 

Jwhiles:
Fixtures can be exported as an array, and also allow for name and namespace properties. These features are not currently mentioned in the documentation. I've added a brief description of these properties, and an example of exporting an array of fixtures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ It is also possible to export an array of fixtures from a single file. The above
 ```js
 export default [{
   component: SearchBox,
-  namespace: 'Searching for pets'
+  namespace: 'Searching for pets',
   name: 'Dog search',
   state: {
     searchQuery: 'Who let the dogs out?'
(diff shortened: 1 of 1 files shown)


skidding:
Hi @Jwhiles,

Thanks for this. I think it's valuable to surface this information, but here are my suggestions regarding content and styling:
- How about mentioning the possibility of exporting a list of fixtures after the "Where to put fixtures?" sub-section, since it's related to the file system? I'd also like to keep the fixtures in that example as brief as possible, without state, name or namespace which will be explained later independently.
- Maybe rename "Other fixture Properties" to "Fixture name and namespace" and focus on the following information:
  - The name and namespace are detected automatically from the file name and file path respectively, but they can be overridden with custom values
  - The name and namespace are used in the Tree View navigation of the Playground UI

skidding:
I made the amendments myself. Thanks.

PR #586, opened 2018-02-07
[Draft] Basic Redux Form example based on https://codesandbox.io/s/mZRjw05yp

kfern:
Basic Redux Form example based on https://codesandbox.io/s/mZRjw05yp

rsz_2018-02-07_20_16_24-reduxform_default_ _react_cosmos

TODO:

- Testing?


skidding:
@kfern This is great, thank you for taking care of such an important chore! 👍 

Here are my thoughts:
- Can you put the createStore function and reducer in a separate file? Eg. `store.js` or `configureStore.js` it's usually called. This reflects a real life example more and I believe would make the Cosmos parts from cosmos.proxies.js clearer.
- You can safely remove `cosmos.config.js`. Cosmos is deliberately supporting not needing a config file when relying on defaults, and not all examples have a config file already.

No need for tests for this example. In the future I'd like to convert the Cypress tests, that currently run only over the "local-state" example, to spawn multiple jobs and run tests over all examples in parallel. But I don't know what the best approach for this is and it's definitely out of scope.

Finally, a README explaining the contents of the example would be ideal, but also optional. I've seen this in other projects and I thought they were very useful, but we didn't do this for any of the existing examples so far, so it would be a first. Completely up to you.

🤘

kfern:
@skidding You are right. I will change the PR

skidding:
@kfern everything looks great! Let me know if you're done and I will merge.

Note: It seems there's a git conflict on the `yarn.lock` file. In case you need help updating your fork here's a quick tutorial from GitHub: https://help.github.com/articles/syncing-a-fork/.

kfern:
@skidding I have updated the PR but I am not sure that I have done it correctly. If the yarn.lock file is correct, the PR is finished

skidding:
Thank you @kfern!

Issue #560, opened 2017-12-18
Current API for defining fixtures doesn't allow fixtures to be typechecked


andreypopp:
I think the current API is not optimal for codebases which use either Flow or TS because it doesn't allow to check fixtures for type errors.

Instead I'm proposing to have a `createFixture` function:
```
function createFixture<P: {}, C: React.ComponentType<P>>(fixture: {
  component: C,
  name?: string,
  props?: P,
  children?: React.Node,
}) {
  return fixture;
}
```

which provide a type witness to Flow/TS how `props` relates to `component`.

Which would make fixtures look like:
```
import {createFixture} from 'react-cosmos'
import MyComponent from '../MyComponent'

export default createFixture({
  component: MyComponent,
  props: {
    label: 'Hello, World'
  }
})
```

Digging Cosmos more and more! Thanks!

skidding:
> which provide a type witness to Flow/TS how props relates to component.

Is that a working example? 

I've been wondering if there's a way to have a generic `Fixture` type that can be used like this: `fixture: Fixture<Component>` and it would know to expect `fixture.props` to match the Props type of Component. But I don't have enough experience with Flow/TS to understand if that's possible.

Btw, I just typed some fixtures for the first time today :P

```js
// @flow
import MyComponent from '../../MyComponent';
import type { Props } from '../../MyComponent';

const fixture: { props: Props } = {
  component: MyComponent,
  props: {
    // secret sauce...
  }
};

export default fixture;
```

Not ideal. Happy to continue discussion on making this more straightforward.

andreypopp:
> Is that a working example?

Yes. I've copy pasted this from my fixture code.

> I've been wondering if there's a way to have a generic Fixture type that can be used like this: fixture: Fixture<Component> and it would know to expect fixture.props to match the Props type of Component. But I don't have enough experience with Flow/TS to understand if that's possible.

The problem with this proposed API is that the type annotation with `Fixture` is optional while using `createFixture` guarantees that fixtures will be type checked.

maxsalven:
@andreypopp I tried using the function you defined, and it works great for type checking (although VSCode shows the errors in the component definition rather than in the fixture, but not a huge deal). However, any fixture I have wrapped in `createFixture` no longer shows up in the correct place in the file tree on the left side of Cosmos. Have you had this issue?

@skidding, if I run `yarn print-fixture-files` then it simply shows `"components": []` for the impacted components.

skidding:
[...]

PR #591, opened 2018-02-13
Support fixture call expression exports

skidding:
Related to #560, specifically @maxsalven's [comment](https://github.com/react-cosmos/react-cosmos/issues/560#issuecomment-353928361). This doesn't bundle the `createFixture` helper into Cosmos, so anyone who wishes to add types to their fixtures will have to host a copy of it for now. But this PR extends the fixture parser to be able to detect component paths from fixtures where a call expression (ie. function call) is exported instead of an object or array of object.


skidding:
Published under `3.7.0`

Issue #614, opened 2018-03-06
Children props in JSX do not work


fusionstrings:
### What's up?

Readme mentions (https://github.com/react-cosmos/react-cosmos#children) that children props of a fixture can have JSX in them.

```js
export default {
  component: Text,
  props: {
    someProp: true,
    children: (
      <div>
        <p>Fixture ain't afraid of JSX</p>
        <p>Fixture ain't afraid of nothin!</p>
      </div>
    )
  }
};
```

This doesn't seem to be working though
### Mkay, tell me more...

I see the following error.
```
×
ReferenceError: React is not defined
(anonymous function)
/Users/<username>/cosmos-test/__fixtures__/text.fixture.js:8
   5 | component: Text,
   6 | props: {
   7 | 	someProp: true,
>  8 | 	children: (
   9 | 		<div>
  10 | 			<p>Fixture ain't afraid of JSX</p>
  11 | 			<p>Fixture ain't afraid of nothin!</p>
```
This doesn't happen when using non-jsx children

skidding:
Hi @fusionstrings,

This is more related to JSX/Babel than Cosmos. Babel compiles JSX into React.createElement() calls, which requires React to be imported in the same file to begin with. Yes, this is a confusing thing that the community hasn't figured out how to solve (to my knowledge).

Try adding this at the beginning of the fixture (only if you have JSX):

```js
import React from 'react';
```

Also, if you're willing, it might be a good idea to submit a PR updating the children fixture example to:

```diff
+import React from 'react';
+
export default {
  component: Text,
  props: {
    someProp: true,
    children: (
      <div>
        <p>Fixture ain't afraid of JSX</p>
        <p>Fixture ain't afraid of nothin!</p>
      </div>
    )
  }
};
```

fusionstrings:
Hey, @skidding Thanks for clarifying. Of course...React needs to be in scope. Also doesn't matter how small the contribution is, thanks for giving the opportunity :)

https://github.com/react-cosmos/react-cosmos/pull/615

PR #615, opened 2018-03-06
Emphasise React needs to be in scope for JSX

fusionstrings:
React needs to be in scope for JSX to work.
Related https://github.com/react-cosmos/react-cosmos/issues/614


skidding:
@fusionstrings thanks!

PR #617, opened 2018-03-13
Includes StateProxy as the last proxy when using Loader's createContext

xavcz:
> The problem solved by this PR has been referenced in many different discussions, but not in a specific issue afaik.

Problem
---
`StateProxy` is only used [when mounting the loader directly](https://github.com/react-cosmos/react-cosmos/blob/763ccd495533a3ff54b525bcf1f589644909187d/packages/react-cosmos-loader/src/mount.js#L25-L33), but not when using the generic way to get the loader with `createContext`.

Let's say you need `StateProxy` to run some assertions on your components: you need to manually add it to your `cosmos.proxies.js` list to have access to it. This will result in a duplication of the proxy on the UI.

(Possible) Solution
---
Include `StateProxy` in `createContext` so it's consistent with the rest of cosmos!

This enables test like the one in the diff 🎉

### DIFF
--- a/packages/react-cosmos-loader/src/create-context.js
+++ b/packages/react-cosmos-loader/src/create-context.js
@@ -2,6 +2,7 @@
 
 import until from 'async-until';
 import React from 'react';
+import createStateProxy from 'react-cosmos-state-proxy';
 import Loader from './components/Loader';
 import { isComponentClass } from './utils/is-component-class';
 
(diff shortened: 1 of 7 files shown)


packages/react-cosmos-loader/src/create-context.js
```diff
@@ -82,9 +84,13 @@ export function createContext(args: ContextArgs): ContextFunctions {
           updatedFixture = { ...fixture };
         }
 
+        if (!StateProxy) {
```
skidding: I don't know why I did this lazy initialization in the past, but how about just doing `const StateProxy = createStateProxy({ updateInterval: 50 });` once, outside `createContext`?

xavcz: Yep

Issue #610, opened 2018-03-05
No fixtures detected when a cwd path includes dotfile.


maruware:
### What's up?

No fixtures detected when a cwd path includes dotfile.
e.g. /Users/maruware/.ghq/github.com/maruware/project-x

### Mkay, tell me more...

function ```findFixtureFiles``` consists
* glob with absolute
* filter with micromatch

```js
export async function findFixtureFiles(
  args: Args
): Promise<Array<FixtureFile>> {
  const { fileMatch = defaultFileMatch, cwd = process.cwd() } = args || {};

  const allPaths = await globAsync('**/*', {
    cwd,
    absolute: true,
    ignore: '**/node_modules/**'
  });
  const fixturePaths = micromatch(allPaths, fileMatch);
//...
```

The problem may be solved with
* glob with no absolute
* micromatch with [dot option](https://www.npmjs.com/package/micromatch#optionsdot)

skidding:
@maruware thanks for bringing this up, I wasn't aware of it!

> micromatch with dot option

Sounds line an easy fix.

>  glob with no absolute

Not sure I understand how this would work. Do you have a favorite solution out of the two?

xavcz:
Solved in #618, will be released under `react-cosmos@4` asap 👍

PR #618, opened 2018-03-13
fix: find fixture files when cwd includes dotfile #610

xavcz:
Adding the micromatch `dot` option as suggested in #610 does the job! 👍 Great catch!

It has the same behavior as [minimatch `dot` option](https://github.com/isaacs/minimatch#dot):

> Note that by default, a/**/b will **not** match a/.d/b, unless dot is set.


skidding:
This is great! Just one mention: I see this more of a `BREAKING CHANGE` than `fix`, because for new (possibly unwanted) fixtures might pop up for existing users. I would've merged myself but wanted to see if you agree on this first and if so, set the semver prefix accordingly.

PS. I think the same is valid for f480fd863e10ef4028dbdc09da397242fd0edaaa, because existing users might end up having StateProxy configured twice for test environemnts where they already set it up to compensate for the previous behavior. So I say we release a silent `v4` after these PRs to stay true to semver.

xavcz:
Ok, makes sense!

Should we try to include the Apollo PR into that too? The new Apollo proxy introducing a breaking change compared to the initial Apollo proxy implementation

skidding:
> Should we try to include the Apollo PR into that too?

Yes!

PR #621, opened 2018-03-16
Simplify local Apollo server

skidding:


skidding:
@xavcz good idea putting the server locally. I went ahead and simplified it a bit as having a detached package.json/yarn.lock seemed error-prone.

Issue #626, opened 2018-03-20
cosmos is hardcoding node_env to 'development'


scottdickerson:
### What's up?
When starting cosmos from the command line it's hardcoding the node_env to 'development'.  https://github.com/react-cosmos/react-cosmos/blob/63e6c0574ceb12e13a9882e7f00fa5cf4d13262a/packages/react-cosmos/bin/cosmos.js#L5  This is fine by default, but I need to override it to pass a specific node_env so that I can instrument my classes for code coverage.

### Mkay, tell me more...

Allow a node env to be passed and if none passed default to 'development'.

skidding:
Yes! This should be fixed.

https://github.com/react-cosmos/react-cosmos/blob/0aea3ff15cb0321c0ffbd319168b2d564ab07f75/packages/react-cosmos/bin/cosmos.js#L4-L5

and 

https://github.com/react-cosmos/react-cosmos/blob/0aea3ff15cb0321c0ffbd319168b2d564ab07f75/packages/react-cosmos/bin/cosmos-export.js#L4-L5

Should be

```diff
-process.env.BABEL_ENV = 'development';
+process.env.BABEL_ENV = process.env.BABEL_ENV || 'development';
-process.env.NODE_ENV = 'development';
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
```

and 

```diff
-process.env.BABEL_ENV = 'production';
+process.env.BABEL_ENV = process.env.BABEL_ENV || 'production';
-process.env.NODE_ENV = 'production';
+process.env.NODE_ENV = process.env.NODE_ENV || 'production';
```

@scottdickerson Would you be willing to submit a patch for this? I'd be happy to review and release it.

scottdickerson:
absolutely @skidding PR https://github.com/react-cosmos/react-cosmos/pull/630 put in to fix this, thanks!

skidding:
Thanks @scottdickerson! 

Published under `react-cosmos@4.0.0-beta.4`, check it out. 

cc @NiGhTTraX I think you needed this as well.

PR #630, opened 2018-03-22
fix(bin/cosmos*.js): check the local process.env variables

scottdickerson:
Need to check whether downstream consumers have passed process.env to our cosmos and cosmos-export playground scripts before hardcoding to development or production respectively. to fix issue #626


xavcz:
Yay, thanks Scott!

PR #632, opened 2018-03-28
Allow parsing css from node_modules by default

### LINKED ISSUE #631: Allow loading 3rd party CSS from a Node package
### What's up?

Let's say you want to use CSS from a third party Node package, like [Semantic UI](https://react.semantic-ui.com), [Material UI](https://github.com/mui-org/material-ui) or [Blueprint](https://blueprintcss.io/).

To have these global styles available for each fixture, we can use `globalImports` in the Cosmos config:

```js
module.exports = {
	// ...
	globalImports: ['semantic-ui/semantic.min.css'] // or './node_modules/semantic-ui-css/semantic.min.css'
};
```

This import refers to a file inside the `node_modules` directory, that will be loaded by Webpack at some point.

The thing is that when Webpack tries to load this file, we get the following error:

```sh
ERROR in ./node_modules/semantic-ui-css/semantic.min.css
Module parse failed: Unexpected character '@' (11:0)
You may need an appropriate loader to handle this file type.
|  *
|  */
| @
```

Neither `css-loader` (here) or `style-loader` are used by Webpack to load this css file.

I've hit this issue on Friday, and @cfnelson also hit it hard with Material this weekend (and maybe also @cal0610 in https://github.com/react-cosmos/react-cosmos/issues/579#issuecomment-375195649 ?) 

### Mkay, tell me more...

Using the default webpack config provided by Cosmos, this is actually what is intended: [we `exclude` the `node_modules` directory when using `css+style-loader` (code)](https://github.com/react-cosmos/react-cosmos/blob/58bcd54e28a3e151813b9ef49e120a88bb2fbc2f/packages/react-cosmos/src/server/default-webpack-config.js#L30).

Is this line set to avoid parsing the 📦 `node_modules` rabbit hole?

My workaround so far has been to recreate my own basic config, without this line 🗡 

It has been a pretty frustrating experience when hitting this bug, should we allow loading CSS from `node_modules` by default?

xavcz:
> Fixes #631 

## Problem
One uses a 3rd party library such as Material UI or Semantic UI in their project, and want Cosmos to consume the same CSS.

They don't have a custom webpack config.

Cosmos fails, unable to parse the CSS imported from `node_modules`

## Proposed solution

Two loaders entries (with two different tests) to bring the most clarity to both users and maintainers about what's going on with the default webpack config:
- allow importing personal CSS
- allow importing 3rd party CSS

### DIFF
--- a/packages/react-cosmos/src/server/__tests__/default-webpack-config/style-css-loader.js
+++ b/packages/react-cosmos/src/server/__tests__/default-webpack-config/style-css-loader.js
@@ -18,11 +18,21 @@ jest.mock('import-from', () => ({
   }
 }));
 
-it('includes style-loader + css-loader', () => {
+it('parses personal css with style-loader + css-loader', () => {
   const config = getDefaultWebpackConfig('/foo/path');
   expect(config.module.rules).toContainEqual({
     test: /\.css$/,
     loader: '/style/path!/css/path',
     exclude: /node_modules/
+  });
+});
+
+it('parses 3rd party css with style-loader + css-loader', () => {
+  const config = getDefaultWebpackConfig('/foo/path');
+
+  expect(config.module.rules).toContainEqual({
+    test: /\.css$/,
+    loader: '/style/path!/css/path',
+    include: /node_modules/
   });
 });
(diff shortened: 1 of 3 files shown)


skidding:
Cheers, mate!

Just published `4.0.0-beta.5` with this. Try it out.

PR #637, opened 2018-04-02
Improve monorepo dev experience

### LINKED ISSUE #625: Improve monorepo dev experience
### What's up?

An abstract from the more detailed [Spectrum conversation](https://spectrum.chat/cosmos?thread=3c28f7de-1e90-43ef-a9ff-99791b073a66):

> The Cosmos repo is now managed via Yarn and Lerna, which for the most part (package linking and publishing, respectively) work great. But there are a few shortcomings that I keep bumping into and I'd love to see fixed:
>
> - Flow types are lost between packages
> - Sharing Flow types is hacky
> - `PKG/lib/*` cross-package imports are awkward
> - It's easy to make mistakes and run tests on the latest source of package A, which imports an old compiled output of package B
> - Test coverage is not properly calculated

### Mkay, tell me more...

@ovidiubute's responses are a great starting place:

> Your problems stem from the fact that you are depending on the production artifacts of packages, rather than the actual source code. A monorepo is awesome at sharing code, not artifacts, so you need a clear separation between these two concepts. Afaik you can export everything you need from each package and Yarn will pick everything up and resolve the imports. Now, what I think you need is a way to publish production level artifacts from each package which you will NOT use in the monorepo at all. This can probably be done using custom npm publish rules. The drawback is, of course, that you may miss something between the cracks because you're not using your own production code.

> I think the Jest monorepo is a good starting point. They store types root level in a types dir, so they're accessible to every single package (located in the packages directory). For each package, they export everything needed for all other packages, and then for production, they have a main field in package.json that points to build/index.js.

> Also, in all packages, everything in src/ is ignored via .npmignore, only what's in build gets published.

> Using yarn workspaces is the key feature as it will parse all package dependencies and hoist them top level if they versions match (which they should, since it's a monorepo). One extra awesome feature is using Jest workspaces, which means you can work in a package, run all unit-tests from the root repo level, and Jest will figure out all of the packages that depend on what you've changed (in watch mode), and only run the affected unit-tests.

I'll take a closer look at the [Jest monorepo](https://github.com/facebook/jest), run some tests on the Cosmos monorepo and report back.

I'm thinking of including this in the upcoming v4 release.

skidding:
Fixes #625

* [x] Move types at root level
* [x] Exporting all public APIs at root package level (`PKG/*.js`)
  * [x] shared
  * [x] voyager2
  * [x] Set `package.json:main` to `index.js` on all packages (which have a default entry point)
* [x] Enable Flow types across packages
* [x] Create script for pointing package entry points to `src` or `dist`
  * [x] Rename `lib` to `dist`
  * [x] Update `files` in every package.json 
* [x] Prettify build console output
  * [x] Prettify `link-deps` as well
* [x] Figure out publish process
* [ ] ~~Document `createFixture` Flow API~~ Do it later, after using it in dependent projects
  * [x] ~~Try: Type fixture.state in createFixture~~ [React.ComponentType](https://flow.org/en/docs/react/types/#toc-react-componenttype) only accepts props
* [x] Use `PropsProxy` in all proxies from monorepo
  * [x] Document `PropsProxy` type
  * [x] Remove `prop-types` deps in proxy packages
* [x] Convert default exports to named exports
  * [x] Add to CONTRIBUTING guide
* [x] Convert arrow function exports to function declarations
  * [x] Add to CONTRIBUTING guide

Considerations:
* Don't add `react-cosmos-flow` deps to avoid putting Flow types in users' node_modules. Packages should (?) access react-cosmos-flow from root-level node_modules.
* How to prevent commiting `dist` paths

Unresolved:
* [x] ~~How to preserve types when calling an API from another package?~~ By pointing package entry points to `src`
* [x] ~~Reconcile the fact that the repo doesn't require all but `react-cosmos-playground` to be build for examples to work~~ `react-cosmos-playground` and all server packages need to be built for examples to work
  * [x] ~~How to run `yarn build` after `yarn install` but not after `yarn add`? `postinstall` hook is ran after both and this because annoying. Most likely Cosmos dev don't want to wait 30s after installing a new dependency~~. Removed build step from postinstall hook
    * [x] Add instructions to run `yarn build` after `yarn install` in CONTRIBUTING guide


skidding:
OK, **ready for review.** I'm excited about this PR because it will make Cosmos development more enjoyable!

@ovidiubute @xavcz @tkjone if you want to take a look and help, here are some pointers.

A lot of it is chores related to adding Flow types, converting default exports to named exports, etc.

But the main part, which addresses all points from to the [original Spectrum chat](https://spectrum.chat/thread/3c28f7de-1e90-43ef-a9ff-99791b073a66), is best explained in the rewritten CONTRIBUTING sections:

1. [Compiled vs source code](https://github.com/react-cosmos/react-cosmos/blob/b95dab3744aee2153585fcc2df5f5268d7684c31/CONTRIBUTING.md#compiled-vs-source-code)

  image

2. [Small section about Flow](https://github.com/react-cosmos/react-cosmos/blob/37b5338acb3f857581c6664323cff7ff04d80dfc/CONTRIBUTING.md#flow)

  image

On top of that, I went ahead and made the script output more sexy. Namely, check out the `build` script:

cosmos-build
(Thanks @ovidiubute for creating the initial script, which served the repo well for a long time)

If you take a look at these you should be able to let me know if the new structure is clear and makes sense to you.

Appreciate any feedback! ✌️

ovidiubute:
Sounds good to me, awesome work!

xavcraft:
Yeah man, that looks great!

Especially excited by the Flow types, I'm going to see how I can develop in a better way with that 😄

skidding:
Thanks, guys!

PR #644, opened 2018-04-11
feat: Improve troubleshooting

skidding:


skidding:
We have this nice error message that we show when the user webpack build fails to generate an index.html page. Which is usually because `html-webpack-plugin` is missing. But other times it's something like an error in the webpack config, and we used to show the same screen, further confusing users and overall looking silly as a tool.

image

The first step in communicating error messages it to make sure we never tell the user to install a package when they have it! So I added another default screen to show when the `/loader/index.html` isn't being served, but the user has `html-webpack-plugin` installed. We know this buy embedding some information related to user dependencies in the playground options (aka `PlaygroundOpts`). In the future we can use these meta data to respond to other situations like when user doesn't have Babel loader, etc.

image

@xavczen if you have some time to take a look and tell me what you think I'd be very grateful! 🙏

PR #650, opened 2018-04-15
Document Flow helper

skidding:


skidding:
@andreypopp can you please share your thoughts on this? It mainly uses the `createFixture` function you [proposed](https://github.com/react-cosmos/react-cosmos/issues/560). I tried to present its shortcomings as well.

Also, do you still use `createFixture` in your projects?

cc @xavczen @maxsalven

PR #656, opened 2018-04-24
Add support for IE 11

cbranch101:
In `react-cosmos-playground` I had to add a polyfill for `fetch` and in `react-cosmos` I added a polyfill for `Object.values` and `Promise`

This let's everything work in IE 11


skidding:
Thank you @cbranch101!

Published under `4.0.0-rc.2`

PR #674, opened 2018-05-11
feat: Serve static files from root path by default

skidding:


skidding:
Remaining tasks:
- [ ] Construct error overlay filename with config.publicUrl
- [ ] Update webpack contributing guide section
- [ ] Update export readme section to mention using a static server to serve to export
- [ ] Heavy testing

PR #678, opened 2018-05-15
fix: xhr proxy should resetup mocks in hot module mode

SimeonC:
Currently webpack hot module handler removes the mocks but doesn’t re-add them, this fixes that.

The issue I was seeing was that when a module was hot loaded the xhrMocks were removed, meaning that subsequent requests went out to the internet not the xhrMock.

### DIFF
--- a/packages/react-cosmos-xhr-proxy/src/index.js
+++ b/packages/react-cosmos-xhr-proxy/src/index.js
@@ -24,6 +24,8 @@ export function createXhrProxy({ fixtureKey = 'xhr' }: Options = {}) {
         module.hot.status(status => {
           if (status === 'check') {
             xhrMock.teardown();
+          } else if (status === 'apply') {
+            this.mock();
           }
         });
       }


skidding:
Hi @SimeonC,

Can you provide a failing example? I tried this in the Axios example it seems like the proxy instance is created from scratch whenever the component is hot reloaded, thus attaching the mocks again.

xhr-mock

SimeonC:
Yea, I couldn't replicate it in that example either (it definitely happens in my repo though).
I'll have another poke around and see if I can replicate it when I have the time, I have a feeling this may be due to some other proxies I'm using or possibly MobX which prevents the re-render. (But this did get things working on my machine).
(Actually thinking about it, it quite probably is MobX or something else playing with `componentShouldUpdate` but I'll confirm that).

SimeonC:
Ahhh, I found it. There seems to have been 2 issues. One of which was fixed by the upgrade, the other is considerably weirder - and I can actually reproduce it. So it looks like the webpack HMR was watching my `<PROJECT_ROOT>/.gtm` (git time manager) plugin file and triggering updates. This probably is pointing to some issue in my webpack setup that should be more discreet in watching for changes (I've also noticed `<PROJECT_ROOT>/node_modules/.cache/*` files triggering this same behaviour).

#### Terminal
```sh
[Cosmos] webpack build invalidated by .gtm/1528880846.event
webpack building...
webpack built 1266c0e27527f30f7adb in 452ms
```

#### Browser Console with following code in XHR Proxy:
```jsx
export function createXhrProxy({ fixtureKey = 'xhr' }: Options = {}) {
  class XhrProxy extends Component<ProxyProps> {
    constructor(props: ProxyProps) {
      super(props);

      // webpack shiz
      // $FlowFixMe
      if (module.hot) {
        module.hot.status(status => {
          if (status === 'check') {
            console.log('[XHR] teardown on check');
            xhrMock.teardown();
          }
        });
      }

      console.log('[XHR] Constructor Mock');
      this.mock();
    }

    componentWillUnmount() {
      if (xhrMock.__prevProxy === this) {
        console.log('[XHR] unmock on unmount');
        this.unmock();
      }
    }

    mock() {
      // Clear mocks from a previous FetchProxy instance
      // Warning: A page can only have one FetchProxy instance at the same time
      console.log('[XHR] unmock in mock()');
      this.unmock();
      console.log('[XHR] Build Mocks');
      const mocks = this.props.fixture[fixtureKey];
      if (mocks) {
        xhrMock.setup();

        mocks.forEach(options => {
          const { url, method, response } = { ...mockDefaults, ...options };
          xhrMock[method.toLowerCase()](url, response);
        });
        xhrMock.__prevProxy = this;
      }
    }

    unmock() {
      xhrMock.teardown();
    }

    render() { ... }
  }

  return XhrProxy;
}
```
builder builder react cosmos 2018-06-13 18-10-44

SimeonC:
I've reopened this MR, but I'm not sure in looking at it if this would be the best fix. Even in this case we would end up doing a lot more teardowns and setups than really necessary.

SimeonC:
😃 Thanks, that was it all along. I'll close this again - that's definitely the way to fix it.
Thanks for your patience!

skidding:
🍻

Maybe in the future we find a way to have a better default for `watchDirs`. It'd be great if we could tell webpack to ignore a list of known files or file types from the watch list. I don't remember it being an option, though.

SimeonC:
There is a way of doing it - Create React App does it with a custom regex here https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/config/webpackDevServer.config.js

A good start might be `.something` files and folders. Normally they’re just config/ide files in my projects which I don’t really want being watched and would have protected against my problem.

 (I use my own fork of CRA so that was what I was starting to fidget with when I got your message about watchDirs).

skidding:
@SimeonC that sounds like a good idea. Would you be willing to draft a new issue for this so we don't forget it?

PR #686, opened 2018-05-22
docs: semicolon in object literal typo

mpachin:
Fixed small typo with semicolon in module.exports object literal 😊


skidding:
Thanks @mpachin!

PR #689, opened 2018-05-23
feat: Add field in config for find-fixture-files (#687)

### LINKED ISSUE #687: Add field in config for find-fixture-files method
### What's up?

The project under discussion have such directory tree:

```
project
│
└─── node_modules
│   │
│   └─── ...
│   
└─── src
│   │
│   └─── node_modules
│   │   │
│   │   └─── @subproject1
│   │   │   │
│   │   │   └─── ...
│   │   │
│   │   └─── @subproject2
│   │   │   │
│   │   │   └─── ...
...
```
`@subproject`'s directories hold components which I want react-cosmos to keep track of.

Right now `cosmos.config.js` haven't field which could tell react-cosmos to not ignore subprojects in `./src/node_modules`, but keep ignore `./node_modules`, and that means I cant make react-cosmos display fixtures placed in there.

### Mkay, tell me more...

I'd like to add an optional field in `cosmos.config.js` which will [replace default value](https://github.com/react-cosmos/react-cosmos/blob/7e11e8a9f4c99c74c6f040358e9e99f9fc55643e/packages/react-cosmos-voyager2/src/server/find-fixture-files.js#L37) if specified.

mpachin:
Fixes #687

### DIFF
--- a/README.md
+++ b/README.md
@@ -931,11 +931,17 @@ module.exports = {
 
 #### Custom fixture paths
 
-The `fileMatch` and `exclude` options are used to detect fixture files. The default fileMatch value is meant to accommodate most needs out of the box:
+The `fileMatch`, `fileMatchIgnore` and `exclude` options are used to detect fixture files. The default `fileMatch` value is meant to accommodate most needs out of the box:
 
 ```
 '**/__fixture?(s)__/**/*.{js,jsx,ts,tsx}',
 '**/?(*.)fixture?(s).{js,jsx,ts,tsx}'
+```
+
+The default `fileMatchIgnore` value is meant to [ignore node_modules folder](docs/file-match-ignore.md):
+
+```
+'**/node_modules/**'
 ```
 
 > Note: Set the `rootPath` to a dir parent to all fixture files when using a [custom config path](#custom-config-path)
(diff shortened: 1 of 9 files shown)


skidding:
This is great!

I only have two mentions:
1. I appreciate the detailed README, but I'm afraid your scenario is too specific for the main README (it's already huge...). If you think it's valuable, you can move everything from "If your projects directory tree looks like this:" to a separate md file in the [docs dir](https://github.com/react-cosmos/react-cosmos/tree/da9c355a2edca480603873b3ecef23bdabfa7623/docs) and link to it. Another option is to link to #687 as an example of where `fileMatchIgnore` can be useful.
2. This is a pet peeve of mine and I apologize if it comes off annoying. I noticed you've always kept `fileMatch`, `exclude` and `fileMatchIgnore` in the same order, which is cool. But can you change the order as follows please? `fileMatch`, `fileMatchIgnore` and `exclude`. I find this makes more sense since `fileMatchIgnore` comes in effect before `exclude` and it looks better visually for `fileMatch*` options to be grouped.

Thanks!

mpachin:
Corrected 😊

skidding:
@mpachin Please test your new option on `4.4.0`! 🎉

mpachin:
@skidding just checked it in my project, everything works! :smiley::sparkles:

Issue #703, opened 2018-06-07
You should not use <Link> outside a <Router>


teimurjan:
I faced a problem with creating a fixture for a component which has <Link> but without wrapping with `withRouter`. I have such a component:
```js
const AuthForm = ({ intl }: Props) => (
  <div className="auth-form">
    <Link to="/auth/login">
      <Button className="login-button mt-5 py-2" size="lg" color="primary" block>
        {intl.formatMessage({ id: 'authForm.button.login' })}
      </Button>
    </Link>

    <Link to="/auth/register">
      <button className="btn btn-lg btn-white-outline sign-up-button mt-4 mb-5 btn-block py-2">
        {intl.formatMessage({ id: 'authForm.button.signUp' })}
      </button>
    </Link>
  </div>
);

export default AuthForm;
```

My cosmos.proxies.js looks like:
```js
// @flow
import createRouterProxy from 'react-cosmos-router-proxy';

export default [createRouterProxy()];
```
And my fixture:
```js
export default [
  {
    name: 'Initial state',
    component: AuthForm,
    props: {
      intl: stubIntl,
    },
    url: '/auth',
  }
]
```

I also tried:
```js
// @flow
import createContextProxy from 'react-cosmos-context-proxy';
import PropTypes from 'prop-types';

const ContextProxy = createContextProxy({
  childContextTypes: {
    router: PropTypes.object.isRequired,
  },
});

export default [ContextProxy];
```
When the fixture was:
```js
export default [
  {
    name: 'Initial state',
    component: AuthForm,
    props: {
      intl: stubIntl,
    },
    context: {
      router: {},
    },
  },
];

```
None of these methods are working. Everytime I refresh the page I see `You should not use <Link> outside a <Router>`. What do I do wrong?

teimurjan:
I found out the problem. As I've changed `rootPath` in configs, file with proxies should be in the root folder, not near the config file.

mpachin:
Faced the same problem since my projects config files placed outside of rootPath.
[Here we can see](https://github.com/react-cosmos/react-cosmos/blob/e70ef0fe9a0cd982b8245ab1ebaceaa0647bbdd3/packages/react-cosmos-config/src/index.js#L117) that proxies taken from `rootPath`/`proxiesPath`.
In case similar with mine:
```
projectFolder
│
└─── configFiles
│   │
│   └─── react-cosmos
│       │
│       └─── cosmos.config.js
│       │
[...]

PR #704, opened 2018-06-07
docs: proxiesPath outside of rootPath

mpachin:
Since proxiesPath is always relative to rootPath this may be beneficial to [mention that in docs](https://github.com/react-cosmos/react-cosmos/issues/703).

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ yarn add --dev react-cosmos-fetch-proxy react-cosmos-redux-proxy react-cosmos-ro
 
 Then create `cosmos.proxies.js` (in your project's root directory or next to cosmos.config.js) and export a list of proxies in the order they should load–from outermost to innermost.
 
-> `proxies.cosmos.js` requires compilation so you may need to place it next to your source files (eg. if the `src` dir is whitelisted in babel-loader). Use `proxiesPath` option to customize its location.
+> `proxies.cosmos.js` requires compilation so you may need to place it next to your source files (eg. if the `src` dir is whitelisted in babel-loader). Use [proxiesPath](docs/proxies-outside-source.md) option to customize its location.
 
 Here's an example where we mock the Fetch API and add Redux and React Router providers:
 
--- a/docs/proxies-outside-source.md
+++ b/docs/proxies-outside-source.md
@@ -1,0 +1,25 @@
+If your projects directory tree look similar to this:
+```
+projectFolder
+│
+└─── configFiles
+│   │
+│   └─── react-cosmos
+│       │
+│       └─── cosmos.config.js
+│       │
+│       └─── cosmos.proxies.js
+│   
+└─── src
+│   │
+│   └─── ...
+...
+```
+And you want to hold your `cosmos.proxies.js` outside of root directory you need to define `proxiesPath` in `cosmos.config.js` relatively to `rootPath`.
+```js
+module.exports = {
+    rootPath: '../../src/',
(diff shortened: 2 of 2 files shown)


skidding:
Hey @mpachin,

> Since proxiesPath is always relative to rootPath

All config paths are relative to rootPath. This is already mentioned in two places:

image
image

> so you can skip placing `cosmos.proxies.js` in src directory and instead keep it near config.

This can be misleading. The reason why we ever suggest putting `cosmos.proxies.js` in `src` is Create React App (and other similar setups) where only the src dir is compiled by Babel. In those cases, which are very popular, putting the cosmos.proxies file in a config dir fails.

And I wouldn't want to overwhelm users with too much specification around where to place files. Again, Cosmos docs are already very stuffy.

That said, #703 is proof that something can be improved. I propose this: When no cosmos.proxies file is found, we check if there is one relative to the config file's dir. If true, we warn the user that proxiesPath should be relative to rootPath and give them an example of what the value should be. Wdyt?

mpachin:
@skidding first of all thank you for your quick and thoughtful responses and your work on such great tool :heart: 

I think this is wonderful idea :+1:  This definitely will made dev experience more joyful.

Maybe sequence of checks should look like this?
1. Check for errors first - if developer don't use proxies at all and no errors were found we shouldn't explicitly search for `cosmos.proxies` file
2. If errors were caught - we may check if `cosmos.proxies` file placed near `cosmos.config` and display a message for user with content you described
3. If `cosmos.proxies` weren't found near `cosmos.config` I feel that it would be great to display a thorough message in which:
    * mentioned appropriate default dir to place `cosmos.proxies`
    * mentioned role of `rootPath` and `proxiesPath` and that second one is relative to the first

skidding:
> @skidding first of all thank you for your quick and thoughtful responses and your work on such great tool ❤️

My pleasure! And thanks for your contributions!

> Maybe sequence of checks should look like this?

I like your ideas. Here's a few more insider info relevant to this flow.

`proxiesPath` is always defined. Even if you don't specify it, it has a default. [See here](https://github.com/react-cosmos/react-cosmos/blob/58baebac3004cecb24a1fd334ac6514fe1a28c97/packages/react-cosmos-config/src/index.js#L28). This is useful when someone creates `cosmos.proxies` file but doesn't create `cosmos.config`, which is optional.

Because of the above, we currently don't warn if there's no file found at `proxiesPath`. We just move along.

To achieve what you're suggesting, I think we need to expose a new `react-cosmos-config` method, which returns the raw user config. Right now [getCosmosConfig](https://github.com/react-cosmos/react-cosmos/blob/58baebac3004cecb24a1fd334ac6514fe1a28c97/packages/react-cosmos-config/src/index.js#L40-L62) returns the user config merged with the default config, with all paths resolved to `rootPath`.

I'm thinking we could call `getRawCosmosConfig`, which would be [this part](https://github.com/react-cosmos/react-cosmos/blob/58baebac3004cecb24a1fd334ac6514fe1a28c97/packages/react-cosmos-config/src/index.js#L41-L55) except it would return an empty object when the config is missing. Then we would be able to _know_ that the user defined their proxiesPath to an invalid path.

Reading the raw config would also give us the raw proxiesPath value, which would allow us to construct `resolve(dirname(getUserConfigPath()), rawConfig.proxiesPath)` to determine if a proxies file has been created near the config file, like in your case when you stumbled into this issue.

mpachin:
@skidding Thank you for pointing this out, I'll try to use this info and look for solution in nearest time.
Closing this pull request for now.

PR #712, opened 2018-06-12
chore: Extract non-webpack core from react-cosmos/server

skidding:


skidding:
TODO:
- [ ] Why is CodeMirror syntax highlighting and code folding not working anymore?

PR #737, opened 2018-07-02
fix: Empty Apollo's cache when switching mocked fixture

xavczen:
👋 hey there!

What's up?
---
When using an `ApolloProxy` defined with `clientOptions` and browsing mocked fixture, the cache is still persisted! 

screen recording 2018-07-01 at 10 10 pm

Solution provided by this PR
---
If we are on a mocked fixture (`isMockedFixture`), be sure to use a `new` InMemoryCache` instance :smile:

screen recording 2018-07-01 at 10 09 pm

What could be better?
---
Changing (again) the API of the ApolloProxy to set the client options thanks to a function, so we are sure nothing is persisted in the module `cosmos.proxies.js` :thinking_face:

### DIFF
--- a/packages/react-cosmos-apollo-proxy/src/index.js
+++ b/packages/react-cosmos-apollo-proxy/src/index.js
@@ -50,9 +50,11 @@ Read more at: https://github.com/react-cosmos/react-cosmos#react-apollo-graphql.
       };
 
       if (isMockedFixture) {
+        // ensure that the cache is not persisted between mocked fixtures
+        options.cache = cache;
         options.link = createFixtureLink({
           apolloFixture,
-          cache,
+          cache: options.cache,
           fixture: this.props.fixture
         });
       }


skidding:
Great catch!

Just to make sure I understand. Is the underlying issue the fact that we mutate the user's `clientOptions` inside the proxy?

Would this also work?

```js
if (isMockedFixture) {
  options = {
    ...options,
    link: createFixtureLink({
      apolloFixture,
      cache,
      fixture: this.props.fixture
    });
  }
}
```

xavczen:
> Is the underlying issue the fact that we mutate the user's clientOptions inside the proxy?

Yep 😞 

Thanks for the recommendation of cloning, but even when cloning with different approaches (not just `Object.assign`), the cache is still mutated 🤐

skidding:
Interesting.!

So do we need to pass the `cache` instance to both `ApolloClient.options` as well as `ApolloClient.options.link`?

xavczen:
> `ApolloClient.options`
Needs a `cache` to work 👍 

> `ApolloClient.options.link`
Only the `FixtureLink` might need the cache to read data from it, for example when executing a mutation:
```js
export default {
  component: Author,
  props: {
    authorId: 123
  },
  apollo: {
    // mocked response for the mutation named UpvotePost
    UpvotePost: {
      resolveWith: ({ cache, variables, fixture }) => {
        // read data from the cache
        const data = cache.readQuery({
          query: QUERY,
          variables: { authorId: fixture.props.authorId }
        });

        const post = data.author.posts.find(
          post => post.id === variables.postId
        );

        return {
          upvotePost: {
            ...post,
            votes: post.votes + 10
          }
        };
      }
    }
  }
};
```

Issue #566, opened 2018-01-02
Providing a 'route' to a fixture doesn't provide the `match`, `location`, etc. props


rhysforyou:
### What's up?

When React Router v4 renders a `component` that was provided to a `Route`, it passes a few props along like `match`, `history`, and so on. However, when I provide a `route` for a fixture while using the `react-cosmos-router-proxy` package, these props aren't passed to my component.

### Mkay, tell me more...

Right now I'm getting around this by wrapping the component in `withRouter`, perhaps we could have React Cosmos do this itself when a `route` is provided?

skidding:
Interesting. It looks like you're trying to do 

```
<Route path="/user/:username" component={User}/>
```

when the Router Proxy does this

```
<Route path="/user/:username">
  <User />
</Route>
```

> Right now I'm getting around this by wrapping the component in withRouter, perhaps we could have React Cosmos do this itself when a route is provided?

It would work in your case (it seems like the components you're trying to load are probably root level components for each route, correct me if I'm wrong), but would be an extra wrapper for all the other cases when you have non-root level components already wrapped with `withRouter` throughout the app.

We could extend `react-cosmos-router-proxy` to respond to `fixture.routerDecorator: true` (feel free to propose other naming). @rpowelll what do you think?

rpowelll:
Yeah what I want is something like

```jsx
<Route path={fixture.route} component={fixture.component} />
```

But instead what seems to be happening is something like

```jsx
<Route path={fixture.route} render={() =>
  <CosmosWrapperStuff>
    <fixture.component />
  </CosmosWrapperStuff>
}/>
```

I think the suggested config is probably a good solution if we're not willing to change the current behavior.

skidding:
@rhysforyou you can use @maxsalven's feature from #745 starting with version `4.6.0-2`.

PR #745, opened 2018-07-06
Provide Router Props

maxsalven:
Set `provideRouterProps: true` on a fixture to pass in `location`, `history`, and `match` from the `Route`.

Fixes https://github.com/react-cosmos/react-cosmos/issues/566

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ export default {
 };
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -466,6 +466,18 @@ export default {
 };
 ```
 
+If your component needs props from the `Route` component (history, location, and match), you can enabled the `provideRouterProps` flag.
```
skidding: Small typo: `enable`

PR #760, opened 2018-07-20
feat: apollo-proxy support returning a promise

SimeonC:
resloveWith function should be able to handle returning a Promise.

This is useful for delaying the result to allow mocking of longer running requests.
In my specific case I wanted to mock all but one apollo graphql request so I could use this to manually make the graphql call for that one particular fixture.

I've added tests for this and some extra details to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,11 @@ Once configured, your components enhanced by `react-apollo` will behave as they 
 
 Mocking at the fixture level is done by specifying an `apollo` key in your fixture.
 
-The proxy will look for a `resolveWith` or a `failWith` key in order to return the appropriate mock value: this can be an object or a function returning an object.
+The proxy will look for a `resolveWith` or a `failWith` key in order to return the appropriate mock value. This mocked return can be one of;
+
+- an object
+- a function returning an object
+- a function that returns a Promise that either resolves or rejects with an object
 
 See examples below or check [the fixtures defined in the Apollo example](examples/apollo/components/__fixtures__/Author).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -534,7 +534,11 @@ Once configured, your components enhanced by `react-apollo` will behave as they
 
 Mocking at the fixture level is done by specifying an `apollo` key in your fixture.
 
-The proxy will look for a `resolveWith` or a `failWith` key in order to return the appropriate mock value: this can be an object or a function returning an object.
+The proxy will look for a `resolveWith` or a `failWith` key in order to return the appropriate mock value: this can be;
```
skidding: `s/;/:`

packages/react-cosmos-apollo-proxy/src/fixtureLink.js
```diff
@@ -11,19 +11,47 @@ export function createFixtureLink({ apolloFixture, cache, fixture }) {
...
-            typeof resolveWith === 'function'
-              ? resolveWith({ cache, variables, fixture })
-              : resolveWith.data
-                ? resolveWith.data
-                : resolveWith,
-          errors:
-            (typeof resolveWith !== 'function' && resolveWith.errors) ||
-            undefined
-        });
+        let result;
```
skidding: Any reason to hoist the `result` variable here?

SimeonC: Missed refactor - I'll change it. Along with trying to flatten the function down a bit more.

SimeonC:
Pushed an update with the review changes. 👍

skidding:
Thanks!

PR #762, opened 2018-07-20
Use a function for identifying a component in a fixture

sdgandhi:
I had to read through the source code and experiment to find this out, so exposing it in the readme would save others time!

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,24 @@ export default {
 };
 ```
 
+#### Component
+
+You may identify a component directly, or provide a function. This allows for quick wrapping of components. If you find yourself doing this often with the same component, it might be time to try [proxies](#proxies).
+
+```js
+export default {
+  component: props => (
+    <Well>
+      <SearchBox {...props} />
+    </Well>
+  ),
+  state: {
+    searchQuery: 'Who let the dogs out?'
+  }
+};
+```
+
+
 #### Init hook
 
 This is an advanced feature and should only be used when a desired state can't be reproduced via [proxies](#proxies).
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ export default {
 };
 ```
 
-#### Component
+#### Wrapper component
 
 You may identify a component directly, or provide a function. This allows for quick wrapping of components. If you find yourself doing this often with the same component, it might be time to try [proxies](#proxies).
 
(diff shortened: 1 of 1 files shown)


skidding:
Great idea passing down the props! Somehow I missed this pattern completely till now.

I suggest two changes:
- Replace `Component` with `Wrapper component`, because "Component" sounds more like the primary use case, where you pass a component class.
- Replace `fixture.state` from this example, because actually it won't work (although this will be supported in the next major version), with some example props, because they are correctly passed down to `SearchBox`.

sdgandhi:
@skidding addressed those changes.

However, I'm not sure why you say the previous `fixture.state` won't work since it's used in the example right above it.

skidding:
Thanks!

> I'm not sure why you say the previous fixture.state won't work since it's used in the example right above it

Because the state would try to be applied to the `props => (...)` wrapper component, which is a stateless component and by definition can't hold state. And also because I think in that example would imagine the state is being appled to the inner `<SearchBox ...>` component, which is not the case.

sdgandhi:
@skidding ahh you're right.

Is the upcoming fix for this going to give `(props, state) => (...)` so state can be give to the correct component with `...state`?

skidding:
> Is the upcoming fix for this going to give (props, state) => (...) so state can be give to the correct component with ...state

I wish it worked like this. But you can't inject state into a component using `...state`. Part of what makes Cosmos cool is the ability to inject state into a component, which is otherwise unusual in the React land. The current API, however, only allows injecting state into a root component class attached to a fixture (ie. `{ component: MyComponent ... }`). My plan is to use composition to be able to do stuff like this in the future:

```js
// some-fixture.js
export default (
  <Wrapper>
    <ComponentState searchQuery="Who let the docs out?">
      <SearchBox />
    </ComponentState>
  </Wrapper>
)
```

But this functionality is part of a bigger project I want to tackle: JSX fixtures – Being able to express Cosmos fixtures as regular JSX elements (like in the example above). It's the next big thing on my list after UI plugins, so stay tunned. See my [roadmap](https://github.com/react-cosmos/react-cosmos/blob/52a0771bce7e2576931582bfb00c4e00e581c947/TODO.md#roadmap-summer-of-cosmos) for an overview.

sdgandhi:
That's awesome, thanks for the info.

PR #768, opened 2018-07-26
fix: correctly detect if the component is forwarding ref

SimeonC:
This adds support for usage of React 16.3 [`forwardRef`](https://reactjs.org/docs/forwarding-refs.html).
This is normally important when using HOC as when using `forwardRef` `isComponentClass` returns false but we want to apply the ref anyway. Of course, if the WrappedComponent isn't a Component it would fail - but that's not our problem.

```jsx
export function withShops(WrappedComponent: *) {
  const withShopsComponent = forwardRef((props: any, ref: any) => (
    <Consumer>
      {({ merchantUserShops: shops }) => (
        <WrappedComponent {...props} shops={shops} ref={ref} />
      )}
    </Consumer>
  ));
  withShopsComponent.displayName = `WithShops(${getDisplayName(
    WrappedComponent
  )})`;
  return withShopsComponent;
}
```


skidding:
Hey! Is anything from the [react-is](https://www.npmjs.com/package/react-is) package that can help us detect this more elegantly?

SimeonC:
Cool, didn't know about that package. Looks like it would allow us to do it much cleaner - I'll give it a go.

SimeonC:
So I can use some of that package to help out, but the functions are designed for use in TestRender apparently so don't detect forwardRef components here: https://github.com/facebook/react/blob/bc1ea9cd96c81467c574b134349eafc303835d0e/packages/react-is/src/ReactIs.js#L29-L33

I've updated to make it a bit more consistent overall.

SimeonC:
I couldn't write any tests due to this open PR on enzyme: https://github.com/airbnb/enzyme/pull/1592

skidding:
Looking good!

Before merging:
- Please resolve conflict, I just merged a branch.
- Regarding tests, you may want to give it another try with the [latest fork of enzyme react adaptor used in master](https://github.com/react-cosmos/react-cosmos/blob/fb73be64433d44f8776ed6387c15d03d591d251c/package.json#L57). I started using the new Context API, which Enzyme also doesn't support yet, so I resorted to using that fork in the meantime. _Maybe_ it also works with refs? Wild shot, but I'm ready to merge this anyway.

SimeonC:
Cool, looks like the latest fork of enzyme adapter does work (at least locally). Added tests and resolved conflicts.

PR #778, opened 2018-08-01
Adds support for spread in fixtures

SimeonC:
Fixtures utilising ES6 spread would error out when trying to find the component and hence would not correctly appear in the same folder as others.

It bugged me that some of my fixtures would inexplicably go into a different folder, turns out this was why. Also the check for `prop.key.name === 'component'` would fail on a spread as `key` was undefined so it didn't matter if I had `component` defined after the spread, the spread would just break it all.

After matches what my directory structure is.
spread detection copy

### DIFF
--- a/packages/react-cosmos-voyager2/src/server/__tests__/esm/multi-spread/__fsmocks__/Bold.js
+++ b/packages/react-cosmos-voyager2/src/server/__tests__/esm/multi-spread/__fsmocks__/Bold.js
@@ -1,0 +1,9 @@
+// @flow
+
+import React, { Component } from 'react';
+
+export default class Bold extends Component<{ name: string }> {
+  render() {
+    return <strong>{this.props.name}</strong>;
+  }
+}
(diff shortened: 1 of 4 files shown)


skidding:
Hi @SimeonC,

Good idea!

1. Is there any reason why you also changed the `multi-vars/__fsmocks__/joined.fixture.js` file? I think it belongs to a separate test case so I'd leave it as before.
2. The build is failing due to Flow errors. Do you need any help? Note that the type annotations in the AST code is far from ideal as I already used `any` types and `$FlowFixMe` in one place, so feel free to make some concessions as long as the return type is not compromised. Also, let me know if you have any issue running Flow or any repo command locally.

SimeonC:
Updated with fixes. 1 was just me not checking properly, 2 was forgetting to actually run the flow checker, forgot to set it up in my IDE so no errors were showing up. Should be fixed now.

PR #800, opened 2018-08-15
Add CircleCI config

skidding:


skidding:
CircleCI seems 50% to 100% faster, but I'll leave both CIs run in parallel for a couple of PRs before removing Travis.

image

image

PR #803, opened 2018-08-16
React router snapshot fix (Keys)

JannesV:
The keylength for the memoryrouter was not set thus the router generated new keys every run and breaking snapshots tests in the process. This simple prop addition fixes said issue.

### DIFF
--- a/packages/react-cosmos-router-proxy/src/index.js
+++ b/packages/react-cosmos-router-proxy/src/index.js
@@ -33,7 +33,7 @@ export function createRouterProxy() {
     const location = buildLocation(url, locationState);
 
     return (
-      <MemoryRouter initialEntries={[location]}>
+      <MemoryRouter keyLength={0} initialEntries={[location]}>
         <LocationInterceptor
           onUrlChange={handleUrlChange}
           onLocationStateChange={handleLocationStateChange}


maxsalven:
Hi Jannes, I helped author some of the key based code. We already explicitly set the key to “mocked”, specifically to provide a stable snapshot. There is a test here:

https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-router-proxy/src/__tests__/stable-snapshots.js

Would you be able to provide a failing test to help better understand the problems with the current implementation? Thanks!

JannesV:
@maxsalven One of my test breaks for some reason on the key changing in the snapshot. As you said the first entry is mocked but when there are multiple entries only the last one changes. Why did you opt to set it to mocked instead of omitting it completely ?

```
                 "entries": Array [
                      Object {
                        "hash": "",
                        "key": "mocked",
                        "pathname": "/",
                        "search": "",
                        "state": undefined,
                      },
                      Object {
                        "hash": "",
                        "key": "95psso",
                        "pathname": "/1",
                        "search": "",
                        "state": undefined,
                      },
                    ],
```

skidding:
Hi @JannesV,

As I understand, setting `keyLength={0}` means that the [keys generated by `history`](https://github.com/ReactTraining/history/blob/c0d2ac7adcee6d35791a72db417baf0d57b63548/modules/createMemoryHistory.js#L30-L33) will be strings with 0 length, so `""`.

Can you please add a comment explaining the effect and reason for doing this before the return line? I first thought it's related to the length of the entries and others might be similarly confused. Thanks!

And just to make sure I understand: The first entry will always have the key set to `"mocked"`, whereas any other entries (generated by test actions) will have the key set to `""`. Right? Ideally we would clear the initial key as well for consistency, but that would be a breaking change for people with existing snapshots so I'm good with this.

PR #827, opened 2018-08-29
feat: Dark launch JSX fixtures

skidding:


skidding:
> Dark launch ? 😱 👺

Yeah, meaning it will be in _production_, but nothing will change for existing users. However, if you set `next: true` in your config a brand new (experimental) universe will open up to you!

I'll document how to use the new APIs once this is ready. 🤘

PR #849, opened 2018-09-13
fix(export): Log errors and fail command on errors

### LINKED ISSUE #834: Webpack errors should fail the export command
### What's wrong?

When running the regular dev command you will see Webpack errors like `Module not found: Error: Can't resolve......`, but the export command completely hides that and from the console output it looks like a successful build.

### Steps to reproduce

_(I'm running this on latest version, 4.6.4)_

1. Introduce a bug where a file is importing another file that doesn't exist.
2. `cosmos-export`
3. The build should now fail and display the errors, but instead the output looks the same as a successful build.

### Additional context

I looked into it a bit and it looks like we might want to check the `stats` object here: https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos/src/server/web/export.js#L31 (webpack docs: https://webpack.js.org/api/node/#stats-haserrors-). Unfortunately I ran into build issues with Babel when trying to get the `react-cosmos` repo running locally, and I don't have time currently to dig deeper to fix it. But maybe someone else can pick up from here :)

koggdal:
Only errors coming from the error object from the `run` method were logged before. But there might be other kinds of errors coming from webpack and these need to be handled as well.

Fixes #834 

Example output:

screen shot 2018-09-13 at 17 46 14

### DIFF
--- a/packages/react-cosmos/src/server/web/export.js
+++ b/packages/react-cosmos/src/server/web/export.js
@@ -31,6 +31,11 @@ const runWebpackCompiler = (webpack, config) =>
     compiler.run((err, stats) => {
       if (err) {
         reject(err);
+      } else if (stats.hasErrors()) {
+        const error = new Error('Errors occurred');
+        //$FlowFixMe
+        error.webpackErrors = stats.toJson().errors;
+        reject(error);
       } else {
         resolve(stats);
       }
@@ -87,8 +92,15 @@ export async function generateExport() {
         console.log(outputPath);
       },
       err => {
-        console.error('[Cosmos] Export Failed! See error below:');
-        console.error(err);
+        console.error('[Cosmos] Export Failed! See errors below:\n');
+        if (err.webpackErrors) {
+          err.webpackErrors.forEach(error => {
+            console.error(`${error}\n`);
+          });
+        } else {
+          console.error(err);
+        }
+        process.exit(1);
       }
     );
 }
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,7 +6,3 @@ coverage:
       default:
         target: 90%
         threshold: 5%
-    patch:
-      default:
-        target: 80%
-        threshold: 25%


koggdal:
Let's see if it builds now! The file isn't tested very well, so 80% code coverage requirement forced me to add a test for some existing things as well.

packages/react-cosmos/src/server/web/export.js
```diff
@@ -87,8 +92,17 @@ export async function generateExport() {
...
       err => {
-        console.error('[Cosmos] Export Failed! See error below:');
-        console.error(err);
+        console.error('[Cosmos] Export Failed! See errors below:');
+        if (err.webpackErrors) {
+          err.webpackErrors.forEach(error => {
+            console.error('');
+            console.error(error);
+          });
+          console.error('');
```
skidding: Can we remove this? I don't think it's needed and isn't consistent with the `else` branch anyway.

packages/react-cosmos/src/server/web/export.js
```diff
@@ -87,8 +92,17 @@ export async function generateExport() {
         console.log(outputPath);
       },
       err => {
-        console.error('[Cosmos] Export Failed! See error below:');
-        console.error(err);
+        console.error('[Cosmos] Export Failed! See errors below:');
+        if (err.webpackErrors) {
+          err.webpackErrors.forEach(error => {
+            console.error('');
+            console.error(error);
```
skidding: Maybe
```
console.error(`\n${error}`);
```
is cleaner?

skidding:
Hi @koggdal,

Thanks for fixing this!

> The file isn't tested very well, so 80% code coverage requirement forced me to add a test for some existing things as well.

Sorry about that hard requirement regarding coverage. I don't find it useful all the time so I think I'll disable it after this PR.  I find mocking `console.error` and `process.exit` overkill to be honest, so if you don't mind, it would be easier if you reverted the test changes and left the initial fix in this PR. Which, aside from stylistic comments, is good to merge.

koggdal:
@skidding Okay sure! I find it a bit sad to not test important things though.

skidding:
> @skidding Okay sure! I find it a bit sad to not test important things though.

Thanks! What do you feel is important and not tested? I'm interested in broadening my testing philosophy and maybe we can add some tests later.

koggdal:
> Thanks for your contribution! You can start using your code under v4.7.0-10

Thanks!

> Btw, I prefered reverting that commit because I'm busy this week so it was easier for me to review the first part and merge it quicker. Sorry if I came off as dismissive!

No worries, totally understand your reasoning! :) 👍

PR #875, opened 2018-10-11
Extend httpProxy options to allow more options

jpeyper:
[http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) has a lot of other options available than just `target`

This PR is to extend the existing method of configuring the proxy to allow those other options to be used.

It also allows you specify multiple proxies against different contexts in the similar way the [webpack-dev-server proxy](https://webpack.js.org/configuration/dev-server/#devserver-proxy) does

### DIFF
--- a/README.md
+++ b/README.md
@@ -1107,6 +1107,7 @@ module.exports = {
   hot: false,
 
   // HTTP proxy specific requests to a different target
+  // For advanced usage see https://github.com/react-cosmos/react-cosmos/pull/875
   httpProxy: {
     context: '/api',
     target: 'http://localhost:4000/api'
--- a/packages/react-cosmos-flow/config.js
+++ b/packages/react-cosmos-flow/config.js
@@ -7,6 +7,8 @@ export type ExcludePatterns = ExcludePattern | Array<ExcludePattern>;
 type WebpackConfig = Object;
 type WebpackConfigOverride = (WebpackConfig, { env: string }) => WebpackConfig;
 
+export type PluginConfig = { [prop: string]: mixed };
+
 type BasicHttpProxyConfig = { context: string };
 type AdvancedHttpProxyConfig = { [contextKey: string]: string | {} };
 export type HttpProxyConfig = BasicHttpProxyConfig | AdvancedHttpProxyConfig;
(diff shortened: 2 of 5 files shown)


jpeyper:
So... I'm not sure this is wanted by you/anyone else, however it would be really handy to us and our usage of react-cosmos. We have a complicated arrangement of runtime assets that are stored behind an authenticated request, so access to the `auth` (and `path-rewrite`) proxy options is required.

I'm not sure I got the flow types right... I was struggling to get everything up and running on my pc in this corporate environment.

I also couldn't find any existing tests that cover the the `httpProxy`config, but if I have missed adding any tests I should add please point me in the right direction.

Also, this is one of my first commits ever on an open source library I didn't control, so if I've made any mistakes in approaching this, please forgive me and let me know so I can learn :)

jpeyper:
Apparently I did get the types wrong... help?

skidding:
Hi @jpeyper,

Thanks for your thoughtful PR!

I think Flow can't decide between the two cases because `httpProxy.context` can be present in each. You might get a passing result by changing the if to `if (typeof httpProxy.context === 'string') {`. If it still doesn't work let me know and I'll check out your branch and figure it out. Flow can be stubborn :)

As a minimal hint that this feature exists, I advise adding the following line in the README docs:

```diff
   // HTTP proxy specific requests to a different target
+  // For advanced usage see https://github.com/react-cosmos/react-cosmos/pull/875
   httpProxy: {
     context: '/api',
     target: 'http://localhost:4000/api'
   },
```

jpeyper:
> As a minimal hint that this feature exists, I advise adding the following line in the README docs:
> 
> ```diff
>    // HTTP proxy specific requests to a different target
> +  // For advanced usage see https://github.com/react-cosmos/react-cosmos/pull/875
>    httpProxy: {
>      context: '/api',
>      target: 'http://localhost:4000/api'
>    },
> ```

Ok, I've added the minimal message to the README.

Thanks for your pointers, I managed to get the types work out (I think... please check them carefully for me).

skidding:
Excellent @jpeyper! 👏

skidding:
This contribution has been released under `react-cosmos@4.7.0-17` (beta channel).

Issue #544, opened 2017-12-09
Sorting of HOC'd Components in File Tree


maxsalven:
### What's up?

Components with HOC names get sorted based on their actual component names (`withRouter(Import)`) rather than their parsed component names (`Import withRouter`), which means the order in which we display components can be 'wrong' with respect to what UI displays.

<img width="225" alt="screen shot 2017-12-09 at 5 04 43 pm" src="https://user-images.githubusercontent.com/4144910/33794268-6be63e82-dd03-11e7-8ed9-fad84d3e7ef0.png">

skidding:
Fixed by #876

PR #876, opened 2018-10-12
Ignore HOC names when sorting components

tryggvigy:
Hey, awesome project!

I want to get your thoughts on this before writing more code/tests.

## The issue
image

These components should semantically, stand side-by-side and be sorted under "T". But a few of them are connected components (the hoc from react-redux) and that pulls them to be sorted under "C".

## This PR will
skips all HOC names when components are sorted.

Another approach could be to allow users to provide a `sortComponentsBy: () =>{}` function through `cosmos.config.js`.

### DIFF
--- a/packages/react-cosmos-voyager2/src/client/get-components.js
+++ b/packages/react-cosmos-voyager2/src/client/get-components.js
@@ -152,7 +152,7 @@ export function getComponents({
     });
   }
 
-  return sortBy(components, getObjectPath);
+  return sortBy(components, sortComponentsBy);
 }
 
 function getFileNameFromPath(filePath: string) {
@@ -215,4 +215,14 @@ function warnAboutIncompatFixtures(
 
 function getObjectPath(obj: { name: string, namespace: string }): string {
   return obj.namespace ? `${obj.namespace}/${obj.name}` : obj.name;
+}
+
+function sortComponentsBy(obj: { name: string, namespace: string }): string {
+  // remove all trailing ')', split string by '(' and get the first part.
+  // This gives the component name removed of all potential HOC names.
+  const componentName = obj.name
+    .replace(/\)/g, '')
+    .split('(')
+    .pop();
+  return obj.namespace ? `${obj.namespace}/${componentName}` : componentName;
 }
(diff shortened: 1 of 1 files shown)


tryggvigy:
> Another approach could be to allow users to provide a sortComponentsBy: () =>{} function through cosmos.config.js.

This could perhaps be provided through the `runtimeConfig` introduced in https://github.com/react-cosmos/react-cosmos/pull/877

skidding:
Hi @tryggvigy,

This looks good! I do have some suggestions:

- You can use regular expression to achieve the same thing. Eg.   
   ```js
   'connect(withRouter(MyComponent))'.replace(/^(.*\()?(.+?)\)*$/, '$2') // MyComponent
   ```
- The regexp makes it bit harder to read, but I think you can make the function self explanatory by calling it something like `stripHocNamesFromComponentName`, which also makes the current comments redundant. Instead, it would be more interesting to add some comments with component name examples and their respective return value, so that the reader understands the possible function input and output. Like the ones used [here](https://regex101.com/r/xqysHx/1/) to test the regexp.

> I want to get your thoughts on this before writing more code/tests.

Don't worry too much about tests as I plan to refactor this part of the UI anyway.

> > Another approach could be to allow users to provide a sortComponentsBy: () =>{} function through cosmos.config.js.
>
> This could perhaps be provided through the `runtimeConfig` introduced in https://github.com/react-cosmos/react-cosmos/pull/877

We can't put unserializable options (_function_ in this case) in cosmos.config because it is not imported directly in the client bundle. cosmos.config uses Node code so isn't suited for bundling. Instead, some serialiable options are stringified and safely injected into the client bundle. But I'll address this in the other PR.

tryggvigy:
> We can't put unserializable options (function in this case) in cosmos.config 

Ah, of course! 🤦‍♂️ 

> I do have some suggestions:

Your suggestions make perfect sense to me. I'll make the changes when I have some time!

tryggvigy:
Renamed the function, used the new regexp and added some explanatory comments in 6b6fc60, feel free to take a look when you get the time @skidding

skidding:
Excellent! 👏

Fixes #544

skidding:
Try it out under `react-cosmos@4.7.0-16`

PR #877, opened 2018-10-12
Add runtimeConfig and preferCollapsedFixtures

tryggvigy:
Hey, me again with an idea as a pull request. 
This is very much just a spike to get some thoughts on.
If this is desired we can discuss how to do this in a good way.

## This PR will
- add the optional `runtimeConfig` key to `cosmos.config.js`. 
  - The `runtimeConfig` key will be passed to the `mountPlayground` and therefor be accessible in the playground.
  - The idea behind this config is to allow customizations to the runtime. For example `runtimeConfig.preferCollapsedFixtures`
- Add `runtimeConfig.preferCollapsedFixtures`. If true it will cause all fixtures that do not have a saved expansion state to be collapsed by default.
  -  image
  - partially discussed in https://github.com/react-cosmos/react-cosmos/issues/251

"Hide all/ Show all" buttons + saved expansions state, would achieve the same thing more or less and I think thats been discussed in the past. If thats the plan, feel free to close this PR.

### DIFF
--- a/packages/react-cosmos-flow/config.js
+++ b/packages/react-cosmos-flow/config.js
@@ -7,6 +7,8 @@ export type ExcludePatterns = ExcludePattern | Array<ExcludePattern>;
 type WebpackConfig = Object;
 type WebpackConfigOverride = (WebpackConfig, { env: string }) => WebpackConfig;
 
+type RuntimeConfig = { preferCollapsedFixtures: boolean };
+
 export type PluginConfig = { [prop: string]: mixed };
 
 export type Config = {
@@ -30,6 +32,7 @@ export type Config = {
   watchDirs: Array<string>,
   modulesPath: string,
   plugin: PluginConfig,
+  runtimeOptions?: RuntimeConfig,
   // Deprecated
   proxies?: Array<string>,
   componentPaths: Array<string>,
--- a/packages/react-cosmos-playground/cosmos.config.js
+++ b/packages/react-cosmos-playground/cosmos.config.js
@@ -1,5 +1,8 @@
 module.exports = {
   globalImports: ['src/utils/global.less', 'src/register-plugins'],
   exclude: [/_shared/],
-  port: 8090
+  port: 8090,
+  runtimeConfig: {
+    preferCollapsedFixtures: true
+  }
 };
--- a/packages/react-cosmos-playground/src/components/FixtureList/data-mapper.js
+++ b/packages/react-cosmos-playground/src/components/FixtureList/data-mapper.js
@@ -4,10 +4,10 @@ import sortBy from 'lodash.sortby';
 
 const NODE_ORDER_BY_TYPE = ['directory', 'component'];
 
-function getExandedValue(savedExpansionState, path) {
+function getExandedValue(savedExpansionState, path, defaultValue = true) {
   return Object.prototype.hasOwnProperty.call(savedExpansionState, path)
     ? savedExpansionState[path]
-    : true;
+    : defaultValue;
 }
 
 function parseFixtureArray(componentName, fixtureArray, savedExpansionState) {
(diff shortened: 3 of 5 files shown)


tryggvigy:
>  I hope this won't piss people off, but I say we just switch to collapsing fixtures by default for now. What do you think?

I agree 100%. When skimming the list I think most folks will be looking for components not states. So allotting screen space to components over states by default makes sense.

I'll make the changes (remove the runtime config and collapse fixtures by default) when I get some time!

tryggvigy:
@skidding I removed the runtime config and default to collapsed fixtures in ca5dfe2

skidding:
Hi @tryggvigy,

The build is failling. Can you please update the tests to reflect your changes?

tryggvigy:
> The build is failling. Can you please update the tests to reflect your changes?
Fixed in 688eb74

tryggvigy:
noticed integration tests failing. I'll take a look when I can

skidding:
Oh, right! 😅

The Cypress tests that actually click through the interface expected all fixtures to be visible. They're all in [this file](https://github.com/react-cosmos/react-cosmos/blob/f389fe0d93df01da192a7331f4d92799027d0be4/cypress/integration/example-local-state.js). You should be able to add another step to toggle the component and then click on fixture.

Thanks!

tryggvigy:
I tried 

```diff
+ const componentNameButton = getSelector('index__componentName');
// The first menu button is the fixture editor toggle
const editorButtonSel = `${getSelector('index__button')}:eq(1)`;

function selectFixture(fixtureName) {
+   // expand the fixture
+   cy.get(componentNameButton).click();

  cy.get(fixtureButtonSel)
    .contains(fixtureName)
    .click();
}
```

But that times out :/ 
`CypressError: Timed out retrying: Expected to find element: '[class^="index__componentName__"]', but never found it.`

`index__componentName___xxxxx` seems to be the cssModules class that expands/collapses fixture

skidding:
Yeah, it got a bit tricker. The Cypress tests are no longer deterministic, because of how collapsed state is stored into LocalStorage. Initially components are collapsed, but after the first run they will be open. So toggling blindly isn't enough.

How about this? 😎

```diff
diff --git a/cypress/integration/example-local-state.js b/cypress/integration/example-local-state.js
index 627d72c2..433db058 100644
--- a/cypress/integration/example-local-state.js
+++ b/cypress/integration/example-local-state.js
@@ -1,13 +1,26 @@
 import { getSelector } from '../support/utils/css-modules';
 
+const componentButtonSel = getSelector('index__componentName');
 const fixtureButtonSel = getSelector('index__fixture');
 // The first menu button is the fixture editor toggle
 const editorButtonSel = `${getSelector('index__button')}:eq(1)`;
 
-function selectFixture(fixtureName) {
+function selectFixture(componentName, fixtureName) {
   cy.get(fixtureButtonSel)
     .contains(fixtureName)
-    .click();
+    .then($fixtureBtn => {
+      if ($fixtureBtn.is(':visible')) {
+        return $fixtureBtn.click();
+      }
+
+      return cy
+        .get(componentButtonSel)
+        .contains(componentName)
+        .then($componentBtn => {
+          $componentBtn.click();
+          $fixtureBtn.click();
+        });
+    });
 }
 
 function toggleFixtureEditor() {
@@ -60,7 +73,7 @@ describe('Local state example', () => {
 
   context('select fixture', () => {
     beforeEach(() => {
-      selectFixture('one-two-three');
+      selectFixture('CounterList', 'one-two-three');
     });
 
     it('should add active class to fixture button', () => {
@@ -82,7 +95,7 @@ describe('Local state example', () => {
 
   context('fixture editor', () => {
     beforeEach(() => {
-      selectFixture('one-two-three');
+      selectFixture('CounterList', 'one-two-three');
       toggleFixtureEditor();
     });
 
@@ -107,7 +120,7 @@ describe('Local state example', () => {
 
   context('fixture update', () => {
     beforeEach(() => {
-      selectFixture('one-two-three');
+      selectFixture('CounterList', 'one-two-three');
       toggleFixtureEditor();
 
       // Click three times on the first button
```

tryggvigy:
> The Cypress tests are no longer deterministic, because of how collapsed state is stored into LocalStorage. Initially components are collapsed, but after the first run they will be open. So toggling blindly isn't enough.

ah ofc! That makes sense. Yeah that looks great!

PR #878, opened 2018-10-15
Enable Apollo proxy cache configuration for mocked fixture

xavczen:
:wave: , hello there!

This PR:
- adds support for Apollo `InMemoryCache` configuration (see https://www.apollographql.com/docs/react/advanced/caching.html#configuration), which is dank useful;
- fixes a bug to avoid execution of `failWith` & `resolveWith` (fixes fixtures `mock-simple-error` and `real-error`)


skidding:
Try it out under `4.7.0-16`

PR #893, opened 2018-11-15
feat: Enable multiple remote renderers

skidding:


skidding:
This one was a bit of an indulgence on my part, but its implications on the long run are huge. Check this out:

multiple-renderers

Issue #899, opened 2018-11-27
Getting Started Steps outdated


brianespinosa:
### What's wrong?

The Getting Started steps in the README will technically not work right now because the babel-loader that gets installed if you to not ask for an exact version will be using the newest version of Babel. The loader will be running a new version of Babel but the old version of the presets.

The steps only work when running babel-loader > 7.x.x

### Steps to reproduce

1.  Go through the setup steps on an empty project using the default config
2.  Install the suggested peer dependencies
3.  Babel error when building fixture: `Plugin/Preset files are not allowed to export objects, only functions.`

**NOTE:** I have not yet taken the time to get Cosmos working with these steps and the newest babel-loader and `@babel/preset-env` and `@babel/preset-react`. I will update this issue with more information once I get there. Ideally we can update the Get Started steps with both.

brianespinosa:
Related: https://github.com/babel/babel/issues/6808

brianespinosa:
Got the steps working for the latest babel-loader and the correct presets in a bare node project. I'm going to submit a PR with updates to the README shortly.

PR #900, opened 2018-11-27
Update Getting Started steps for latest @babel

brianespinosa:
Solves #899 for outdated getting started steps. These updated steps now work with the latest Babel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,13 +133,15 @@ Add package.json scripts
 
 Run `npm run cosmos` or `yarn cosmos` and go to [localhost:8989](http://localhost:8989) 🎉
 
-> If you rely on the default webpack config, make sure to install the Babel and webpack plugins yourself. Depending on your needs, you'll probably want `babel-preset-env babel-preset-react babel-loader style-loader css-loader html-webpack-plugin`. Finally, add `.babelrc` to your project root.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,14 +133,16 @@ Add package.json scripts
...
+> If you rely on the default webpack config, make sure to install the Babel and webpack plugins yourself. Depending on your needs, you'll probably want `@babel/core @babel/preset-env @babel/preset-react babel-loader style-loader css-loader html-webpack-plugin`. Finally, add `.babelrc` to your project root.
 >
 > ```
 > {
->   "presets": ["env", "react"]
+>   "presets": ["@babel/env", "@babel/react"]
 > }
 > ```
 
+_NOTE: Babel v8 works differently than v7 and the package naming conventions changed. The above steps and Babel config are for the latest v8 version._
```
skidding: A few notes:
- Please put this note inside the quote above, to keep mentions of 3rd party deps together
- Babel packages are namespaces since 7.x, not 8.x (which doesn't exist, only for babel-loader which is webpack related)
- You can say something like "This example applies for Babel 7.x and above..."
- You may want to link to https://babeljs.io/docs/en/v7-migration as well

brianespinosa: Ah, thanks for the catch on the Babel version. The major version on babel-loader threw me off. I'll make these changes now.

I was trying to figure out why you had that quote block! Makes complete sense now.

PR #923, opened 2019-01-30
Proxy redux using peer-dependency

### LINKED ISSUE #903: react-cosmos-redux-proxy 4.6.1 doesn't compatible with react-redux 6.0.0
### What's wrong?

react-redux 6.0.0 was just released and there were some breaking changes: https://github.com/reduxjs/react-redux/releases
Using react-cosmos-redux-proxy 4.6.1 with react-redux 6.0.0 will get the following error:

> Could not find "store" in the context of "Connect(***)". Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(***) in connect options.

### Steps to reproduce

1.  Install "react-redux": "^6.0.0"
2. npm run cosmos

andyrichardson:
#### Description:
Add wider support for _redux_ and _react-redux_.

Fixes #903

#### Changes:
 * Use redux implementation for application-state managment (fixtures are wrapped in a _react-redux_ provider from peer dependencies).
 * Remove _createStore_ argument - _createStore_ from _redux_ peer dependency is used.
 * Remove _alwaysCreateStore_ argument - If a state is provided, store is created and Provider wraps fixture component. If a state is not provided, store is not created and the Provider is not rendered.
 * Remove _disableApplicationState_ argument - this can be added back if needed, I just wasn't sure what it was aiming to do (seemed unrelated to redux).

#### Todo:
 - [ ] E2E test all major versions of redux/react-redux
 - [ ] Document changes in README


skidding:
Hi @andyrichardson,

Thanks for working on this, but some existing functionality has been removed in the process. One reason why the Redux store is created by hand at the moment is to be able to listen to changes and update the fixture ([relevant piece of code](https://github.com/react-cosmos/react-cosmos/blob/a764a0a81a16973c94e7e8f918f34e4b316f2534/packages/react-cosmos-redux-proxy/src/index.js#L93-L100)).

The `disableLocalState` flag also has a purpose. Cosmos reads the loaded component's state automatically and shows it in the fixture editor. Redux-connected components (those generated by `connect` HoC) have state, but it's cached (I think) Redux state which shouldn't be shown to the user because the Redux state is already captured. So when use Redux-connected components we disable local state (pass down a flag to inner proxies) to avoid this duplication.

You can see the Redux state being updated in the the main README gif.

skidding:
Maybe publish a separate package for react-redux6+ support. Eg. `react-cosmos-redux-proxy2` or `react-cosmos-redux-proxy6` (to match the version number). And if you want to skip the whole fixture update thing (which greatly simplifies this proxy) you can maintain this proxy in a personal repo and we'll link to it in the Redux part of the docs. I think this is the most effortless approach to add support for react-redux6+ users. And I wouldn't invest _too_ much in this proxy anyway since `Cosmos Next` will require a different approach.

andyrichardson:
Hey @skidding, thanks for the feedback 👍 I knew I was missing something - looks like I'll have to revisit some of this (will close for the time being).

PR #927, opened 2019-02-06
Adding Github Pages instructions to the Readme fixes #925

### LINKED ISSUE #925: Github Pages needs .nojekyll file
We want to use cosmos as a communciation tool with a client but I couldn't get `cosmos-export` to run correctly on GHPages. It kept throwing 404 errors on the playground file:

    GET https://myaccount.github.io/whatever/_playground.js 404

It turns out that Github Pages uses Jekyll which filters out any file starting with an underscore so `_playground.js` and other files will be gone when `index.html` needs them

The solution for me was to put an empty `.nojekyll` file in my /public/ folder. 

https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/

Just thought I'd post this here to save anyone else trouble. It took me forever to figure out this wasn't a problem with Cosmos at all.

MattReimer:
Just a quick note about the `.nojekyll` file needed when rendering on Github Pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Jump to:
 - [Integration with popular tools](#integration-with-popular-tools)
   - [Create React App](#create-react-app)
     - [With react-app-rewired](#with-react-app-rewired)
+  - [Github Pages](#github-pages-and-jekyll)
   - [Next.js](#nextjs)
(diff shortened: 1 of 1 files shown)


skidding:
Excellent. Thanks for sharing your wisdom!

PR #957, opened 2019-05-04
Add support for overriding webpack config

### LINKED ISSUE #676: Call webpack config function with env object
### What's up?

Right now a function webpack config is called with env string (eg. "development" or "production"). But webpack calls it with all environment vars, which looks like `{ prod: true }`.

* [Read about the feature here](https://github.com/react-cosmos/react-cosmos#custom-webpack-config)
* [Here is where the function is current called](https://github.com/react-cosmos/react-cosmos/blob/4a13b775617a031817008a8611e70db515dc355e/packages/react-cosmos/src/server/extend-webpack-config.js#L43)

### Mkay, tell me more...

- https://webpack.js.org/configuration/configuration-types/#exporting-a-function
- https://webpack.js.org/api/cli/#environment-options
- https://webpack.js.org/guides/environment-variables/

skidding:
Fixes #676


skidding:
@kevin940726 This feature allows you to extend your webpack config, as discussed on Slack.

I didn't pass the base config to the exported user config because that would violate webpack's API and could result in some user webpack configs not being compatible with Cosmos. Instead, I mirrored the arguments webpack passes to function configs as defined here: https://webpack.js.org/configuration/configuration-types/#exporting-a-function.

Additionally, to give users the possibility to extend their config without creating a new one, I added `overridePath` which works similar to an existing config in Cosmos Classic. I documented webpack configuration in Cosmos Next: https://github.com/react-cosmos/react-cosmos/blob/d4c509f2c170e9a5063129513865a040a182ede1/NEXT.md#webpack. Can you please take a look and let me know if it covers all your needs?

Otherwise this is ready to merge and release.

kevin940726:
@skidding Looks good! I just built it and tested myself, it works like a charm, thx! (whoa I didn't know that webpack supports function config natively)

Issue #946, opened 2019-04-11
Yarn PnP compatibility


stellarhoof:
### What's wrong?

When using yarn plug n play, I cannot run cosmos since it tries to import the `debug` package without having it declared in its dependencies (although I don't know if this is because yarn is more strict with respect to dependencies). I cloned the react-cosmos repo and added `debug` to `packages/react-cosmos/package.json` and it worked.

`react-cosmos-voyager2` is missing `lodash`

stellarhoof:
Now `packages/react-cosmos` is missing `slash`. It'd be useful if the entire repo was ran through yarn pnp to figure out missing ones since it has stricter dependency requirements.

skidding:
@stellarhoof Thanks for reporting these!

What's the best way to check? I just ran `yarn --pnp` in the Cosmos repo root and it didn't complain.

skidding:
These issues should be fixed in `v5.0.0-alpha.2` (presuming you're interested in Cosmos Next).

But let me know if you have an answer to my previous question. I'd like to include a small CI step to prevent this from happening in the future. Thanks!

stellarhoof:
Getting there, but there are still missing deps. To reproduce, make an empty directory with a stub `package.json` like
```
{
  "name": "stub",
  "version": "0.0.0",
  "installConfig": {
    "pnp": true
  },
  "scripts": {},
  "dependencies": {
    "react-cosmos": "5.0.0-alpha.2"
  }
}
```
Then do `yarn install` followed by `yarn cosmos`

skidding:
Hmm, maybe I'm not grasping how pnp works, but here's an error I don't understand.

> ERROR in /Users/ovidiu/Library/Caches/Yarn/v4/npm-react-cosmos-5.0.0-alpha.2-f4ccdad5b29e38ed61b443fe03a3b3f7e0c28201/node_modules/react-cosmos/dist/plugins/webpack/client/errorOverlay/reactErrorOverlay.js
Module not found: Error: **Can't resolve 'react-error-overlay'** in '/Users/ovidiu/Library/Caches/Yarn/v4/npm-react-cosmos-5.0.0-alpha.2-f4ccdad5b29e38ed61b443fe03a3b3f7e0c28201/node_modules/react-cosmos/dist/plugins/webpack/client/errorOverlay'

The `react-cosmos` package _does_ have `react-error-overlay` as a dependency.

[...]

PR #989, opened 2019-07-11
fixed peerDependencies

marco-silva0000:
This fix will allow for this lib to be used with yarn berry and yarn pnp without having to manually hack into the yarn.lock file.

when forking the inicial branch, upgrading to yarn berry, i had the warnings at end of post.
These warnings cause errors when using cosmos on my packages with PNP.
I have a branch on my fork with this library on yarn2, but I haven't tested it.

With these fixes the warnings go away(the ones related to this lib). This is related to #946 so hopefully we can move on that issue.

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ @typescript-eslint/parser@npm:1.11.0 doesn't provide typescript@* requested by @typescript-eslint/experimental-utils@npm:1.11.0
➤ YN0002: │ react-cosmos-shared2@workspace:packages/react-cosmos-shared2 doesn't provide react@^0.14.8 || ^15.0.1 || ^16.0.0 requested by react-element-to-jsx-string@npm:14.0.2
➤ YN0002: │ react-cosmos-shared2@workspace:packages/react-cosmos-shared2 doesn't provide react-dom@^0.14.8 || ^15.0.1 || ^16.0.0 requested by react-element-to-jsx-string@npm:14.0.2
➤ YN0002: │ react-cosmos-playground2@workspace:packages/react-cosmos-playground2 doesn't provide webpack@^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0 requested by html-webpack-plugin@npm:3.2.0
➤ YN0002: │ react-cosmos@workspace:packages/react-cosmos doesn't provide webpack@^4.0.0 requested by webpack-dev-middleware@npm:3.7.0
➤ YN0002: │ react-cosmos-example@workspace:example doesn't provide webpack@^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0 requested by html-webpack-plugin@npm:3.2.0
➤ YN0000: └ Completed in 0.19s
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 1.11s
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 4.05s
➤ YN0000: Done with warnings in 5.46s
```

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -8,7 +8,10 @@
     "start": "cosmos"
   },
   "dependencies": {
-    "react-cosmos": "^5.0.0-alpha.15"
+    "react": "^16.0.0",
+    "react-cosmos": "^5.0.0-alpha.15",
+    "react-dom": "^16.0.0",
+    "webpack": "^4.0.0"
   },
   "devDependencies": {
     "html-webpack-plugin": "^3.2.0"
--- a/packages/react-cosmos-fixture/package.json
+++ b/packages/react-cosmos-fixture/package.json
@@ -11,5 +11,9 @@
     "react-cosmos-shared2": "^5.0.0-alpha.14",
     "react-is": "^16.8.6",
     "socket.io-client": "^2.2.0"
+  },
+  "peerDependencies": {
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
   }
 }
--- a/packages/react-cosmos-playground2/package.json
+++ b/packages/react-cosmos-playground2/package.json
@@ -16,5 +16,8 @@
     "react-plugin": "^2.0.0-alpha.6",
     "socket.io-client": "^2.2.0",
     "styled-components": "^4.3.2"
+  },
+  "peerDependencies": {
+    "webpack": "^4.0.0"
   }
 }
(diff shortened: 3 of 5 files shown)


skidding:
@marco-silva0000 Thanks for looking into this. I'm not sure this is the right approach, though.

First of all:
- I don't see why any package would require `webpack` as a dependency (let alone react-cosmos-playground2 which is published prebuilt), because only the server uses webpack and it imports it dynamically (at runtime) if it detects it in the user's node_modules.
- The example package shouldn't really matter to users because it's not published.

But most importantly, and probably what _does_ help your situation:
- Adding `react` and `react-dom` to peerDependencies does sound like a good idea. They used to be there in the past. The reason I removed them is because people who use stuff like create-react-app which has react/react-dom as a transitive dependency (not explicit in the user's package.json) will get annoying warnings whenever they do npm/yarn install saying Cosmos depends on react and react is missing, when in fact react is NOT missing, it's just installed by create-react-app (or similar) implicitly.

Let me know if you have any thoughts on this. I would definitely like to solve the PnP situation but I don't to introduce other issues at the same time. Thanks again for taking the time to fix this!

skidding:
Closing this because it's outdated. Please check out [this comment](https://github.com/react-cosmos/react-cosmos/issues/946#issuecomment-531263762).

PR #1048, opened 2019-09-23
Update useValue example in README

johnrees:


johnrees:
Actually I misunderstood the code, I guess `'number'` should be `'count'` in this case

skidding:
@johnrees You're right, please replace with `count` and I'll merge this. Sorry for using the wrong example. 🤦‍♂️

johnrees:
No worries @skidding, thanks for the sharing the project! I've force-pushed the update to this PR

PR #1077, opened 2019-10-21
Fix building and linting on Windows

arseneyr:
Builds and lints on Windows. Note that several unit tests still fail because of inconsistent path handling. For example, `mockFs.ts` stores file paths with forward slashes but `detectCosmosConfigPath()` uses `path.resolve`. However, the number of code changes to eliminate all test failures seems too invasive for this PR. 

Personally, I think converting to forward slashes **everywhere** on Windows is more trouble than it's worth. From what I can tell, the only place they are useful is when generating the `userDeps` file/string because backslashes would have to be escaped again to generate the correct Javascript. Everywhere else, the ultimate consumer of paths is some Node API that can handle either separator just fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "prettify": "prettier --write \"{{packages/*/src,examples,cypress,scripts}/**/,}*.{js,jsx,ts,tsx,css,md}\"",
     "check-types": "tsc",
-    "lint": "eslint '**/*.{ts,tsx}'",
+    "lint": "eslint \"**/*.{ts,tsx}\"",
     "test:unit": "jest --config .jest/config.js",
     "test:watch": "yarn test:unit --watch",
     "test:coverage": "yarn test:unit --coverage",
--- a/packages/react-cosmos-playground2/webpack.config.js
+++ b/packages/react-cosmos-playground2/webpack.config.js
@@ -36,7 +36,7 @@ module.exports = {
     rules: [
       {
         test: /\.tsx?$/,
-        include: [/packages\/react-cosmos-playground2\/src/],
+        include: src,
         use: {
           loader: 'babel-loader'
         }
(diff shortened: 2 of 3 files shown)


skidding:
@arseneyr Thanks, I appreciate the clean PR!

> Personally, I think converting to forward slashes **everywhere** on Windows is more trouble than it's worth. From what I can tell, the only place they are useful is when generating the `userDeps` file/string because backslashes would have to be escaped again to generate the correct Javascript. Everywhere else, the ultimate consumer of paths is some Node API that can handle either separator just fine.

[This](https://github.com/react-cosmos/react-cosmos/blob/8c4b1b162c8bcc80d7add3f962461211e0c880ed/packages/react-cosmos/src/shared/slash.ts#L7-L10) is the main reason why. If it weren't for `node-glob` we probably wouldn't bother. And if there are paths we never _glob_ I think they can safely be stored in their native form.

> Note that several unit tests still fail because of inconsistent path handling.

Can you show me a screenshot or a list of the failing tests? Depending on the volume, it might be worse solving this as well to make the project more Windows friendly.

arseneyr:
[Here's the test log.](https://github.com/react-cosmos/react-cosmos/files/3758336/windows_test_unit.txt)

> This is the main reason why. If it weren't for node-glob we probably wouldn't bother. And if there are paths we never glob I think they can safely be stored in their native form.

Right, I saw that comment and the referenced issue. Searching through the code, the only globbing I found (besides a few build scripts) was for fixture and decorator files in `findUserModulePaths()`. And these patterns are hardcoded in `getFixturePatterns()`/`getDecoratorPatterns()`, so no slash conversion is required there. Perhaps there was more globbing going on with the proxies in the previous version of Cosmos?

skidding:
> [Here's the test log.](https://github.com/react-cosmos/react-cosmos/files/3758336/windows_test_unit.txt)

Thanks. 7 tests, 4 test suits failing, seems manageable.

> Perhaps there was more globbing going on with the proxies in the previous version of Cosmos?

Probably. My memory is a bit fuzzy after almost 3 years...

If you ever want to remove `slash` calls wherever unnecessary and fix those tests on Windows I'm for it, and I can do my share of testing on macOS.

PR #1078, opened 2019-10-23
Advanced httpProxy options

LogKal:
Advanced httpProxy options include: target, secure, pathRewrite, and logLevel.
Advanced options are optional, httpProxy will accept either a string value to the target, or an object containing the advanced configuration.

Information on exactly how these options are used within http-proxy-middleware are found [here](https://github.com/chimurai/http-proxy-middleware#http-proxy-options).

Example cosmos.config.json:
```
{
  "globalImports": ["polyfills.ts", "globalStyle.css"],
  "staticPath": "static",
  "httpProxy": {
    "/api": {
      "target":"http://localhost:5001",
      "secure": false,
      "pathRewrite":{
        "^/api":""
      },
      "logLevel": "debug"
    },
    "/test": "http://localhost:5002"
  }
}
```

@skidding please review the changes to httpProxy/config.schema.json that we discussed briefly via Slack

### DIFF
--- a/packages/react-cosmos/src/plugins/httpProxy.ts
+++ b/packages/react-cosmos/src/plugins/httpProxy.ts
@@ -2,12 +2,6 @@ import { CosmosConfig } from '../config/shared';
 import httpProxyMiddleware from 'http-proxy-middleware';
 import { DevServerPluginArgs } from '../shared/devServer';
 
-// TODO: Support advanced configuration
-// See https://github.com/react-cosmos/react-cosmos/pull/875 for context
-// Supporting all httpProxy options isn't as easy anymore, because the Cosmos
-// config is now JSON. We can either support a serializable subset of useful
-// options (like "pathRewrite") or optionally support an external config module.
-
 type HttpProxyConfig = {
   [context: string]:
     | string
@@ -15,20 +9,20 @@ type HttpProxyConfig = {
         target: string;
         secure?: boolean;
         pathRewrite?: { [rewrite: string]: string };
-        logLevel?: string;
+        logLevel?: 'error' | 'debug' | 'info' | 'warn' | 'silent';
       };
 };
 
 export function httpProxy({ cosmosConfig, expressApp }: DevServerPluginArgs) {
   const httpProxyConfig = getHttpProxyCosmosConfig(cosmosConfig);
   Object.keys(httpProxyConfig).forEach(context => {
+    let proxy = {};
     if (typeof httpProxyConfig[context] === 'string') {
-      const target = httpProxyConfig[context] as string;
-      expressApp.use(context, httpProxyMiddleware(context, { target: target }));
+      proxy = { target: httpProxyConfig[context] };
     } else if (typeof httpProxyConfig[context] === 'object') {
-      const proxy = httpProxyConfig[context] as object;
-      expressApp.use(context, httpProxyMiddleware(proxy));
+      proxy = httpProxyConfig[context];
     }
+    expressApp.use(context, httpProxyMiddleware(proxy));
   });
 }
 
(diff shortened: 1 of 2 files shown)


skidding:
@LogKal Thanks for this.

A few comments
- You can remove the `TODO` as this PR adds support for advanced proxy config.
- Any idea the diff shows all lines as changed? This might be a GitHub bug. In any case, make sure to run Prettier on the changed files. If your IDE doesn't do it automatically, you can manually run `yarn prettier --write path/to/file`.
- I think you can make TypeScript play along without forced casting (`as`). I believe you need to make logLevel compatible with `http-proxy-middleware` first. This should work:
```ts
type HttpProxyConfig = {
  [context: string]:
    | string
    | {
        target: string;
        secure?: boolean;
        pathRewrite?: { [rewrite: string]: string };
        logLevel?: 'error' | 'debug' | 'info' | 'warn' | 'silent';
      };
};

// ...
  Object.keys(httpProxyConfig).forEach(context => {
    const config = httpProxyConfig[context];
    if (typeof config === 'string') {
      expressApp.use(context, httpProxyMiddleware(context, { target: config }));
    } else if (typeof config === 'object') {
      expressApp.use(context, httpProxyMiddleware(config));
    }
  });
```
- Also update the `logLevel` in the schema with the predefined values. And try this approach for the `patternProperties` of `httpProxy`:
```json
"patternProperties": {
  ".*": {
    "anyOf": [
      { "type": "string" },
      {
        "type": "object",
        "description": "Advanced HTTP proxy config",
        "additionalProperties": true,
        "required": ["target"],
        "properties": {
          "target": { "type": "string" },
          "secure": { "type": "boolean" },
          "pathRewrite": {
            "type": "object",
            "patternProperties": {
              ".*": { "type": "string" }
            }
          },
          "logLevel": {
            "type": "string",
            "enum": ["error", "debug", "info", "warn", "silent"]
          }
        }
      }
    ]
  }
}
```

LogKal:
@skidding The requested changes have been made. I am not sure why the diff was not actually showing a diff, my latest commit shows an accurate diff. Sorry about the formatting and empty commits, I am not very experienced with git/github and have not used prettier before, but I intend on using it now! 👍 Also fairly new to TS.

I am however happy to make contributions and learn when I am able to.

packages/react-cosmos/src/plugins/httpProxy.ts
```diff
@@ -1,22 +1,31 @@
...
+export function httpProxy({ cosmosConfig, expressApp }: DevServerPluginArgs) {
+  const httpProxyConfig = getHttpProxyCosmosConfig(cosmosConfig);
+  Object.keys(httpProxyConfig).forEach(context => {
+    let proxy = {};
+    if (typeof httpProxyConfig[context] === 'string') {
+      proxy = { target: httpProxyConfig[context] };
+    } else if (typeof httpProxyConfig[context] === 'object') {
+      proxy = httpProxyConfig[context];
+    }
+    expressApp.use(context, httpProxyMiddleware(proxy));
```
skidding: Sorry for being so nitpicky, but can please only register the middleware inside each `if` condition?

Right now if for whatever reason the config is neither a string nor an object (even though the schema advises against it) we register a proxy middleware with an empty config. For convenience, and to help TypeScript, you can put `httpProxyConfig[context]` in a constant before checking its type.

This is what I propose:

```ts
    const config = httpProxyConfig[context];
    if (typeof config === 'string') {
      expressApp.use(context, { target: config });
    } else if (typeof config === 'object') {
      expressApp.use(context, config);
    }
```

Thanks!

LogKal: Okay, no worries, I see what you are saying.

skidding:
@LogKal Figured out the diff error. You were using a different style of line breaks and GitHub doesn't signal that in an obvious way, like this:

image

Fixing in #1079

The good thing is I now added an ESLint rule that will warn about this and a Prettier option to normalize them automatically.

skidding:
OK, I published `react-cosmos@5.0.0-beta.16` with your change.

**You can start using the advanced HTTP config!**

I've also published an update to the Cosmos config in the JSON schema store: https://github.com/SchemaStore/schemastore/pull/829/files

Thank you for your contribution!

LogKal:
Awesome! Happy I was able to help out, will be on the lookout for other improvements, thanks for the chance to learn something new.

PR #1113, opened 2019-12-20
add support for async webpack config

### LINKED ISSUE #1114: Webpack config returning a Promise
### What's wrong?

My webpack config is a function returning a promise like in the official docs:
https://webpack.js.org/configuration/configuration-types/#exporting-a-promise

Bug I get the message 
```
WARNING in configuration
The 'mode' option has not been set, Webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/
```

I tested with an sync config and it works well.

Could it be possible to accept Webpack config as a potential promise.

### Steps to reproduce

```typescript
// working webpack.config.js
module.exports = function(env, argv) {
    return {mode: "development"};
};

// not working webpack.config.js
module.exports = async function(env, argv) {
    return {mode: "development"};
};
```

### User info

- Cosmos version 5.0.0
- Webpack version 4.39.0

GrigoriyZander:
Resolves #1114

According to [Webpack docs](https://webpack.js.org/configuration/configuration-types/#exporting-a-promise), webpack config can be exported as a Promise.
It seems like there are no extra tests needed for this changes.

### DIFF
--- a/packages/react-cosmos/src/plugins/webpack/devServer.ts
+++ b/packages/react-cosmos/src/plugins/webpack/devServer.ts
@@ -36,10 +36,10 @@ export async function webpackDevServer({
     return;
   }
 
-  const webpackConfig = getDevWebpackConfig(
+  const webpackConfig = (await getDevWebpackConfig(
     cosmosConfig,
     userWebpack
-  ) as WebpackConfig;
+  )) as WebpackConfig;
 
   // Serve static path derived from devServer.contentBase webpack config
   if (cosmosConfig.staticPath === null) {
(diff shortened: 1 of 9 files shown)


skidding:
This looks great! I'll take a closer look later today and publish a patch asap if everything's alright.

skidding:
Thank you @GrigoriyZander!

If only so contributions were so concise and effective ❤️

Please check out `5.0.1` and let me know if it works as expected.

GrigoriyZander:
Great!
Thank you for quick response. Works as expected.

PR #1172, opened 2020-03-29
Create better API for getting fixture info

skidding:
Resolves https://github.com/react-cosmos/rfcs/issues/14

- [x] Move shared functions from playground2 package to shared2 package
- [x] Add `cleanPath` array to let user choose text delimiter
- [x] Add `filePath` (full path to file, prepend `cosmosConfig.rootDir`)
- [x] Remove `fixtureId`?
- [ ] Add docs
  - [ ] Add docs for `fixturesSync`
  - [ ] Remove `getFixtureUrls` from docs (just docs!)
  - [ ] Add docs for `getFixtureInfo`
    - [ ] Mention why there is no `Sync` method (why it's sync by default compared to older API)
    - [ ] Mention added `http://` to URLs


skidding:
Hi @andyrichardson,

Here's a draft of the better API for querying fixture info on the sever side.

There's a bunch of file moves and renames that you needn't bother with, but here's the gist in case you're willing to take a look – no pressure!

There's a new method called `getFixtureInfo` (sync from the get go). The source is [here](https://github.com/react-cosmos/react-cosmos/blob/b4fedd50c481b3df5466dc473f653376c7ed467c/packages/react-cosmos/src/getFixtureInfo.ts), and the return type is an array of `FixtureInfo`.

https://github.com/react-cosmos/react-cosmos/blob/b990dfd80e87ef661041ce4db0979933d6ffe404/packages/react-cosmos/src/getFixtureInfo.ts#L21-L26

The other three are self explanatory, but `cleanPath` is an array of strings that represents the path of the fixture in the generated fixture tree. It's _clean_ because it doesn't contain special dirs and redundant hierarchies (using the same logic we use in the playground UI). It's an array to allow users to join the path in any way. You might do `cleanPath.join(' ')` or `cleanPath.join(' - ')`, etc. to create fixture identifiers.

[Here](https://github.com/react-cosmos/react-cosmos/blob/b990dfd80e87ef661041ce4db0979933d6ffe404/example/fixtureInfo.test.ts) is an example of the generated output:

https://github.com/react-cosmos/react-cosmos/blob/b990dfd80e87ef661041ce4db0979933d6ffe404/example/fixtureInfo.test.ts#L9-L31

Let me know if there's anything you would change, and if it's something you would use instead of the current APIs ✌️

andyrichardson:
I'll take a deeper look tomorrow but `getFixtureInfo` looks great! Particularly the naming (playgroundUrl and rendererUrl) 🔥

The only thing I can think off the top of my head is that _getFixtureInfo_ sounds singular - maybe something like _getFixtureItems_ would make it more obvious that a collection is returned (plucking at straws here)

skidding:
> The only thing I can think off the top of my head is that _getFixtureInfo_ sounds singular

Good point. I'm also not in love with the name. Not a fan of `getFixtureItems` either, because _items_ is such a generic term. But it would be great if the name contained a plural.

On a second thought...

Now I'm wondering if I should slap another property on the return type and swallow the `getFixtures` method as well, and have fewer APIs to think about.

```diff
type FixtureApi = { 
   cleanPath: string[]; 
   filePath: string; 
   playgroundUrl: string; 
   rendererUrl: string; 
+   getElement: () => React.ReactElement<any>
}; 
```

Of course it couldn't be called `getFixtures` for backwards compatibility, but I'm totally cool with creating a `getFixtures2` method until the next major version, or even beyond 😄.

At first I thought adding the element getter would increase the environment deps for this API, but actually I forgot we already import all fixture modules in `getFixtureUrls` as well, to get the exported fixture names. So I think we could get away with just one method to rule them all!

andyrichardson:
I think having all the info / accessors for a single fixture in one object is a good shout :+1:

One question about _cleanPath_, am I right in thinking it would work something like this?

- `['Component']` - Component.fixture.tsx w/ default export (not an object)
- `['Component', 'object key']` - Component.fixture.tsx w/ default export (object containing 'object key' property)

Can you see any advantages / disadvantages to using a structure like this?

```ts
type FixtureApi = {
  /** Fixture name (only on object export) */
  name?: string;
  fileName: string;
  path: string
  playgroundUrl: string; 
  rendererUrl: string; 
  getElement: () => React.ReactElement<any>
}; 
```

In all honesty, even if I were to regex the filepath to get the file name, it wouldn't be that big an issue - the biggest advantage of the changes you've done is having all the fixture metadata in one place!

skidding:
> One question about cleanPath, am I right in thinking it would work something like this?

You're right. You can't reliably tell what the last string of the clean path is. It can be the fixture name (default export object key), the file name, or even the folder name when you use `index.fixture` files.

So if you want the file name or fixture name specifically, the clean path is not transparent enough to use as a source reliably. I still think it's useful in many cases (it's what I would use for a fixture identifier), but maybe `treePath` is a better name?

> Can you see any advantages / disadvantages to using a structure like this?

I like it. It doesn't cost anything to offer more info. How about this?

```ts
type FixtureApi = {
  name: string | null; // 'object key'
  fileName: string; // 'Component'
  filePath: string; // /path/to/src/section/Component.fixture.tsx or /path/to/src/section/__fixtures__/Component.tsx
  treePath: string; // ['section', 'Component', 'object key']
  playgroundUrl: string; 
  rendererUrl: string; 
  getElement: () => React.ReactElement<any>
}; 
```

skidding:
One thing that's not clear to me is how `fileName` and `name` would be used when you organize fixtures by folders. Given the API above, how would use the returned properties to create a snapshot file name for example?

skidding:
I've made the changes mentioned above:

Renamed the function to `getFixtures2` for now and it returns this type:

https://github.com/react-cosmos/react-cosmos/blob/72f4d03b7f293b9e70283abefcc9e7db2a2f13dd/packages/react-cosmos/src/getFixtures2.ts#L31-L39

Here's an example output:

https://github.com/react-cosmos/react-cosmos/blob/72f4d03b7f293b9e70283abefcc9e7db2a2f13dd/example/fixtureInfo.test.ts#L8-L29

And here is the referenced fixture:

https://github.com/react-cosmos/react-cosmos/blob/72f4d03b7f293b9e70283abefcc9e7db2a2f13dd/example/Counter/index.fixture.tsx#L5-L19

Let me know what you think and how you would use this API to achieve what you're after. My only worry now is offering too much overlapping information 😅. Maybe it isn't obvious why `treePath` would be needed, and maybe for most people it's not needed, if they only have single or multi fixtures in a project. The example project combines all possible ways of defining a fixture for showcase purposes, but in a real project the file structure is usually more uniform.

andyrichardson:
That looks great to me 👏 

> One thing that's not clear to me is how fileName and name would be used when you organize fixtures by folders. Given the API above, how would use the returned properties to create a snapshot file name for example?

I've been doing something along the lines of `'${fileName} - ${filePath}'` but there is a chance of conflicts on duplicate file names.

skidding:
> I've been doing something along the lines of `'${fileName} - ${filePath}'` but there is a chance of conflicts on duplicate file names.

- I don't think there's a chance of conflicts because `filePath` is unique.
- From the look of [your reference directory](https://github.com/FormidableLabs/urql-devtools/tree/341070df25e9a1d1f467898a151320cc1cda885a/src/panel/__image_snapshots__) is seems you use something like `'${fileName} - ${name}'` – is this what you meant? If so indeed there is a chance of conflicts.
- Note that `filePath` is the _full_ disk path. I wonder if it would be useful to include both `absFilePath` and `relFilePath` (the latter would be relative to the `cosmosConfig.rootDir`, and isn't returned at the moment).
- I still think doing `treePath.join(' - ')` would output the file names of your current reference directory. But I don't mind including all the information because it doesn't cost and each piece of data might be most useful in some scenario.

Thanks for helping me iron out this API, btw! I know this back and forth can be tiring.

andyrichardson:
> From the look of your reference directory is seems you use something like '${fileName} - ${name}' – is this what you meant? If so indeed there is a chance of conflicts.

Yup, my bad!

> Thanks for helping me iron out this API, btw! I know this back and forth can be tiring.

Not at all, thanks for all the changes - looking forward to them 🙌

skidding:
Cool. Merged and released it under `react-cosmos@5.2.0-alpha.0` or `react-cosmos@next`. Please try it out and let me know if everything works as expected.

skidding:
Btw the fixture is called `getFixtures2` and is called like [this](https://github.com/react-cosmos/react-cosmos/blob/fd286f514b807fdb5d8815572212cbd84915b996/example/fixtures2.test.ts).

skidding:
Hey @andyrichardson,

One last request for this!

I've fixed a small bug, simplified the function signature to `getFixtures2(cosmosConfig)` and [updated the docs](https://github.com/react-cosmos/react-cosmos/tree/61285e3958af1c64dede47885f1d01589dfe9276/docs#fixtures-1). Can you please give the latest `5.2.0-alpha.8` another run on your end? Just want to make sure I didn't break anything and I'll let v5.2 loose.

🙏

andyrichardson:
For sure, will give it a shot first thing tomorrow morning 💯

andyrichardson:
Just did a quick pass - everything looks good!

 - Here's our [alpha run](https://app.circleci.com/pipelines/github/FormidableLabs/urql-devtools/643/workflows/686c7494-106e-46f4-ad6c-8cddd5678eb5/jobs/3515) and [stable run](https://app.circleci.com/pipelines/github/FormidableLabs/urql-devtools/640/workflows/534e4795-4964-4234-9726-c3041c55e752/jobs/3489) 
 - All 55 fixtures are being returned as expected :+1:
 - Diff is failing by a **very** slight percentage - pretty sure it's just my setup as I'm assuming no changes have been made to the renderer

skidding:
Indeed I didn't change anything to the renderer. It seems the `:` rendered slightly faded in one of the snapshots 😅.

Awesome,`5.2.0` is out in the open, and you've made a dent in React Cosmos 🎉

PR #1174, opened 2020-04-01
Add provide entry to cosmos override function

### LINKED ISSUE #1173: Unable to set/append entry on webpack config override
### What's wrong?

I'm looking to add some tweaks to the cosmos playground welcome page. I can't do this because the `entry` override gets removed instead of merged.

### Steps to reproduce

1. Create an cosmos.overrides.js
2. Set an entry value (array, object, string)
3. Run a cosmos build

### Suggestions

It would be great if the full cosmos webpack config was provided to the override function rather than a partial one. [Here](https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos/src/plugins/webpack/webpackConfig/export.ts#L24) is where the entry property is overriding the user-specified overrides.

andyrichardson:
## About
FIx #1173 

Allows users to add additional scripts to cosmos playground.

### DIFF
--- a/packages/react-cosmos/src/plugins/webpack/webpackConfig/default.ts
+++ b/packages/react-cosmos/src/plugins/webpack/webpackConfig/default.ts
@@ -1,7 +1,7 @@
 import resolveFrom from 'resolve-from';
 import webpack from 'webpack';
 import { RENDERER_FILENAME } from '../../../shared/playgroundHtml';
-import { getNodeEnv } from './shared';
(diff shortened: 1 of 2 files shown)


skidding:
See https://github.com/react-cosmos/react-cosmos/issues/1173#issuecomment-607167959. `globalImports` might be want you need.

andyrichardson:
Closing as it sounds like this change wouldn't apply to the react-cosmos playground

PR #1190, opened 2020-04-26
Correct direction in FixtureName

mrm007:
With original `direction: rtl`:

image

Notice how it's not visible when the fixture name doesn't contain a `>`:
image

With patch:

image

### DIFF
--- a/packages/react-cosmos-playground2/src/plugins/Root/RendererHeader.tsx
+++ b/packages/react-cosmos-playground2/src/plugins/Root/RendererHeader.tsx
@@ -168,5 +168,4 @@ const FixtureName = styled.div`
   white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
-  direction: rtl;
 `;


skidding:
I set the direction right-to-left on purpose, because on small screens I think it's more important to see the last part. For example, I think "...Icon clickable" is more useful than "public Icon...", since you have more fixtures that start with "public Icon".

Something doesn't seem right in your examples, though.
1. If the text is truncated, it should have ellipsis on one side "...".
2. It seems like you have enough space to show the entire fixture name
3. Why would the last `>` char become the first `<` char :|

Is your repo public, or can you share a small setup I can investigate? Thanks!

mrm007:
There is no truncation. The fixture name is `<default>`. This was done so it shows up first in the list. You can test the bug yourself by creating a fixture named `<default>`.

mrm007:
> Why would the last `>` char become the first `<` char :|

Because `direction: rtl` is set on the element. See this thread for more info: https://lists.freedesktop.org/archives/libreoffice/2012-March/028506.html (and the [attachment] mentioned).

[attachment]: http://lists.freedesktop.org/archives/libreoffice/attachments/20120319/a6153586/attachment-0001.png

skidding:
> Because `direction: rtl` is set on the element...

Every day I learn something new :). Thanks for sharing.

> If you think this isn't a bug feel free to close the PR. `[default]` as the fixture name also works for me.

I'm happy you found a workaround. There are many other name conventions of course which can serve the same purpose. And as a side note, if you use [multi fixture files](https://github.com/react-cosmos/react-cosmos/blob/a78ac966ad95f029d26aab3d782d01cab7b4fa01/docs/README.md#multi-fixture-files) the order of those fixtures is preserved as defined – maybe this format is better for you.

I will close this PR for now as using `<>` in the fixture name is highly uncommon in my experience, and I haven't found any other way to [put the ellipsis on the left side](https://stackoverflow.com/questions/9793473/text-overflow-ellipsis-on-left-side) while truncating, which I still prefer. That said, I'm now aware that `rtl` direction might have unwanted side effects, so I'll keep an eye on this and if I see other related issue in the future I will accept your change and give up the left-side truncation of the fixture name.

mrm007:
No problem! The `<>` was used as workaround for v4, but now we're the process of migrating to v5 and things are much better. I've created a codemod and a few fixture helpers to make the transition less painful for us. Do you think something like that would be useful for the community? I can create a repo with that stuff.

skidding:
> Do you think something like that would be useful for the community? I can create a repo with that stuff.

That would be amazing, please share if you have the time!

I've also concocted a codemod for a user with thousands of fixtures, but it's very hacky and only works with array fixtures (although that part can be easily adjusted). I can also share it with you if you're insterested.

mrm007:
Alright, I'll post the link here once I put the repo together and we can add more recipes there.

PR #1219, opened 2020-06-14
Options to enable HTTPS server mode for cosmos-server

### LINKED ISSUE #1218: Can I use https with own certs for Cosmos server?
Not found this feature in manual. I can switch port but seems there is no option to enable https for server. Or I missed something?

avin:
Fixes #1218

It can work with custom cert+key files or generate them automatically if there are no any provided.

Valid configs:
```
{
  "https": true
}
```
or
```
{
  "https": true,
  "httpsOptions": {
    "key": "cert/server.key",
    "cert": "cert/server.crt"
  }
}

### DIFF
--- a/packages/react-cosmos/src/config/createCosmosConfig.ts
+++ b/packages/react-cosmos/src/config/createCosmosConfig.ts
@@ -12,8 +12,7 @@ export function createCosmosConfig(
     rootDir,
     exportPath: getExportPath(cosmosConfigInput, rootDir),
     staticPath: getStaticPath(cosmosConfigInput, rootDir),
-    httpsKeyPath: getHttpsKeyPath(cosmosConfigInput, rootDir),
-    httpsCertPath: getHttpsCertPath(cosmosConfigInput, rootDir),
+    httpsOptions: getHttpsOptions(cosmosConfigInput, rootDir),
     publicUrl: getPublicUrl(cosmosConfigInput),
     fixturesDir: getFixturesDir(cosmosConfigInput),
     fixtureFileSuffix: getFixtureFileSuffix(cosmosConfigInput),
@@ -36,20 +35,20 @@ function getStaticPath(cosmosConfigInput: CosmosConfigInput, rootDir: string) {
   return staticPath && path.resolve(rootDir, staticPath);
 }
 
-function getHttpsKeyPath(
+function getHttpsOptions(
   cosmosConfigInput: CosmosConfigInput,
   rootDir: string
 ) {
-  const httpsKeyPath = cosmosConfigInput?.httpsOptions?.keyPath;
-  return httpsKeyPath && path.resolve(rootDir, httpsKeyPath);
-}
-
-function getHttpsCertPath(
-  cosmosConfigInput: CosmosConfigInput,
-  rootDir: string
-) {
-  const httpsCertPath = cosmosConfigInput.httpsOptions?.certPath;
-  return httpsCertPath && path.resolve(rootDir, httpsCertPath);
+  if (
+    cosmosConfigInput.httpsOptions?.keyPath &&
+    cosmosConfigInput.httpsOptions?.certPath
+  ) {
+    return {
+      keyPath: path.resolve(rootDir, cosmosConfigInput.httpsOptions.keyPath),
+      certPath: path.resolve(rootDir, cosmosConfigInput.httpsOptions.certPath),
+    };
+  }
+  return null;
 }
 
 function getPublicUrl({ publicUrl = '/' }: CosmosConfigInput) {
--- a/packages/react-cosmos/src/config/createCosmosConfig.ts
+++ b/packages/react-cosmos/src/config/createCosmosConfig.ts
@@ -12,6 +12,8 @@ export function createCosmosConfig(
     rootDir,
     exportPath: getExportPath(cosmosConfigInput, rootDir),
     staticPath: getStaticPath(cosmosConfigInput, rootDir),
+    httpsKeyPath: getHttpsKeyPath(cosmosConfigInput, rootDir),
+    httpsCertPath: getHttpsCertPath(cosmosConfigInput, rootDir),
     publicUrl: getPublicUrl(cosmosConfigInput),
     fixturesDir: getFixturesDir(cosmosConfigInput),
     fixtureFileSuffix: getFixtureFileSuffix(cosmosConfigInput),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/react-cosmos/src/config/shared.ts [hunk 1]
```diff
@@ -12,6 +12,12 @@ export type CosmosConfig = {
   // unspecified IPv4 address (0.0.0.0) otherwise.
   hostname: null | string;
   port: number;
+  https?: boolean;
+  httpsOptions?: {
+    key: string;
+    cert: string;
+    ca: string;
```
skidding, line 19: Is this used?

packages/react-cosmos/src/shared/devServer/httpServer.ts
```diff
@@ -1,17 +1,62 @@
...
+  if (httpsEnabled) {
+    let credentials: {
+      key: string;
+      cert: string;
+    };
+
+    if (httpsOptions?.key && httpsOptions?.cert) {
+      credentials = {
+        key: fs.readFileSync(httpsOptions.key, 'utf8'),
+        cert: fs.readFileSync(httpsOptions.cert, 'utf8'),
```
skidding, line 28: These are file paths, right? The user should be able to use relative paths like `mycert.crt` or `./mycert.crt`, like all the other paths are handled in the Cosmos config.

Here are two examples of how paths are handled currently, basically resolving them from `rootDir`:
https://github.com/react-cosmos/react-cosmos/blob/1a54f3cf37afd0440d05a7a06897994bdbf56712/packages/react-cosmos/src/config/createCosmosConfig.ts#L27-L35

I'm even thinking of naming the options `keyPath` and `certPath` for both consistency and clarity of their value.

packages/react-cosmos/src/config/shared.ts
avin: yes, it is needed for `httpServer.ts`

packages/react-cosmos/src/shared/devServer/httpServer.ts
avin: made changes

packages/react-cosmos/src/config/shared.ts
avin: `ca` is superfluous - fixed

avin:
made some changes, check this out

packages/react-cosmos/src/config/createCosmosConfig.ts
```diff
@@ -12,6 +12,8 @@ export function createCosmosConfig(
     rootDir,
     exportPath: getExportPath(cosmosConfigInput, rootDir),
     staticPath: getStaticPath(cosmosConfigInput, rootDir),
+    httpsKeyPath: getHttpsKeyPath(cosmosConfigInput, rootDir),
+    httpsCertPath: getHttpsCertPath(cosmosConfigInput, rootDir),
```
skidding, line 16: I would prefer this mirror the `CosmosConfigInput` shape to avoid confusion when working with the config. By this I mean nesting the `httpsOptions` here as well, especially since you can either have none or both, having just one doesn't make sense.

packages/react-cosmos/src/config/shared.ts
```diff
@@ -12,6 +12,13 @@ export type CosmosConfig = {
   // unspecified IPv4 address (0.0.0.0) otherwise.
   hostname: null | string;
   port: number;
+  https?: boolean;
+  httpsOptions?: {
+    keyPath?: string;
+    certPath?: string;
+  };
```
skidding, line 19: This is probably a pet peeve, but I prefer having all options _defined_ (no `?:`) in `CosmosConfig`. This means that `https` is always a boolean that defaults to `false`, `httpsOptions` can either be `null` or have a value, in which case `keyPath` and `certPath` will always a string. You can achieve this type guarantees by providing defaults in `createCosmosConfig.ts` option helpers, like the other options are handled.

packages/react-cosmos/src/config/shared.ts
```diff
@@ -12,6 +12,13 @@ export type CosmosConfig = {
...
+  httpsKeyPath?: string;
+  httpsCertPath?: string;
```
skidding, line 21: I don't think duplicating the types like this is needed. See my comment at `createCosmosConfig.ts:16`

packages/react-cosmos/src/config/createCosmosConfig.ts
```diff
@@ -34,6 +36,22 @@ function getStaticPath(cosmosConfigInput: CosmosConfigInput, rootDir: string) {
   return staticPath && path.resolve(rootDir, staticPath);
 }
 
+function getHttpsKeyPath(
```
skidding, line 39: This is the right approach, but I think you only need one option helper for `httpsOptions` as you can only have one or both nested options. If `httpsOptions` isn't found in the user config it should default to `null` to be consistent with the other optionals.

avin:
Fixed. Let's try again)

skidding:
@avin Can you please try `5.5.0-alpha.0` and let me know if HTTPS works as expected?

(Note that I cleaned up the styling in #1223, but the underlying code is the same)

I tried it and this is what I get in Chrome:

image

I expected to get a warning but it won't even let me skip and view the app anyway. Other browsers allow it, but Chrome says

> You cannot visit localhost right now because the website sent scrambled credentials that Google Chrome cannot process.

Again, does this match your expectation? Personally I don't know what the use-case for serving a local dev tool over HTTPS is, so I can't judge the outcome.

Thanks!

avin:
To skip alert screen for auto-generated certs you have to click "Advanced" 
image

works fine in Chrome and FF for me
image

image

> Personally I don't know what the use-case for serving a local dev tool over HTTPS is, so I can't judge the outcome.

You have to use HTTPS if
1) Debug payment forms (CC card autofill in chrome require HTTPS even on localhost)
2) You proxy localhost api to server that set HSTS headers - so chrome automatically redirect you to https and you cannot use simple http link :(

These are rare cases. But it so happened that I have to deal with both. So I have to use https for react-cosmos :)

skidding:
> These are rare cases. But it so happened that I have to deal with both. So I have to use https for react-cosmos :)

Good to know, thanks!

> To skip alert screen for auto-generated certs you have to click "Advanced"

I know what you mean, but I get a different screen, which doesn't have that "Open anyway" link:

image

I did find a way to overcome it, though, mentioned [here](https://support.google.com/chrome/thread/10551759?hl=en&msgid=31644104).

You literally need to type “thisisunsafe” in that tab (with no visible input) and the app suddenly loads 😅. Ever got this?

avin:
ah, thats a problem of a Chrome on mac :(
it is a normal behavor.

PR #1246, opened 2020-08-07
docs: updated nextjs instructions

frankandrobot:
As discussed in issue, added missing 'babel-loader' dep

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -494,7 +494,7 @@ module.exports = overrides.webpack(config, process.env.NODE_ENV);
 
 ## Next.js
 
-- Install `html-webpack-plugin` as a dev dependency.
+- Install `html-webpack-plugin` and `babel-loader` as a dev dependency.
 - Override the React Cosmos webpack config with `ProvidePlugin` to support JSX without importing React.
 - Set `staticPath` to `public` to load static assets inside React Cosmos.
 


docs/README.md
```diff
@@ -494,7 +494,7 @@ module.exports = overrides.webpack(config, process.env.NODE_ENV);
 
 ## Next.js
 
-- Install `html-webpack-plugin` as a dev dependency.
+- Install `html-webpack-plugin` and `babel-loader` as a dev dependency.
 - Override the React Cosmos webpack config with `ProvidePlugin` to support JSX without importing React.
```
frankandrobot, line 498: nit: in all my nextjs projects, we do *not* import React. I think this may be optional. Reword to the following?

```
- If your project is configured to support JSX without importing React, then override the `ProvidePlugin`.
```

skidding: Feel free to improve this part too, but I don't quite get the second part, which is "then override the `ProvidePlugin" – Do you think Next devs will know what to do with this info?

Initially Cosmos works with Next without providing any webpack config/override, but if you want ProvidePlugin, you need to create a webpack override file for Cosmos and manually add ProvidePlugin to the existing list of plugins. An example for this is included right below these instructions, so maybe it would be enough to add "(see example below)" at the end of this particular instruction.

I'm open to more suggestions. Thanks!

Issue #1272, opened 2020-10-29
"Error: Already listening" with create-react-app 4.0.0


AlexFlasch:
## What's wrong?

I've just started a project using CRA 4.0.0 and intended to use react-cosmos. Switching to using the `cosmos.config.json` recommended for CRA projects causes a React error overlay stating "Error: Already listening" with a stack trace that seems to point to React Refresh being involved?

This could be related to #1150 since CRA 4.0.0 has officially moved to React Refresh.

## Steps to reproduce

1.  Create a new CRA project with the base template (make sure to use the latest CRA version) `npx create-react-app cra-4-error`

2.  Add a `cosmos.config.json` with the following contents:
```
{
  "staticPath": "public",
  "watchDirs": ["src"],
  "webpack": {
    "configPath": "react-scripts/config/webpack.config"
  }
}
```
(the same as recommended in the Create React App section of the docs)

3.  Run `cosmos` and the error "Error: Already listening" should appear on the page.

## Screenshots

<img width="1680" alt="Screen Shot 2020-10-29 at 4 49 37 PM" src="https://user-images.githubusercontent.com/2762315/97636485-6a3d8d00-19f6-11eb-96e4-7cb660cebcd2.png">

### Server terminal output

no terminal error output

### Browser console output

<details>
<summary>output</summary>

```
Uncaught Error: Already listening
    at Module.ae (index.js:1)
    at Object.init (reactErrorOverlay.js:22)
    at Object../node_modules/react-cosmos/dist/plugins/webpack/client/errorOverlay/index.js (index.js:9)
    at __webpack_require__ (bootstrap:851)
    at fn (bootstrap:150)
    at Object../node_modules/react-cosmos/dist/plugins/webpack/client/index.js (index.js:15)
    at __webpack_require__ (bootstrap:851)
    at fn (bootstrap:150)
    at Object.<anonymous> (main.js:13)
    at __webpack_require__ (bootstrap:851)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.js:1

WebSocket connection to 'ws://localhost:5000/sockjs-node' failed: Connection closed before receiving a handshake response
```

</details>

### Cosmos version

react-cosmos@5.5.1

### If bug occurred on upgrade: Working Cosmos version

n/a

### Cosmos config

```
{
  "staticPath": "public",
  "watchDirs": ["src"],
  "webpack": {
[...]

PR #1273, opened 2020-11-01
Fix playground error when Webpack config uses `react-refresh-webpack-plugin`

AlexFlasch:
Fix for #1272 

After adding the `react-refresh` package to `react-cosmos-playground2`'s dev dependencies I am no longer receiving "Error: Already listening" when running my project using `react-scripts@4.0.0`'s Webpack config. It appears that `react-refresh` is a dependency of `@pmmmwh/react-refresh-webpack-plugin` and without it the new hot module reloading doesn't function properly.

I tried directly installing `react-refresh` as a dependency to my Cosmos project, but the playground still gives me the same error referenced in #1272. 

I understand this causes the project to take on an extra dependency just for users with a specific configuration, but I'm unsure of how else to solve the issue. If you have any suggestions or this sparks a different idea for how to solve it I'd be happy to re-implement it that way as well!

### DIFF
--- a/packages/react-cosmos-playground2/package.json
+++ b/packages/react-cosmos-playground2/package.json
@@ -16,6 +16,7 @@
     "react-dom": "17.0.1",
     "react-is": "^17.0.1",
     "react-plugin": "^2.0.0-alpha.8",
+    "react-refresh": "^0.9.0",
     "socket.io-client": "2.3.1",
     "styled-components": "^5.2.0"
   }
--- a/packages/react-cosmos-playground2/package.json
+++ b/packages/react-cosmos-playground2/package.json
@@ -16,7 +16,6 @@
     "react-dom": "17.0.1",
     "react-is": "^17.0.1",
     "react-plugin": "^2.0.0-alpha.8",
-    "react-refresh": "^0.9.0",
     "socket.io-client": "2.3.1",
     "styled-components": "^5.2.0"
   }
--- a/packages/react-cosmos/package.json
+++ b/packages/react-cosmos/package.json
@@ -24,6 +24,7 @@
     "react-cosmos-plugin": "^5.5.0",
     "react-cosmos-shared2": "^5.5.1",
     "react-error-overlay": "^6.0.8",
+    "react-refresh": "^0.9.0",
     "regenerator-runtime": "^0.13.7",
     "resolve-from": "^5.0.0",
     "slash": "^3.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10565,6 +10565,11 @@ react-plugin@^2.0.0-alpha.8:
     lodash "^4.17.11"
     ui-plugin "^2.0.0-alpha.4"
 
+react-refresh@^0.9.0:
+  version "0.9.0"
+  resolved "https://registry.yarnpkg.com/react-refresh/-/react-refresh-0.9.0.tgz#71863337adc3e5c2f8a6bfddd12ae3bfe32aafbf"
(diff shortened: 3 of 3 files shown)


skidding:
This is surprising, because a dev dependency to `react-cosmos-playground2` wouldn't normally affect Cosmos users in any way. Those (dev)dependencies won't be installed when you install Cosmos, they are only used when react-cosmos-playground2 is built. And since there is no reference to react-refresh in the react-cosmos-playground2 package, I can't imagine how this change affects anything.

Can you please explain how you tested before and after this change to reach this conclusion? Thanks!

AlexFlasch:
After installing the dev dependency I `yarn link`'d my fork of react-cosmos into my own project.

However, you're definitely right about dev dependencies not carrying through to end users, and likely this just worked for me because I was using a local copy of react-cosmos linked to my project, and thus still had the dev dependencies installed.

In that case I could put it into the normal list of dependencies, but I'm unsure of how that would affect bundling or anything like that for `cosmos-export`.

Do you have any ideas for how to better test this change? I don't believe unit or e2e tests will really catch this. (Also I can't get cypress working from a WSL2 bash terminal on Windows... 😕 )

skidding:
> After installing the dev dependency I `yarn link`'d my fork of react-cosmos into my own project.

Unfortunately yarn/npm link doesn't work well for testing a complex package like react-cosmos, which looks for user dependencies in the same node_modules directory as it is installed. The method I've been using over the years is to `cp -R` the dist folder of the cosmos fork into the project I'm testing it in. You can use a small utility to do this automatically whenever cosmos gets rebuilt for faster iteration.

I'm sorry I don't have more time to get involved with this and contribute meaningfully. For now I'll say that it doesn't make sense to me to add `react-refresh` to the `react-cosmos-playground2` package. The Playground is the Cosmos UI, which isn't coupled to any compiler technology and shouldn't be aware of anything webpack related. If the `react-refresh` dependency is indeed the solution, it should be added to `react-cosmos` package, which is the glue that brings everything together, and where webpack is configured to work with the user's pipeline.

Let me know if you make more progress with this. Thanks!

skidding:
Looked more closely into this, and to be honest I can't imagine how this would fix the issue. The "Error: Already listening" stems from the fact that Cosmos includes a copy of CRA's error overlay by default, and with the latest changes in CRA 4 the error overlays ends up being called twice.

A potential solution is to allow Cosmos users to disable the built in error overlay [referenced here](https://github.com/react-cosmos/react-cosmos/blob/e92f1ecbce529af4b4706172b7e5aae15823ff1e/packages/react-cosmos/src/plugins/webpack/client/index.ts#L8-L28).

But even after you get past that error, you'll see another error in your browser console
> WebSocket connection to 'ws://localhost:5000/sockjs-node' failed: Connection closed before receiving a handshake response

This means Fast Refresh won't work inside Cosmos, even if you left it enabled and got past the "Already listening" error. So for now disabling Fast Refresh like you suggested is the way to go.

The bigger fish to fry, is making Cosmos play well with Fast Refresh, something I've researched [here](https://github.com/react-cosmos/react-cosmos/issues/1272#issuecomment-733091647).

Let me know if you have any thoughts.

I think this PR can be closed, but it might be useful to add a [Troubleshooting](https://github.com/react-cosmos/react-cosmos/tree/e92f1ecbce529af4b4706172b7e5aae15823ff1e/docs#troubleshooting) entry for the "Already listening" error with a link to your cross-env solution until we figure out Fast Refresh. If you can make a PR for that I'd be grateful.

AlexFlasch:
I'd be happy to close this PR as well. As mentioned in #1272 the approach of adding `react-refresh` as a dependency  unfortunately will not suffice to fix the issue.

I will try to open a separate PR in the next day or two to add an entry for the `cross-env` fix for CRA 4 users.

PR #1280, opened 2020-12-10
Make the output.path undefined

### LINKED ISSUE #1279: clean-webpack-plugin breaks react-cosmos
### What's wrong?

`devServer.ts` forces the webpack output path to `/`:

https://github.com/react-cosmos/react-cosmos/blob/52ef3d16cbad56f7a1637e94adfd791cde8899fa/packages/react-cosmos/src/plugins/webpack/webpackConfig/devServer.ts#L69-L71

`clean-webpack-plugin` passes this output path to `delSync`:

https://github.com/johnagan/clean-webpack-plugin/blob/021d2b68efd18fbd99e02c169d9e32300f15b4f4/src/clean-webpack-plugin.ts#L310-L317

But that can result in errors when it tries to read paths it doesn't have access to. Under WSL I get the following error when running `yarn cosmos`:

```log
✖ ｢wdm｣: Error: EACCES: permission denied, scandir '/c/$Recycle.Bin/S-1-5-18'
    at Object.readdirSync (fs.js:854:3)
```

As a result react-cosmos just sits there waiting for the renderer without any apparent error in the browser.

### Steps to reproduce

1. Find any project using `clean-webpack-plugin`
2. Add `react-cosmos`
3. Run `yarn cosmos`

### Workaround

I worked around this by adding the following `webpack.override.js` to my setup:

```js
module.exports = (webpackConfig) => {
  // Drop clean-webpack-plugin because react-cosmos sets the compiler
  // output path to '/' and that will cause it to error when it does
  // del.sync with `cwd` set to `compiler.options.output.path` as it tries
  // to read from directories it doesn't have permission to read from.
  //
  // We shouldn't need clean-webpack-plugin for cosmos builds anyway.
  const plugins = webpackConfig.plugins.filter(
    (plugin) =>
      plugin.constructor && plugin.constructor.name !== 'CleanWebpackPlugin'
  );
  return { ...webpackConfig, plugins };
};
```

birtles:
Setting the output path to '/' causes plugins like clean-webpack-plugin to
scan the entire filesystem, reporting errors when they encounter files
they cannot read.

Other tools like create-react-app set simply set the output path to undefined:

https://github.com/facebook/create-react-app/blob/c1714042b60ff48cb08407bec236c0a92d26773b/packages/react-scripts/config/webpack.config.js#L157-L158

Fixes #1279

### DIFF
--- a/packages/react-cosmos/src/plugins/webpack/webpackConfig/__tests__/customDevConfig.ts
+++ b/packages/react-cosmos/src/plugins/webpack/webpackConfig/__tests__/customDevConfig.ts
@@ -97,7 +97,6 @@ it('create output', async () => {
   expect(output).toEqual(
     expect.objectContaining({
       filename: '[name].js',
-      path: '/',
       publicPath: '/',
     })
   );
--- a/packages/react-cosmos/src/plugins/webpack/webpackConfig/__tests__/defaultDevConfig.ts
+++ b/packages/react-cosmos/src/plugins/webpack/webpackConfig/__tests__/defaultDevConfig.ts
@@ -43,7 +43,6 @@ it('create output', async () => {
   expect(output).toEqual(
     expect.objectContaining({
       filename: '[name].js',
-      path: '/',
       publicPath: '/',
     })
   );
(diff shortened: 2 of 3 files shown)


skidding:
@birtles Thanks for the prompt PR!

Published `react-cosmos@5.5.3` with your fix. Please try out out and let me know if everything works as expected.

birtles:
> @birtles Thanks for the prompt PR!
> 
> Published `react-cosmos@5.5.3` with your fix. Please try out out and let me know if everything works as expected.

Thank you! I just tried it out and it works great. Thanks again.

PR #1290, opened 2021-02-01
Add setting to override webpack output.filename template

pavelivanov:
**Problem:**

Production build generates `main.js` file. This file doesn't have any hash in name to invalidate cache. Webpack has an option for this: https://webpack.js.org/configuration/output/#template-strings.

**Solution:**

Add "outputFilename" property to cosmosConfig to allow users to override `output.filename` value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea
+
 node_modules
 npm-debug.log
 yarn-error.log
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-.idea
-
 node_modules
 npm-debug.log
 yarn-error.log
--- a/docs/README.md
+++ b/docs/README.md
@@ -180,7 +180,7 @@ module.exports = (webpackConfig, env) => {
 };
 ```
 
-### Output Filename
+### Output filename
 
 Cosmos overrides `output.filename` value in webpack final config to `[name].js` by default. Some users are faced with caching `main.js` file. Use the `webpack.includeHashInOutputFilename` setting to change filename template to `[name].[contenthash].js`.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -182,7 +182,7 @@ module.exports = (webpackConfig, env) => {
 
 ### Output filename
 
-Cosmos overrides `output.filename` value in webpack final config to `[name].js` by default. Some users are faced with caching `main.js` file. Use the `webpack.includeHashInOutputFilename` setting to change filename template to `[name].[contenthash].js`.
+Cosmos overrides `output.filename` value in webpack final config to `[name].js` by default. Some users are faced with caching `main.js` file when building app with `cosmos-export`. Use the `webpack.includeHashInOutputFilename` setting to change filename template to `[name].[contenthash].js`.
 
 ```json
 {
--- a/docs/README.md
+++ b/docs/README.md
@@ -180,6 +180,18 @@ module.exports = (webpackConfig, env) => {
 };
 ```
 
+### Output Filename
+
+Cosmos overrides `output.filename` value in webpack final config to `[name].js` by default. Some users are faced with caching `main.js` file. Use the `webpack.includeHashInOutputFilename` setting to change filename template to `[name].[contenthash].js`.
+
+```json
+{
+  "webpack": {
+    "includeHashInOutputFilename": true
+  }
+}
+```
+
 ## Fixtures
 
 Fixture files contain a default export, which can be a React Component or any React Node.
@@ -539,6 +551,10 @@ This is a `.babelrc` example for Next.js:
 #### "localhost:3001/\_\_get-internal-source..." 404s?
 
 - [Try changing your webpack `devtool` to something like `cheap-module-source-map`](https://github.com/react-cosmos/react-cosmos/issues/1045#issuecomment-535150617).
+
+#### main.js file is cached?
+
+- [Set `includeHashInOutputFilename` to `true`](https://github.com/react-cosmos/react-cosmos/tree/master/docs#output-filename).
 
 #### Serving a static export from a nested path?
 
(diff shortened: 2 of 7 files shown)


skidding:
The only thing I would change is to move the `outputFilename` option into the webpack-related Cosmos config. I'll explain.

The _core_ Cosmos config is the one you already found.

https://github.com/react-cosmos/react-cosmos/blob/8c85549f873dfedea017c5c0718e94464e2fe227/packages/react-cosmos/src/config/shared.ts#L8-L39

But it doesn't have webpack related options. This choice was made to treat the webpack integration as a plugin. Only source code from `packages/react-cosmos/src/plugins/webpack` knows about webpack. The rest of the codebase is agnostic.

But of course the user config has all the options in one JSON. So in order to define and access plugin-related Cosmos config, you must create a new type for a specific plugin key (`webpack` in this case), and a plugin-related function that accesses this part of the config and returns the specific type of options. 

This already exists. So basically you need to move the new option from the core Cosmos config to this file.

https://github.com/react-cosmos/react-cosmos/blob/d88061a943457b967daf84a503e73283182f1521/packages/react-cosmos/src/plugins/webpack/cosmosConfig/webpack.ts#L5-L23

This has two implications. First, in the user JSON config, `outputFilename` will be nested under the `webpack` key.

Second, you need to call `createWebpackCosmosConfig` where you want to access `outputFilename`. Here is an example from where the `webpack.overridePath` option is used.

https://github.com/react-cosmos/react-cosmos/blob/dbe83a74f53de7ae3dd61f822aec56e673f725b9/packages/react-cosmos/src/plugins/webpack/webpackConfig/shared.ts#L35

pavelivanov:
@skidding thank you for detailed explanation. I put `outputFilename` there in same way as `publicUrl` and `exportPath` do. First of all I thought that it would be nice to allow users to override whole output (users who do this know what they are doing). By docs:
```
module.exports = (webpackConfig, env) => {
  return merge(webpackConfig, {
    output: {
      filename: '[name].[fullhash:8].js',
      chunkFilename: '[name].[chunkhash:8].chunk.js',
      // ...
    },
  })
}
```
But for now it is a blackbox how webpack config be overrided inside cosmic :(

skidding:
> But for now it is a blackbox how webpack config be overrided inside cosmic :(

Indeed it's a tricky business. In principle Cosmos uses all your webpack config but replaces `entry` and `output`. The entry is the list of fixtures generated by Cosmos, and the output is meant to consumed by Cosmos so again isn't necessarily something in the user space. 

But having hashed file names for exports Cosmos is a real need, so I think this change will be valuable for many people. Aside from this I've never heard anyone wanting to customize anything else in the webpack output of the Cosmos build.

Let me know if you have trouble following my review, I think you're very close to finishing this :).

pavelivanov:
@skidding I decided to replace `outputFilename: string` with `includeHashInOutputFilename: boolean` because webpack `output.filename` type is `string | ((pathData: PathData, assetInfo?: AssetInfo) => string)` and string variable could lead to questions "why can't I add function?" :) 

Please check the docs and the schema I updated. Is this done correctly?

docs/README.md
```diff
@@ -180,6 +180,18 @@ module.exports = (webpackConfig, env) => {
 };
 ```
 
+### Output Filename
```
skidding, line 183: Make `filename` lowercase please? The rest of the document doesn't capitalise words.

docs/README.md
```diff
@@ -180,6 +180,18 @@ module.exports = (webpackConfig, env) => {
...
+
+Cosmos overrides `output.filename` value in webpack final config to `[name].js` by default. Some users are faced with caching `main.js` file. Use the `webpack.includeHashInOutputFilename` setting to change filename template to `[name].[contenthash].js`.
```
skidding, line 185: This is good. I would add that users are faced with this caching issue when dealing with exports. In dev mod the filename is irrelevant, right?

skidding (changes requested):
Good idea with `includeHashInOutputFilename`, I like your reasoning. Also thanks for going the extra mile with the docs and config schema! Minor comments left.

.gitignore
```diff
@@ -1,3 +1,5 @@
+.idea
```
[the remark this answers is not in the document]
pavelivanov: oooops, sorry :) Will remove

docs/README.md
pavelivanov: > ...with this caching issue when dealing with exports

Can you explain please?

> In dev mod the filename is irrelevant, right?

Yes.

.gitignore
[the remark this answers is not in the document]
pavelivanov: https://github.com/pavelivanov/react-cosmos/commit/9da6a876f5d9d283d5103a50e4f714ac8484301c

docs/README.md
pavelivanov: https://github.com/pavelivanov/react-cosmos/commit/9da6a876f5d9d283d5103a50e4f714ac8484301c

docs/README.md
skidding: > Can you explain please?

I mean that this is an option specifically for the `cosmos-export` API. Not everybody generates static Cosmos exports, so I want to make sure people understand where this option is relevant.

pavelivanov: Ok, I mentioned about this in the readme: https://github.com/pavelivanov/react-cosmos/commit/6d919c6d760c05c575e243c9549da11aa3daa44c

skidding (approved):

skidding:
Please try out `react-cosmos@5.6.0-alpha.2` and let me know if this change works as expected.

skidding:
> @skidding checked, it works 😉

Awesome. Thanks a lot for taking care of this!

If you want to check another box, you can update https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/cosmos-config.json with the schema already updated in this repo. This will cause smart editors to update their built in autocomplete with the option you created :).

PR #1306, opened 2021-03-02
Typo fix

amir-hadzic:
Just a small typo fix, noticed this when the page was shared in Slack:

image


skidding:
@amir-hadzic Thanks!

PR #1313, opened 2021-03-15
Lazy fixture importing

skidding:
- [ ] Update fixture list with fixture names as fixture modules are imported.
- [ ] Debug what happens on HMR in lazy mode.
- [ ] Import fixtures statically in static exports.
- [ ] Put dynamic importing under `lazy` config option.
- [ ] Dynamically import decorators as well.
- [x] ~~Decide on best name for `FixtureList`.~~ It's fine.
- [x] Make `name` optional in FixtureId.
  - [ ] Fix selected fixture name in `packages/react-cosmos-playground2/src/plugins/Root/RendererHeader.tsx`.
  - [ ] Refactor `packages/react-cosmos-playground2/src/plugins/FixtureTree/FixtureTree/MultiFixtureButton.tsx`.
  - [ ] Test new case in `packages/react-cosmos-playground2/src/plugins/RendererCore/isValidFixtureSelected.ts`.

## Questions

- [ ] As the user clicks on fixtures they are imported and thus "discovered" (names of multi fixtures are revealed). On HMR, how do we know which fixtures to keep discovered and which to invalidate? Maybe it's OK to invalidate all fixtures aside from the selected one. But the selected fixture also gets invalidated and re-imported, which a) causes to UI to flicker and b) might interfere with HMR.
  - [ ] What if fixture "discovery" is persisted on the backend, and the discovered fixtures are imported non-lazily in the webpack loader until the server is restarted?


skidding:
Superseded by #1443

PR #1321, opened 2021-03-25
fix userDepsTemplate slash for windows

garymacpherson:
Example output of a `cosmos.userdeps.js` file before the fix on Windows.

```
import fixture0 from '.\components\Button.fixture.jsx';

export const rendererConfig = {
  "port": 5000
};

export const fixtures = {
  'src/components/Button.fixture.jsx': fixture0
};

export const decorators = {

};
```

Example after fix

```
import fixture0 from './components/Button.fixture.jsx';

export const rendererConfig = {
  "port": 5000
};

export const fixtures = {
  'src/components/Button.fixture.jsx': fixture0
};

export const decorators = {

};
```

Note the `import fixture0 from '.\components\Button.fixture.jsx';` line has relative pathing that is invalid with Windows. This either needs to be escaped or changed to forward slashes. As the slash package is already here and being used in this file, best to just use that to change to forward slashes in the Windows environment.

No negative impact on OSX/Linux dev environments.

### DIFF
--- a/packages/react-cosmos/src/shared/userDeps/userDepsTemplate.ts
+++ b/packages/react-cosmos/src/shared/userDeps/userDepsTemplate.ts
@@ -70,6 +70,6 @@ function createImportMap(
 
 function resolveImportPath(filePath: string, relativeToDir: string | null) {
   return relativeToDir
-    ? `.${path.sep}${path.relative(relativeToDir, filePath)}`
+    ? slash(`.${path.sep}${path.relative(relativeToDir, filePath)}`)
     : filePath;
 }


skidding:
> This either needs to be escaped

Out of curiosity, what would a relative path look like natively on Windows?

garymacpherson:
> > This either needs to be escaped
> 
> Out of curiosity, what would a relative path look like natively on Windows?

Windows does support back and forward slash, the issue is when backslashes are emitted they need to be escaped to be a valid javascript file.

eg. when using the un-fixed file you get the following error

`> src/cosmos.userdeps.js:8:21: error: Could not resolve ".componentsButton.fixture.jsx" (mark it as external to exclude it from the bundle)` as effectively the unescaped \ is being dropped entirely.

skidding:
@garymacpherson Thanks for this fix and for the explanation!

I've published your patch under `react-cosmos@5.6.2`. Please try it out and let me know if it works as expected.

Issue #1266, opened 2020-09-23
Resolve yarn 2 warnings


davidroeca:
### What's wrong?

When installing `react-cosmos` from yarn 2, the following warnings are raised:

```shell
$ yarn --version
2.1.1
$ yarn install
➤ YN0002: │ react-cosmos-shared2@npm:5.5.0 doesn't provide react@^0.14.8 || ^15.0.1 || ^16.0.0 requested by react-element-to-jsx-string@npm:14.3.1
➤ YN0002: │ react-cosmos-shared2@npm:5.5.0 doesn't provide react-dom@^0.14.8 || ^15.0.1 || ^16.0.0 requested by react-element-to-jsx-string@npm:14.3.1
➤ YN0002: │ react-cosmos@npm:5.5.0 doesn't provide webpack@^4.0.0 requested by webpack-dev-middleware@npm:3.7.2
```

### Steps to reproduce

1. [Install yarn](https://yarnpkg.com/getting-started/install#per-project-install)
2. `yarn init` a new project
3. Within that new project `yarn set version latest`
4. Experience the warnings

### Screenshots

None necessary.

### User info

~- Server terminal output~
~- Browser console output~
- Cosmos version: 5.5.0
~- If bug occurred on upgrade: Working Cosmos version~
~- Cosmos config~
~- If applicable: webpack version~
~- If applicable: webpack config~

### Additional context

In order to resolve these warnings, `react-cosmos` and `react-cosmos-shared2` should list out these as either `peerDependencies` or `dependencies`.

Some more information on behavior of the warning: https://yarnpkg.com/advanced/error-codes#yn0002---missing_peer_dependency

skidding:
Cosmos doesn't put react/react-dom into `dependencies` to avoid duplicate React copies. It used to have them in `peerDependencies` for a while until it got annoying because it produced install warnings when people were using slightly different versions that were still compatible, and also when – and I'm not 100% about this – your codebase relied on another package to install these peer dependencies. So for example you use Create React App, which comes with React and webpack embedded, so your project doesn't have them listed as dependencies. Then you install Cosmos and it complains that it can't find those peer dependencies, even tho they are installed by CRA indirectly. This scenario needs to be validated, but I'm more than happy to get rid of these warnings if we can do it without introducing other warnings or bigger issues.

davidroeca:
[...]

PR #1337, opened 2021-07-04
Restore optional peer dependencies (+ snowpack docs update)

### LINKED ISSUE #1320: "[404] Not Found (/dist/cosmos.userdeps.proxy.js) " with snowpack v3.1.2
### What's wrong?
I've just cloned https://github.com/react-cosmos/react-cosmos-snowpack and upgraded snowpack to latest (3.1.2). 
`yarn start` / `yarn cosmos` runs without error, but `localhost:5000` stuck at "Waiting for renderer...", then snowpack logs:
 ```
[17:33:41] [snowpack] File changed: src/cosmos.userdeps.js
[17:33:48] [snowpack] [404] Not Found (/dist/cosmos.userdeps.proxy.js) (x2)
```

It runs fine on snowpack v3.0.13.
### Steps to reproduce
1. Clone https://github.com/react-cosmos/react-cosmos-snowpack
2. Replace `"snowpack": "^3.0.0"` with `"snowpack": "^3.1.2"` in `package.json`
3. `yarn` && `yarn start`
4. `yarn cosmos`
5. open `localhost:5000` and page keeps showing "Waiting for renderer..."

### User info

- Server terminal output

<details>
  <summary>yarn start</summary>

  ```
yarn run v1.22.10
$ snowpack dev
[17:33:38] [snowpack] Server started in 12ms.
[17:33:38] [snowpack] Local: http://localhost:8080
[17:33:38] [snowpack] Network: http://192.168.10.167:8080
[17:33:41] [snowpack] File changed: src/cosmos.userdeps.js
[17:33:48] [snowpack] [404] Not Found (/dist/cosmos.userdeps.proxy.js) (x2)
```
</details>

<details>
  <summary>yarn cosmos</summary>

  ```
yarn run v1.22.10
$ /home/orcaxs/dev/react-cosmos-snowpack/node_modules/.bin/cosmos
[Cosmos] Using cosmos config found at cosmos.config.json
[Cosmos] Serving static files from public
[Cosmos] See you at http://localhost:5000
[Cosmos] Generated src/cosmos.userdeps.js
```

</details>

- Browser console output
`index.cosmos.js:7 GET http://localhost:8080/dist/cosmos.userdeps.proxy.js net::ERR_ABORTED 404 (Not Found)`
- Cosmos version
Seems both `5.6.0-alpha.6` and `5.6.1` have the same issue
- Cosmos config
```json
{
  "staticPath": "public",
  "userDepsFilePath": "src/cosmos.userdeps.js",
  "experimentalRendererUrl": "http://localhost:8080/_renderer.html"
}
```

nderscore:
## Issue

Based on [this issue conversation](https://github.com/react-cosmos/react-cosmos/issues/1266#issuecomment-698842147), peer dependencies used to exist in this project, but were removed in order to avoid warning messages that appeared for users installing it in projects where `react` (or `webpack`) wasn't a direct dependency, such as CRA.

This is causing compatibility issues with `snowpack`'s [esinstall](https://github.com/snowpackjs/snowpack/tree/main/esinstall) tool, which uses rollup to transform all modules in the project into esmodules. I haven't dug too deep into the _exact_ reason, but when it encounters an undeclared dependency it seems to automatically bundle a copy of it into the module that requested it.

The lack of peer dependencies also causes `yarn` to throw warnings when installing the package.

## Resolution

This PR restores the correct peer dependencies back to the `react-cosmos` and `react-cosmos-shared2` packages. They're marked as optional inside `peerDependenciesMeta` in order to avoid the warning scenario described earlier.

I also updated the instructions for installing cosmos in a snowpack project, to bring it up to date with the latest versions of snowpack.

This resolves #1320 and #1266.

## Notes / Open Questions

I chose `react/react-dom >= 16.8` and `webpack >= 4` in order to loosely match the [requirements from the docs](https://github.com/react-cosmos/react-cosmos/tree/main/docs#requirements), but maybe it would be better to set a cap, at least on the webpack version, to avoid issues arising from them releasing a new major version?

## Todo

The [snowpack demo project](https://github.com/react-cosmos/react-cosmos-snowpack) should probably be updated based on the changes to the install instructions too.

### DIFF
--- a/docs/snowpack.md
+++ b/docs/snowpack.md
@@ -35,7 +35,7 @@ Create `index.cosmos.tsx` next to your existing index module.
 
 ```jsx
 import { mountDomRenderer } from 'react-cosmos/dom';
-import { decorators, fixtures, rendererConfig } from './cosmos.userdeps';
+import { decorators, fixtures, rendererConfig } from './cosmos.userdeps.js';
 
 mountDomRenderer({ rendererConfig, decorators, fixtures });
 
@@ -67,6 +67,16 @@ routes: [
   { src: '/_renderer.html', dest: '/index.cosmos.html' },
   // Other routes...
 ],
+```
+
+Enable the `polyfillNode` flag in the `packageOptions` of your Snowpack config.
+
+```js
+// snowpack.config.js
+packageOptions: {
+  polyfillNode: true,
+  // Other package options...
+},
 ```
 
 ### 4. Start React Cosmos with Snowpack
--- a/packages/react-cosmos-shared2/package.json
+++ b/packages/react-cosmos-shared2/package.json
@@ -10,5 +10,13 @@
     "react-element-to-jsx-string": "^14.3.2",
     "react-is": "^17.0.2",
     "socket.io-client": "2.2.0"
+  },
+  "peerDependencies": {
+    "react": ">= 16.8"
+  },
+  "peerDependenciesMeta": {
+    "react": {
+      "optional": true
+    }
   }
 }
--- a/packages/react-cosmos/package.json
+++ b/packages/react-cosmos/package.json
@@ -33,6 +33,22 @@
     "webpack-dev-middleware": "^4.3.0",
     "yargs": "^16.2.0"
   },
+  "peerDependencies": {
+    "react": ">= 16.8",
+    "react-dom": ">= 16.8",
+    "webpack": ">= 4.0"
+  },
+  "peerDependenciesMeta": {
+    "react": {
+      "optional": true
+    },
+    "react-dom": {
+      "optional": true
+    },
+    "webpack": {
+      "optional": true
+    }
+  },
   "bin": {
     "cosmos": "./bin/cosmos.js",
     "cosmos-export": "./bin/cosmos-export.js",


nderscore:
I tried updating the example project (`example/`) which is being tested by cypress to the latest snowpack, but so far I'm stuck on a different bug than the one I was originally targeting.  

In theory, all that should be needed is to set `workspaceRoot: '../'` in the snowpack config, but I'm running into what I think is [this](https://github.com/snowpackjs/snowpack/issues/3119)/[this](https://github.com/snowpackjs/snowpack/issues/3405) bug where symlinked dependencies aren't being correctly processed. I took a shot in the dark and tried adding as [`'exports'` field](https://nodejs.org/api/packages.html#packages_subpath_exports) to each of the packages in a variety of different ways, but it didn't resolve the issue.

The peer dependencies change still fixes the issues seen when using the package externally / outside of the react-cosmos monorepo. 😕

skidding:
@nderscore Thoughtful contribution! Sorry for taking so long to reply.

The peer dependencies change sounds great. Had no idea `peerDependenciesMeta` existed and it seems like this projects needs it.

The thing I'm uncertain about is the need for `polyfillNode` as Cosmos shouldn't try to import Node modules on the browser. I'd like us to get to the bottom of this before recommending Snowpack users to polyfill Node, because if Cosmos _does_ leak some Node imports into browser code we should fix that. Can we tell what's the problematic module(s)?

Hopefully the Snowpack team addresses the symlink issues. It's a shame latest Snowpack doesn't play well with Cosmos. I'm also using Snowpack with Cosmos in an important project. 

> The snowpack demo project should probably be updated based on the changes to the install instructions too.

I know things are still up in the air, but I would gladly make you a maintainer on the Snowpack demo project so you can update it freely. Would you like me to?

nderscore:
> The thing I'm uncertain about is the need for `polyfillNode` as Cosmos shouldn't try to import Node modules on the browser. I'd like us to get to the bottom of this before recommending Snowpack users to polyfill Node, because if Cosmos _does_ leak some Node imports into browser code we should fix that. Can we tell what's the problematic module(s)?

The cause of the `polyfillNode` requirement is this inner dependency of the old version of `socket.io-client` being used:
```
└── react-cosmos-shared2/FixtureLoader@5.6.3
  └── socket.io-client@2.2.0
    └── engine.io-client@3.3.2
      └── xmlhttprequest-ssl@1.5.5
```
This version of `xmlhttprequest-ssl` includes references to node.js modules (http/fs/etc)

I tried out upgrading `socket.io` + `socket.io-client` to the latest version (v4) across all cosmos packages ([branch here](https://github.com/nderscore/react-cosmos/tree/optional-peer-deps-no-polyfill-node)) and this fixes the build in snowpack without needing to enable `polyfillNode`.

As described in [DEBT.md](https://github.com/react-cosmos/react-cosmos/blob/main/DEBT.md#pinned-dependencies), this upgrade _would_ be problematic for IE11 support. That said, maybe it's not a bad time to be considering dropping IE11 support soon, now that even Microsoft is dropping support for it on their web applications next month? Snowpack probably isn't worth breaking things in IE11 before the final end-of-life date though. 😕 

An alternative solution to avoid upgrading socket.io might be to pin `xmlhttpreqeust-ssl` to `2.0.0` using yarn resolutions in the snowpack project, but I'm not sure if it will break compatibility with the version of `socket.io-client` being used and its definitly a hackier fix than `polyfillNode`.

> I know things are still up in the air, but I would gladly make you a maintainer on the Snowpack demo project so you can update it freely. Would you like me to?

I'd be happy to submit an equivalent PR there after we settle on a solution!

skidding:
Oh man, that IE11 support is really holding us back. I'm open to upgrading `socket.io-client`, dropping support for IE11 and releasing Cosmos 6. I think it's sensitive to do this in a major version. Was already thinking that it would be fair to bump the major version if we add peer dependencies. This way IE11 users can still use Cosmos 5. And if down the road we decide we still want IE11 support in Cosmos 6, we could isolate the WebSockets part from core functionality, as the sockets are only used in advanced features and the React Native integration.

skidding:
@nderscore PnP compatibility should be fixed in v6.

- Please check out the [migration guide](https://github.com/react-cosmos/react-cosmos/blob/main/docs/MIGRATION_V6.md) if you want to try it out. 
- Also updated the [pnp-example](https://github.com/react-cosmos/pnp-example) repo and created a [GitHub Actions workflow](https://github.com/react-cosmos/react-cosmos/blob/6aefc11752c44ad442dfeccd47858f59aa7cf605/.github/workflows/test-pnp.yml) that makes sure Cosmos still works well in that repo after every commit.
- In case I missed something let me know and we'll strengthen the example repo and automated workflow to prevent PnP issues from cropping up again in the future.

Issue #1353, opened 2021-10-17
Typescript compilation errors trigger infinite reloads


messman:
### What's wrong?

If strict Typescript checks are on while editing code, any TypeScript compilation errors will cause a fixture to reload indefinitely until compilation errors are fixed or until the browser tab is closed. The infinite reloading dramatically increases CPU usage by the browser and may cause crashes.

### Steps to reproduce

[See this test repo](https://github.com/messman/cosmos-reload-example). Instructions, as well as a discussion of the problem and research, are in the README.

Essentially:
- Set up a project for React Cosmos that uses strict TypeScript type checking.
- While viewing React Cosmos fixtures in a browser, cause a TypeScript compilation error in a file and save, triggering HMR.
- Watch as the test fixture starts reloading until stopped.

### User info

See the test repo for specifics on versions.
- React Cosmos 5.6.4
- Webpack 5.58.2
- TS Loader 9.2.6
- TypeScript 4.4.4

### Additional context

As discussed previously in #1315, I use strict type checking while working in React Cosmos. After working through performance issues, this is the only sore spot. The test repo linked above includes my research in tracking down the cause of the reloads to `@skidding/webpack-hot-middleware` ([webpack-contrib/webpack-hot-middleware](https://github.com/webpack-contrib/webpack-hot-middleware)), specifically the `reload` option that is hardcoded to `true`. This causes React Cosmos to operate more like `dev-server` than `only-dev-server` for HMR, reloading on errors and when there are no module updates.

I'll often fire up React Cosmos and then edit files to see live changes. Along the way, I'll commonly reorganize multiple files and end up creating many failed webpack builds from TypeScript compilation errors. After five minutes, I'll realize there's an issue with React Cosmos because my MacBook's fan will start getting louder and louder. I'll typically need to close the browser tab in order to calm the CPU down. 

In `plugins/webpack/webpackConfig/devServer.js`:

```js
function getHotMiddlewareEntry() {
  var clientPath = require.resolve('@skidding/webpack-hot-middleware/client');

  return "".concat(clientPath, "?reload=true&overlay=false");
}
```

Changing `reload=true` to `reload=false` or removing it entirely will stop the frame from reloading when a TypeScript compilation error occurs.

As the reloads occur, this HMR logging is printed in the browser console:

[...]

PR #1358, opened 2021-12-02
Create 'disableReload' config option to handle infinite reload case

messman:
From #1353:

> For a full solution, I would suggest the addition of an option in `cosmos.config.json` for controlling the `reload` behavior of HMR. I'd like to know what others think. Cheers.

> In the meantime I like your idea to add a Cosmos config option to disable the reload flag. Would you be willing to make this PR?

This PR adds a new Cosmos config option `disableReload` that, when `true`, sets the internal webpack-hot-middleware `reload` option to `false`. This, in turn, causes the hot reload code to _not_ do a `window.location.reload` when an error is encountered (no modules to update; error while updating modules; etc) by webpack during hot reload.

This config option is not required and defaults to `false`, and as such is not a breaking change.

The option is named `disableReload` for brevity and to match the naming convention for an existing option, `disablePlugins`. "disableReload" is a potentially confusing name, as one may see it as controlling the overall hot reload experience; however, this use of the term "reload" to only apply to a _window location reload_ is consistent with the naming of the `reload` option in webpack-hot-middleware. `reload` controls unexpected update behaviors, not expected (happy-path) "update" ones.

Changes include:
- Updating cosmos config schema (will need to be updated on schemastore)
- Adding TypeScript property
- Updating function responsible for getting the hot middleware configuration string

Tested and fixes issue in #1353 as first described in its [test repo](https://github.com/messman/cosmos-reload-example).

### DIFF
--- a/packages/react-cosmos/config.schema.json
+++ b/packages/react-cosmos/config.schema.json
@@ -121,10 +121,6 @@
       "description": "Custom renderer URL. This setting disables webpack bundling and allows integration with other build tools like Snowpack. [default: null]",
       "anyOf": [{ "type": "string", "minLength": 1 }, { "type": "null" }]
     },
-    "disableReload": {
-      "description": "Disable automatic location reload when webpack gets stuck updating modules during hot reload. [default: false]",
-      "type": "boolean"
-    },
     "dom": {
       "description": "DOM-related options",
       "type": "object",
@@ -155,6 +151,10 @@
         },
         "hotReload": {
           "description": "Enable webpack's Hot Module Replacement. [default: true]",
+          "type": "boolean"
+        },
+        "reloadOnFail": {
+          "description": "When using webpack's Hot Module Replacement, enable automatic location reload after webpack fails to update modules. [default: true]",
           "type": "boolean"
         }
       }
--- a/packages/react-cosmos/config.schema.json
+++ b/packages/react-cosmos/config.schema.json
@@ -121,6 +121,10 @@
       "description": "Custom renderer URL. This setting disables webpack bundling and allows integration with other build tools like Snowpack. [default: null]",
       "anyOf": [{ "type": "string", "minLength": 1 }, { "type": "null" }]
     },
+    "disableReload": {
+      "description": "Disable automatic location reload when webpack gets stuck updating modules during hot reload. [default: false]",
+      "type": "boolean"
+    },
     "dom": {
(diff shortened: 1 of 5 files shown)


skidding:
Hi @messman,

This is great. I love how detailed your PR description is!

How about putting the `disableReload` option inside the `webpack` object config?
It makes sense because it's webpack-related and we get to keep this option away from the Cosmos core. I won't get into the (undocumented...) Cosmos plugin system here but in your case the logic where you provide the default value for the `webpack.disableReload` option would be moved to here: [packages/react-cosmos/src/plugins/webpack/cosmosConfig/webpack.ts](https://github.com/react-cosmos/react-cosmos/blob/333dfd904a2e9d99b20fb086f984b1d21a43ebd9/packages/react-cosmos/src/plugins/webpack/cosmosConfig/webpack.ts).

messman:
> How about putting the `disableReload` option inside the `webpack` config object?

Ah yes, that makes sense. I see how keeping it close to `hotReload` is beneficial. I have updated the code.

I also modified the property to be `reloadOnFail` with a default of `true` rather than `disableReload` with a default of `false`. This is in spite of the argument I made earlier about consistency with webpack-hot-middleware. In my view, `hotReload` and `disableReload` as siblings inside the `webpack` config object would be confusing, and `reloadOnFail` is a better descriptor of the behavior. If you disagree, I can easily change it back to the prior discussed `disableReload` or use a different name. 👍

skidding:
`reloadOnFail` is even better.

Published a patch with this change in `react-cosmos@5.6.5`. Please try it out at let me know if it works as expected.

Thanks!

messman:
@skidding:

`react-cosmos@5.6.5` does have the fix and it works as expected. 🎉  Thanks!

Since the `cosmos.config.json` schema was updated, http://json.schemastore.org/cosmos-config needs to be updated as well. Is that baked in to the deploy process or is it manual?
Once that's handled everything is set.

skidding:
Updating the schema store is manual. Here are some PRs from the past: https://github.com/SchemaStore/schemastore/pulls?q=is%3Apr+author%3Askidding+is%3Aclosed – Do you want to do it?

messman:
> Do you want to do it?

Sure, not a problem. Thanks for the PR list – it was helpful to see the pattern.

Created SchemaStore/schemastore#1966, which also contains changes related to #1299 brought in from the copy.

PR #1366, opened 2022-01-22
React Cosmos v6 roadmap

skidding:


docs/roadmap/README.md
```diff
@@ -0,0 +1,38 @@
+# React Cosmos v6
+
+## Bundler
+
+- [ ] Revamp webpack integration.
+  - [ ] Make webpack and external package (and plugin).
```
brianespinosa, line 6: I am literally in the middle of looking at this today in our monorepo and wondering how to deal with potentially divergent versions of Webpack when it is impacted by different tools. I love that we are considering making it external.

docs/roadmap/README.md
```diff
@@ -0,0 +1,38 @@
...
+  - [ ] [Move from webpack middlewares to webpack-dev-server](https://github.com/react-cosmos/react-cosmos/issues/1272#issuecomment-733091647).
+  - [ ] Fast Refresh out of the box integration.
+- [ ] Integrate with Vite.
+
+## Core
+
+- [ ] Decide what to do with [Lazy fixture importing](https://github.com/react-cosmos/react-cosmos/pull/1313). Salvage what you can and move on. No need to linger if the complexity required to achieve this is unreasonable. Document everything learned for future reference.
+- [ ] Research: ES6 modules (React Cosmos with no bundler and no compiler).
```
brianespinosa, line 14: This would be fantastic.

docs/roadmap/README.md
skidding: Almost all Cosmos issues are webpack-related. Trying to stay compatible with all webpack versions (and all webpack config variations used by various frameworks) out of the box has already been crazy and it's only getting worse as time passes and more versions are released. Would be great to find a more scalable way to deal with webpack and bundler integration in general.

PR #1388, opened 2022-06-02
allow copy to clipboard from playground

fauxparse:
Clipboard functionality is blocked by CORS when Cosmos and the renderer are on different origins.

This change allows testing of components that copy text to the clipboard.


skidding:
@fauxparse Thanks!

PR #1406, opened 2022-12-19
Vite integration

skidding:


vlastoun:
When it will be released?
Do you have example configuration/project how can I include it?
Is it possible to build and serve cosmos with vite?

skidding:
> When it will be released?

@vlastoun Hard to tell the official release date because I'm preparing a major release with a bunch of other changes including a documentation revamp.

But if you're interested in providing early feedback on the Vite integration I can publish an alpha version of React Cosmos 6 sooner than I had planned.

> Do you have example configuration/project how can I include it?

There is a Vite example in the repository: https://github.com/react-cosmos/react-cosmos/tree/main/examples/vite. The setup is very simple.

I suspect we'll have to offer a way to pass a Vite config to Cosmos or to pass certain Vite options, but that will be easy to add once we decide it's needed.

> Is it possible to build and serve cosmos with vite?

Cosmos has two commands, `start` and `export`. With the Vite config, the first will start the Vite dev server and the second will build the Vite project and generate a static export. To serve the static export you'll have to use any basic http server and have it serve the static export directory. As simple as `http-server -p 5001 ./examples/vite/cosmos-export`.

PR #1450, opened 2023-04-21
Fix flag for generating userdeps in custom build setup guide

lr0pb:
I tried to use Cosmos with custom bundler setup and run into issue that `--generate-userdeps` flag mentioned in [`customBundlerSetup.md`](https://github.com/react-cosmos/react-cosmos/blob/main/docs/customBundlerSetup.md) does not generate file. Later I find in [Troubleshooting](https://github.com/react-cosmos/react-cosmos/blob/main/docs/customBundlerSetup.md) section of readme, that same behavior mentioned to fix with `--external-userdeps` flag. Tried this one and it works, so I filled this PR to fix flag name in custom bundler setup guide :)

Btw, overall experience in using it with custom bundler goes smooth, I used it with Parcel. Thanks for your awesome work! 🥇

### DIFF
--- a/docs/customBundlerSetup.md
+++ b/docs/customBundlerSetup.md
@@ -23,7 +23,7 @@ Next, choose a port for the renderer other than the main Cosmos port, say `5050`
 }
 ```
 
-Next, start Cosmos with the `--generate-userdeps` CLI flag. This will generate a `cosmos.userdeps.js` module that contains maps of user module imports (fixtures and decorators) as well the renderer config. Feel free to add this file to .gitignore.
+Next, start Cosmos with the `--external-userdeps` CLI flag. This will generate a `cosmos.userdeps.js` module that contains maps of user module imports (fixtures and decorators) as well the renderer config. Feel free to add this file to .gitignore.
 
 Finally, create a web server using your bundler of choice that serves an `index.html`, which loads a JS module with the following code:
 


skidding:
@lr0pb Awesome, thanks for this! :)

I've been thinking about renaming the `--external-userdeps` flag lately to something simpler but I haven't reached a conclusion yet (ideas welcome). In any case if I do rename it in a future release I'll send you a mention to let you know.

> Btw, overall experience in using it with custom bundler goes smooth, I used it with Parcel. Thanks for your awesome work! 🥇

🤘

lr0pb:
@skidding Well, if I right understand - without this flag same `userdeps` file or something related generated under the hood? In this case, it could be like `--expose-userdeps` or `--export-userdeps` or really `--generate-userdeps` ><
If flag mainly suited to use with custom bundler, it could be `--custom-bundler` / `--custom-renderer` or same variants with `with-` prefix (`--with-custom-renderer`)

skidding:
> @skidding Well, if I right understand - without this flag same `userdeps` file or something related generated under the hood? In this case, it could be like `--expose-userdeps` or `--export-userdeps` or really `--generate-userdeps` >< If flag mainly suited to use with custom bundler, it could be `--custom-bundler` / `--custom-renderer` or same variants with `with-` prefix (`--with-custom-renderer`)

Yeah, that's correct. When you use a plugin like the Vite or Webpack ones they call `generateUserDepsModule()` directly and inject it as a module under the hood, so no need to write the module on disk anymore.

Thanks for the suggestions!

skidding:
@lr0pb Got around to refactoring the "userdeps" module and flag.

- I liked your "expose" suggestion and decided to call it `--expose-imports`.
- Also renamed the file from `cosmos.userdeps.js` to `cosmos.imports.js`. I think the new name is more intuitive.
- If you have TypeScript installed a TS module will be generated instead as `cosmos.imports.ts`.
- You can override the path under `exposeImports` cosmos config option or by passing a value to the CLI arg: `cosmos --expose-imports custom-imports-file.js`.

Already changed in `6.0.0-alpha.17` or `react-cosmos@next`.

lr0pb:
Thank you, got it!
I am also found several other things today, will file issues for them

PR #1471, opened 2023-05-01
WIP

skidding:


packages/react-cosmos-core/src/userModules/userModuleTypes.ts
```diff
@@ -17,13 +14,7 @@ export type ReactFixtureModule = FixtureModule<ReactFixture>;
...
 export type LazyReactFixtureWrapper = LazyModuleWrapper<ReactFixtureModule>;
 
-export type ReactDecoratorProps = {
-  children: ReactNode;
-  fixtureState: FixtureState;
-  setFixtureState: SetFixtureState;
-  onErrorReset: () => unknown;
-};
-export type ReactDecorator = ComponentType<ReactDecoratorProps>;
+export type ReactDecorator = ComponentType<{ children: ReactNode }>;
```
lr0pb, line 17: Does this type should be used in decorators? E.g.
```tsx
// cosmos.decorator.tsx
import { ReactDecorator } from 'react-cosmos-core';

const decorator: ReactDecorator = ({ children }) => {
  return <div>{children}</div>;
};
```
If yes, makes sense to add a mention in https://github.com/react-cosmos/react-cosmos/tree/main/docs#decorators

skidding: Didn't realize I removed the decorator props type which could be useful for users. Thanks for pointing it out. I [exported it](https://github.com/react-cosmos/react-cosmos/blob/76ca48bb9aba923beba088acbe063d1ae5bd3a02/packages/react-cosmos-core/src/userModules/userModuleTypes.ts#L17) again in my last PR.

This way you can annotate the entire decorator or you can annotate the props only if you prefer a different syntax:

```tsx
import { ReactDecoratorProps } from 'react-cosmos-core';

export default function ({ children }: ReactDecoratorProps) => {
  return <div>{children}</div>;
};

// or....

export default ({ children }: ReactDecoratorProps) => <div>{children}</div>;
```

> Can open PR, but not sure about actual text style to write it with

I plan to add JS/TS examples for everything, but I'm not sure how to structure it yet. I'm going to think more about docs after I finish a few more things and publish the 6.0 beta.

PR #1516, opened 2023-05-29
Fix select controls to handle groups in reactions

max-scopp:
See https://github.com/react-cosmos/react-cosmos/discussions/1513


skidding:
Just realized this also needs attention:

https://github.com/react-cosmos/react-cosmos/blob/f442f14ad68600be918faaecaeba61d51e2fb074/packages/react-cosmos-renderer/src/fixture/useSelect/useCreateFixtureState.ts#L39

Stringifying the options list is a trick make the options list a dependency of that hook without invalidating the effect on every render even if the options list is identical. But stringifying an array of string is the equivalent of `array.join(',')` and this won't work anymore when the options can contain group objects. So this needs to be changed to `JSON.stringify(args.options)` or something else.

skidding:
@max-scopp are you still interested in getting this merged? I can help if needed :).

ovidiuch:
Figured out I had write access on the fork so I made the changes myself 😉

PR #1534, opened 2023-07-01
Rename Build plugins

skidding:


skidding:
In the end I decided to keep the "Server plugins" name because they enable more than just integrating with bundlers. Server plugins allow adding functionality to the Cosmos server that can be a backend for UI plugins for example, or any Cosmos server-specific functionality that isn't related to building or bundling of the user code.

PR #1559, opened 2023-11-08
feat: add fixture filename support

ThaisRobba:
Update adding support for `fixture` filenames, like so:

| Input | Output |
| ----- | ------- |
| Captura de Tela 2023-11-08 às 12 48 25 | Captura de Tela 2023-11-08 às 12 48 05 |

Added/modified tests and also manually tested with examples, tried to keep surface changes minimal.

@ovidiubute Please, let me know if it looks good or if there is anything you'd rather see done differently or if this would not be a desired change.


packages/react-cosmos-core/src/fixtureTree/flattenFixtureTree.tsx
```diff
@@ -22,7 +22,10 @@ export function flattenFixtureTree(
       const childNode = children[childName];
       const { data: childData } = childNode;
 
-      if (childData.type === 'fileDir')
+      if (
+        childData.type === 'fileDir' ||
+        (childData.type === 'fixture' && childName === 'fixture')
```
skidding, line 27: @ThaisRobba Is this change needed? The tests seem to pass without it and I think once the fixture tree was created the fixture names no longer matter when flattening it.

ThaisRobba: @skidding You are totally correct - not sure why it wasn't working until then for me (I probably forgot to rebuild at some point). I've reverted this change ^^

packages/react-cosmos-core/src/fixtureTree/createFixtureTree/collapseNamedIndexes.ts
```diff
@@ -37,3 +42,7 @@ export function collapseNamedIndexes(
 function noCaseEqual(a: string, b: string) {
   return a.toUpperCase() === b.toUpperCase();
 }
+
+function toCapitalCase(a: string) {
+  return a.replace(/^./, a[0].toUpperCase());
```
skidding, line 47: This might introduce unwanted results in the tree view in some cases. Here's what I mean.

Before, if you had something like `main-menu/MainMenu.fixture.tsx`, we'd use the fixture module name for the fixture name, and it would end up as `MainMenu` in Cosmos. Or if your project used all kebab case, `main-menu/main-menu.fixture.tsx` would end up as `main-menu` in Cosmos.

With this change, the folder name is used (which of course is the only option when you use the `/fixture.tsx` pattern), and the first letter is capitalized.

So `main-menu/MainMenu.fixture.tsx` becomes `Main-menu` in Cosmos. We could use a kebab case to camel case helper, but I don't think we want that assumption for all folks. Maybe the fixture should show as `main-menu` in Cosmos to follow the component folder name. We could add additional Cosmos config options here but as you know I'd like to keep it simple and find a convention that's reasonable in most cases and basically follow the user's file system.

So what do you think? Would it work for you if we kept the folder name unchanged in the tree view?
- So in the old scenario: `main-menu/MainMenu.fixture.tsx` ends up as `MainMenu`.
- So your new scenario: `main-menu/fixture.tsx` ends up as `main-menu`.
- And if you use camel case for folders, too: `MainMenu/fixture.tsx` ends up as `MainMenu`.

ThaisRobba: Makes a lot of sense - I was concerned using the folder name would break things (like how `dashboard/Dashboard.fixture.tsx` would become `dashboard` if using  the folder name).

I made a modification that keeps the old behavior and uses the folder name only when the fixture is "unnamed", ie: `fixture.tsx`

What do you think of it?
I would be ok with folder name, I just don't want to greatly break anything for others xD

skidding: Looks good to me now. Thanks for making the change!

One minor thing: Would you mind reverting the todo app example modules to what they were before? The unit tests should offer enough coverage for the old and new behavior.

While I've also used this pattern in the past, at the moment I stay away from using generic file names in my projects, like "index" (or "fixture" in this case), simply because when I do a file search (cmd + p in VS Code), or when I have tabs, they are harder to distingush. I'm sure with the right IDE config you can work around this but this is my personal preference at the moment.

ThaisRobba: Sure, no problem, I'm reverting the commit that changed the example code ^^

skidding:
@ThaisRobba Thanks for making this contribution!

FYI added a small change and improved the docs in https://github.com/react-cosmos/react-cosmos/pull/1575.

This has been published under `6.0.0-beta.7` or `@next` NPM dist tag.

Let me know if this works as expected once you upgrade in your project.

PR #1579, opened 2023-12-08
Allow UI plugins to extend initial fixture state

skidding:


birtles:
This looks good and would simplify the dark mode plugin quite a bit.

However, I wonder if it should be integrated with `setFixtureState` somehow? As I understand it, using this API a plugin such as the dark mode plugin would need to call _both_ `setFixtureState` and `setFixtureInitialState` every time the user or system changes the setting.

skidding:
> However, I wonder if it should be integrated with `setFixtureState` somehow? As I understand it, using this API a plugin such as the dark mode plugin would need to call _both_ `setFixtureState` and `setFixtureInitialState` every time the user or system changes the setting.

In my mind they are two different things:

- `setFixtureState` is changing the fixture state live and propagating it to all renderers connected (you can have more than one FYI).
- `extendInitialFixtureState` is deriving fixture state from plugin state, that is used as the initial fixture state whenever a new fixture is selected or a new renderer is connected or an existing renderer reloads.

Maybe I'm too deep in Cosmos internals and this could be simplified into a singur method call, but I can't think of an elegant way at the moment.

I created a very basic, stripped down version of the dark mode plugin that uses this new API. I'd like you to take a look and let me know if it makes sense, and if you have any ideas on how to simplify this minimal example I'll definitely take them into consideration for simplifying the API.

```tsx
import React from 'react';
import { RendererActionSlotProps, RendererCoreSpec } from 'react-cosmos-ui';
import { createPlugin } from 'react-plugin';

export type DarkModeSpec = {
  name: 'darkMode';
  state: {
    darkMode: boolean;
  };
};

const { onLoad, namedPlug, register } = createPlugin<DarkModeSpec>({
  name: 'darkMode',
  initialState: {
    darkMode: false,
  },
});

onLoad(pluginContext => {
  const { getState, getMethodsOf } = pluginContext;
  const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');

  // A) Extend initial fixture state with dark mode state (when renderer
  // connects or when fixture is changed or reloaded)
  rendererCore.extendInitialFixtureState(prevState => ({
    ...prevState,
    darkMode: getState().darkMode,
  }));
});

namedPlug<RendererActionSlotProps>(
  'rendererAction',
  'darkMode',
  ({ pluginContext }) => {
    const { getState, setState, getMethodsOf } = pluginContext;
    const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');

    const darkMode = getState().darkMode;

    function handleToggle() {
      setState({ darkMode: !darkMode }, () => {
        // B) Sync dark mode state with live connected renderer(s)
        rendererCore.setFixtureState(prevState => ({
          ...prevState,
          darkMode: !darkMode,
        }));
      });
    }

    return (
      <button onClick={handleToggle}>{darkMode ? 'dark' : 'light'}</button>
    );
  }
);

export { register };

if (process.env.NODE_ENV !== 'test') register();
```

birtles:
Thanks! When is `onLoad` called? In my testing it never got called. I thought it was for the initial load of the plugin in which case I think it wouldn't be sufficient to only call `extendInitialFixtureState` there because otherwise the state won't be applied to newly selected fixtures?

skidding:
> In my testing it never got called.

This is because of [this bug](https://github.com/react-cosmos/react-cosmos/pull/1578). I just published `6.0.0-beta.8` which should fix the undefined `getConfig` and `onLoad` not getting called.

> I thought it was for the initial load of the plugin in which case I think it wouldn't be sufficient to only call `extendInitialFixtureState` there because otherwise the state won't be applied to newly selected fixtures?

Yes `onLoad` is called once on plugin load. But `extendInitialFixtureState` doesn't update the fixture state when called, but rather registers a fixture state updater that gets applied whenever a new fixture state is created for a fixture. The updater stays active for the lifetime of the plugin.

birtles:
> > In my testing it never got called.
> 
> This is because of [this bug](https://github.com/react-cosmos/react-cosmos/pull/1578). I just published `6.0.0-beta.8` which should fix the undefined `getConfig` and `onLoad` not getting called.

Excellent, thank you!

> Yes `onLoad` is called once on plugin load. But `extendInitialFixtureState` doesn't update the fixture state when called, but rather registers a fixture state updater that gets applied whenever a new fixture state is created for a fixture. The updater stays active for the lifetime of the plugin.

Right so in my case, whenever the user or system changes the dark mode state, I want to update the current state of any fixtures being rendered _and_ update the initial state used for any fixtures that are loaded in future.

So from my naive mental model, I was hoping I could perform some sort of `setFixtureState()` operation and pass in a `persist` flag or something which means the changes are also applied to future fixtures.

It's very easy to just call `setFixtureState()` and `setInitialFixtureState()` but I thought I'd mention it in case you're still thinking about the shape of the API.

skidding:
@birtles Thinking a bit more about this...

It seems like the essence of the functionality needed here is to derive fixture state from plugin state (when that fixture state is created — fixture is selected, user change, etc — shouldn't matter that much).

Maybe if `react-plugin` offered an `onStateChange` method we could do without the `onLoad` logic, too.

In this example the plugin reacts to user or system events and only updates the plugin's state. And deriving the fixture state from the plugin state is separate:

```tsx
const { onStateChange, /* ... */ } = createPlugin<DarkModeSpec>(/* ... */);

onStateChange(pluginContext => {
  const { getState, getMethodsOf } = pluginContext;
  const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');

  return rendererCore.setFixtureState(
    prevState => ({ ...prevState, darkMode: getState().darkMode }),
    { persist: true }
  );
});
```

Notes:
1. We still need your `persist` idea added to `setFixtureState`.
2. `onStateChange` should be called once when plugin loads, too. Similar to how `useEffect` is always called on mount before dependencies change.

This idea solves the issue of having to call `setFixtureState` in two places.

What do you think of it?

skidding:
One, one more update after thinking even more about this :).

I think `onStateChange` is a nice to have but I'd leave that aside for now as it doesn't greatly improve our example.

---

`setFixtureState(updater, { persist: true })` is a nice API but it kinda breaks down when trying to implement it.

Here's why:

- Regular `setFixtureState()` is an imperative method that updates the current fixture state and fires a message to the renderer with the updated fixture state.
- Persisted `setFixtureState()` needs to do something different. It needs to store a fixture state change that will apply to all fixtures selected in the future. This is tricky because:
  1. At the moment fixture state is cleared between fixtures. The rest of the plugins expect this (props panel, state panel, controls panel).
  2. We need a way to undo a persisted fixture state change. This is why `extendInitialFixtureState` in my current implementation returns a callback, to call it when UI plugin unloads and to _unset_ this persisted fixture state change.

So something that works so differently and only returns an _undo_ callback in one case should probably be a separate method. Maybe `extendInitialFixtureState` is not intuitive, so I'll dump a few other suggestions:

- `setInitialFixtureState`
- `setGlobalFixtureState`
- `setPersistentFixtureState`
- `setPersistedFixtureState`

But to be honest I feel like not including "initial" in the name is also misleading. Because what this method does is add something to the initial fixture state when a fixture is selected. Afterwards any plugin can call `setFixtureState` and override the initial fixture state, which means we can't really ensure it's truly "persistent" or "global" (with the current architecture).

skidding:
Adding the rest of the convo from Discord here for posterity, to have the full context in case we revisit this design decision in the future:
 
> @birtles Thanks! I read through the GitHub issues. It makes sense and I can see the undo part is tricky. I don't fully appreciate that part (are plugins ever unloaded?) so catching up over a DM makes sense. Unfortunately our timezones don't overlap too well but I might be online tonight (i.e. morning your time I think) if that works.

> @skidding You’re right, our timezones does make it tricky. With regards to unloading, A) a Cosmos plugin can be enabled/disabled at runtime, the UI is still missing for that but hopefully as Cosmos 6 matures we’ll get to fully realize the concept behind the UI plugin architecture. But also B) If you setFixtuteState with persist=true in a plugin, maybe later you want to set it again with a different value. If you don’t have a way to clean up the first call both setFixtureState calls will persist and compete indefinitely?
>
> The more I think about it the more I’m inclined to go with my initial proposal. It’s not ideal that an API leaks implementation details but I’ll try to explain the reasoning behind my proposal a bit more.
> 
> Fixture state is a tricky concept. It’s a two-way binding on data shared between the Cosmos UI and Cosmos renderer(s), two modules which live in different environments and communicate asynchronously.
> 
> Moreover, there is no guarantee to each part that the other one is connected, or how long the other part will take to react to a state change.
> 
> For example:
> - With React Native you might load the renderer first and the Cosmos UI later.
> - Cosmos UI can connect to multiple renderers at the same time. There is one primary renderer which can manipulate the fixture state and secondary renderers which mirror the fixture state of the primary renderer.
> - A renderer can choose to fully reload therefore flushing its internal state, in which case the Cosmos UI needs to provision it with the previous fixture state.
> 
> The way setFixtureState works now is very naive. It expects renderers to be connected at the time of its call. It even prints a warning if no renderer is connected.
> 
> This is because until now it was more popular for the renderer to manipulate the fixture state than the Cosmos UI.
> 
> For your use case, and others similar I can envision, you need a more robust way to provide fixture state to all present and future renderers.
> 
> If you look at the extendInitialFixtureState example I made for you, you’ll notice it is fundamentally different than setFixtureState, even if you ignore the cleanup part.
> 
> In conceptual terms it works more like a web hook, it provides Cosmos a callback that can be called at any time and will return fixture state based on the current plugin state. So as the plugin state changes, when this callback is called it will return a different fixture state based on the plugin state, which will be applied when the Cosmos UI needs it (new fixture, reloaded fixture, new renderer, etc).
> 
> What I like about this approach is it reduces the need to think about timing and avoids race conditions. It’s more functional.
> 
> Hopefully this clarifies things a bit. I hate to be the one overly rationalizing an ambiguous API :), but every time I tried to rewrite my PR with a simpler API I ran into these complexities. 

> @birtles Hi! Thanks for writing up all that. You know the model best and I can certainly make the setInitialFixtureState approach work for my case. I think it might be possible to avoid having a cleanup function but it sounds like that's not the biggest issue.

PR #1581, opened 2023-12-13
Isolate fixture state between plugins

skidding:


skidding:
The saga continues 😅

I gave this another go and I might be on to something. I came at the problem from a different angle with two changes in mind:

1. Leave the current `setFixtureState` aside and come up with a new method that works for the dark mode plugin use case (ie. plugin that controls global fixture state). It can be pretty simple if we abandon the concept of a state updater (`prevState => nextState`) and instead accept a "slice" of fixture state as an argument. So in the simplified example we would just call `setGlobalFixtureState({ darkMode: true })`. No need to extend previous fixture state, it extends it automatically. This diminishes the issues related to unsetting a state updater that I was worried about. The plugin can just set `darkMode` to `undefined` on unload, or we can add an additional `unsetGlobalFixtureState('darkMode')` method if needed for clean up purposes.
    - `setGlobalFixtureState` is called once when the plugin loads, based on the plugin's config, and once when the button is toggled. But it's the same function call in both places.
2. The second part is more related to the plugin implementation. I was suggesting to use plugin state and syncronize that with the fixture state. This is bad practice as we're not having a single source of truth. Fixture state is already _state_, and in fact it's also Cosmos UI plugin state (just that of the `rendererCore` plugin). The dark mode plugin can just read the plugin config and dump it straight into the fixture state.

With these two changes the minimal version of the plugin is looking much cleaner:

```tsx
import React from 'react';
import { RendererActionSlotProps, RendererCoreSpec } from 'react-cosmos-ui';
import { createPlugin } from 'react-plugin';

export type DarkModeState = {
  darkMode?: boolean;
};

export type DarkModeSpec = {
  name: 'darkMode';
  config: {
    darkMode: boolean;
  };
};

const { onLoad, namedPlug, register } = createPlugin<DarkModeSpec>({
  name: 'darkMode',
  defaultConfig: {
    darkMode: false,
  },
});

onLoad(pluginContext => {
  const { getConfig, getMethodsOf } = pluginContext;
  const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');
  return rendererCore.setGlobalFixtureState({ darkMode: getConfig().darkMode });
});

namedPlug<RendererActionSlotProps>(
  'rendererAction',
  'darkMode',
  ({ pluginContext }) => {
    const { getMethodsOf } = pluginContext;
    const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');
    const { darkMode = false } = rendererCore.getFixtureState<DarkModeState>();

    function handleToggle() {
      rendererCore.setGlobalFixtureState({ darkMode: !darkMode });
    }

    return (
      <button onClick={handleToggle}>{darkMode ? 'dark' : 'light'}</button>
    );
  }
);

export { register };

if (process.env.NODE_ENV !== 'test') register();
```

I got a prototype working together with this decorator:

```tsx
import React, { useContext } from 'react';
import { FixtureContext } from 'react-cosmos-renderer/client';

export type DarkModeState = {
  darkMode?: boolean;
};

export function DarkModeDecorator({ children }: { children: React.ReactNode }) {
  const { fixtureState } = useContext(FixtureContext);
  const { darkMode = false } = fixtureState as DarkModeState;

  return (
    <>
      <div>Theme: {darkMode ? 'dark' : 'light'}</div>
      {children}
    </>
  );
}
```

@birtles I know I've been bombarding you with these changes but when you get the chance I'm curious if you feel like this API is as simple as you'd like it to be as a plugin author.

birtles:
No problem, API design is hard. I spent years on the Web Animations API and still have a lot of regrets about that 😅

Stream of consciousness thoughts:

- `setGlobalFixtureState` makes sense to me as something that is further reaching than `setFixtureState`.
- If you are worried about clean up, once possibility is to namespace plugins' access to the fixture state. That might cause more problems than its worth but in a world where plugins are common, it might make Cosmos more reliable. For example, if my plugin claims the namespace "darkMode" and another rival dark mode plugin also tries to claim that, then Cosmos could produce an error rather than having the two plugins tread on each others' toes. You could even let the package name be the namespace so that npm becomes the arbiter of names and plugins don't need to specify it making the namespace-ing transparent.
  - The main drawback is if you have two plugins that want to share some config state, only one would be able to update it via `setGlobalFixtureState`. Perhaps that could be addressed by a future API, however.
- Yes, with the API you proposed I was envisioning _not_ calling `setState` and just using fixture state and local plugin state (i.e. React state)—the state we pass to the fixture differs from the state we want to track in the plugin since the plugin needs to know "are we tracking system dark mode or has the user overridden the setting?" whereas the fixture only needs to know "what class do I need to set".

So, in summary, the API looks great. Ship it!

skidding:
> No problem, API design is hard. I spent years on the Web Animations API and still have a lot of regrets about that 😅

That's impressive! Good to know Cosmos is getting feedback from a professional API designer 😁.

> `setGlobalFixtureState` makes sense to me as something that is further reaching than `setFixtureState`.

Yes, in fact `setGlobalFixtureState` does call `setFixtureState` when a fixture is loaded, but also persists it for the future. Now, you might argue that this is the perfect description for your `persist: true` option idea and you'd be right. Only reason I didn't go that route is because `setGlobalFixtureState` receives a value and `setFixtureState` can receive a `prevState => nextState` state updater callback, too. Implementing a state updater for `setGlobalFixtureState` seemed awkward when I tried it, which led me to the previous, more confusing implementations. Merging them might still be doable but for now I think the 2 distinct methods for setting fixture state is clear enough. Especially since a plugin will most likely only use one or the other.

> If you are worried about clean up, once possibility is to namespace plugins' access to the fixture state. That might cause more problems than its worth but in a world where plugins are common, it might make Cosmos more reliable. For example, if my plugin claims the namespace "darkMode" and another rival dark mode plugin also tries to claim that, then Cosmos could produce an error rather than having the two plugins tread on each others' toes. You could even let the package name be the namespace so that npm becomes the arbiter of names and plugins don't need to specify it making the namespace-ing transparent.

This is a good idea and I went with it. Not the strictness aspect of it, at least for now, but the API itself. I like the idea of namespacing the fixture state, because it doesn't really make sense from a plugin's standpoint to access Cosmos' entire fixture state. It actually makes it harder for the plugin to work with the fixture state.

This led me to a big detour because something clicked. All the built-in Cosmos plugins that used fixture state were relying on internal types and helpers. So I started refactoring stuff to make sure the built-in plugins were using the same fixture state APIs meant for outside consumption.

It all started with this change in the fixture state type:

```diff
-export type FixtureState = {
-  props?: FixtureStateProps[];
-  classState?: FixtureStateClassState[];
-  controls?: FixtureStateControls;
-} & Record<string, unknown>;
+export type FixtureState = Record<string, unknown>;
```

You can imagine the amount of changes that generated...

But all changes revolved around the same thing: In the renderer and in the Cosmos UI, fixture state is accessed as a slice instead of a whole. There's a new `useFixtureState<T>(name)` fixture hook that is much more friendly than using the `FixtureContext` directly (just realized I need to update the docs on this). And the `rendererCore` UI plugin also returns and allows setting fixture state by name with a generic type argument.

> Yes, with the API you proposed I was envisioning _not_ calling `setState` and just using fixture state and local plugin state (i.e. React state)—the state we pass to the fixture differs from the state we want to track in the plugin since the plugin needs to know "are we tracking system dark mode or has the user overridden the setting?" whereas the fixture only needs to know "what class do I need to set".

OK, well the new APIs should make it easier to do this and I'm curious what the end result for the dark mode plugin will be.

> So, in summary, the API looks great. Ship it!

Just shipped in `6.0.0-beta.9`!

Here's my updated minimal plugin example just to see the updated APIs in action

```tsx
import React from 'react';
import { RendererActionSlotProps, RendererCoreSpec } from 'react-cosmos-ui';
import { createPlugin } from 'react-plugin';

export type DarkModeSpec = {
  name: 'darkMode';
  config: {
    darkMode: boolean;
  };
};

const { onLoad, namedPlug, register } = createPlugin<DarkModeSpec>({
  name: 'darkMode',
  defaultConfig: {
    darkMode: false,
  },
});

onLoad(pluginContext => {
  const { getConfig, getMethodsOf } = pluginContext;
  const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');
  rendererCore.setGlobalFixtureState('darkMode', getConfig().darkMode);
});

namedPlug<RendererActionSlotProps>(
  'rendererAction',
  'darkMode',
  ({ pluginContext }) => {
    const { getMethodsOf } = pluginContext;
    const rendererCore = getMethodsOf<RendererCoreSpec>('rendererCore');
    const darkMode = rendererCore.getFixtureState<boolean>('darkMode');

    function handleToggle() {
      rendererCore.setGlobalFixtureState<boolean>('darkMode', !darkMode);
    }

    return (
      <button onClick={handleToggle}>{darkMode ? 'dark' : 'light'}</button>
    );
  }
);

export { register };

if (process.env.NODE_ENV !== 'test') register();
```

```tsx
import React from 'react';
import { useFixtureState } from 'react-cosmos/client';

export function DarkModeDecorator({ children }: { children: React.ReactNode }) {
  const [darkMode, setDarkMode] = useFixtureState<boolean>('darkMode');

  function handleToggle() {
    // This doesn't make sense for the dark mode plugin, it's just an example
    // of setting the fixture state from the fixture as well
    setDarkMode(prev => !prev);
  }

  return (
    <>
      <button onClick={handleToggle}>
        Theme: {darkMode ? 'dark' : 'light'}
      </button>
      {children}
    </>
  );
}
```

birtles:
> > `setGlobalFixtureState` makes sense to me as something that is further reaching than `setFixtureState`.
> 
> Yes, in fact `setGlobalFixtureState` does call `setFixtureState` when a fixture is loaded, but also persists it for the future. Now, you might argue that this is the perfect description for your `persist: true` option idea and you'd be right. Only reason I didn't go that route is because `setGlobalFixtureState` receives a value and `setFixtureState` can receive a `prevState => nextState` state updater callback, too. Implementing a state updater for `setGlobalFixtureState` seemed awkward when I tried it, which led me to the previous, more confusing implementations. Merging them might still be doable but for now I think the 2 distinct methods for setting fixture state is clear enough. Especially since a plugin will most likely only use one or the other.

I think it's fine too given the parameters you pass are different in nature.

> This led me to a big detour because something clicked. All the built-in Cosmos plugins that used fixture state were relying on internal types and helpers. So I started refactoring stuff to make sure the built-in plugins were using the same fixture state APIs meant for outside consumption.

I saw a whole bunch of commits come in so I figured there was some major refactoring going on!

> But all changes revolved around the same thing: In the renderer and in the Cosmos UI, fixture state is accessed as a slice instead of a whole. There's a new `useFixtureState<T>(name)` fixture hook that is much more friendly than using the `FixtureContext` directly (just realized I need to update the docs on this). And the `rendererCore` UI plugin also returns and allows setting fixture state by name with a generic type argument.

Yes, slicing is a better term. I think everyone familiar with Redux etc. will be familiar with that.

> > Yes, with the API you proposed I was envisioning _not_ calling `setState` and just using fixture state and local plugin state (i.e. React state)—the state we pass to the fixture differs from the state we want to track in the plugin since the plugin needs to know "are we tracking system dark mode or has the user overridden the setting?" whereas the fixture only needs to know "what class do I need to set".
> 
> OK, well the new APIs should make it easier to do this and I'm curious what the end result for the dark mode plugin will be.

Thanks, I think it will be a lot simpler with this new API.
 
> > So, in summary, the API looks great. Ship it!
> 
> Just shipped in `6.0.0-beta.9`!

Great! I'll update my plugin sometime this week and let you know!

> Here's my updated minimal plugin example just to see the updated APIs in action

Thank you! That's very helpful.

birtles:
@skidding I had a go at integrating this but at [this line](https://github.com/birchill/react-cosmos-dark-mode/blob/5a6bc9288756aae341a47b8848d90d394bd85aa0/src/index.tsx#L34) I get:

```
TypeError: getConfig is not a function
```

skidding:
> @skidding I had a go at integrating this but at [this line](https://github.com/birchill/react-cosmos-dark-mode/blob/5a6bc9288756aae341a47b8848d90d394bd85aa0/src/index.tsx#L34) I get:
> 
> ```
> TypeError: getConfig is not a function
> ```

@birtles See https://github.com/birchill/react-cosmos-dark-mode/pull/1/files

PR #1662, opened 2024-11-28
refactor: migrate from NativeModule to TurboModuleRegistry

ThaisRobba:
## Context

Expo 52 seems to have changed how it deals with the NativeModules, probably because of defaulting to the new architecture. This migrates the `getSocketURL` to use TurboModuleRegistry, which works with both architectures consistently.

## Testing

I've tested on an expo project with and without the new architecture, both rendered correctly.

I have not tested on a pure RN project but it *should* work, would appreciate if someone with a project that knowingly has this issue pitched in so that we can validate if this fix works.

I tried to keep things as much as they were and I'm open to any and all criticism :)


ovidiuch:
Thanks @ThaisRobba!

PR #1668, opened 2024-12-16
Upgrade to React 19

ovidiuch:
To read
 - https://react.dev/blog/2024/04/25/react-19-upgrade-guide#changes-to-the-reactelement-typescript-type
 - https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref
 - https://react.dev/blog/2024/12/05/react-19#cleanup-functions-for-refs
 - https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-react-test-renderer

To do
- [x] Fix props type errors.
- [x] Replace `element.ref` with `element.props.ref`. **This is a breaking change.**
  - [x] Fix ref type errors.
  - [x] Implement ref cleanup when wrapping user refs and add tests.
  - [x] ~~Remove `forwardRef` support, or keep it for backwards compatibility?~~ There's only [one reference to ForwardRef](https://github.com/react-cosmos/react-cosmos/blob/97b077ae5c2643c36c9444a9b3bff1f9b2154e2c/packages/react-cosmos-renderer/src/fixture/FixtureCapture/classState/decorateFixtureRefs/isRefSupported.ts#L17), which we'll keep for now because forwardRef is [still supported](https://react.dev/reference/react/forwardRef) in React 19 even though it's deprecated.
- [x] Replace deprecated `react-test-renderer`.
  - [x] Fix `getFixtures()` test.
  - [x] Fix `act()` warnings.
  - [x] Fix ref tests.
- [x] Fix missing `react-native` dependency error.
- [x] ~~Upgrade Nextra~~ Remove Nextra docs from monorepo workspaces (to keep using Nextra 2 with React 18). Done in #1672
  - [x] ~~See if you can `nohoist` next and keep docs in the monorepo? https://github.com/zackerydev/noist~~ It works, but a lot of indirect next & nextra dependencies need to be hoisted which isn't manageable over time.


shiftyp:
What are the outstanding tasks on this PR? Anything you might need help with?

ovidiuch:
> What are the outstanding tasks on this PR? Anything you might need help with?

Thanks, but nothing I can think of at the moment. Most of the items in the task description are done or partially done since December. Tests are already passing, except for the Nextra docs build which hasn't been upgraded. I think what's left is some cleanup (some temporary `@ts-ignore`s to sort out and the react-test-renderer deprecation)  and double-triple checking to make sure everything works well with the changes made. I just haven't gotten the time to sit with this PR since before the holidays. My goal is to do finish it and create a release this month, hopefully I won't run into any complications. In any case I'll post updates here as soon as I manage to carve some time and pick up where I left off.

ovidiuch:
@shiftyp Just pushed a new canary version with full support for React 19. This will likely be released as a new major Cosmos version after some testing. For now please install `react-cosmos@canary` (same for all Cosmos packages) and let me know if everything works well for you. Thanks!

PR #1678, opened 2025-02-04
Update decorator docs with global CSS import guidance

receter:
Added instructions to import global CSS (e.g., normalize.css) via the `globalImports` option in the Cosmos config.

### Explanation of the Problem We Were Solving

When using React Cosmos, importing global CSS files (like `normalize.css`) within a decorator file can lead to unintended styling issues. Specifically, these styles end up being appended after fixture-specific styles, which disrupts the intended CSS cascade. This can cause global styles to override fixture styles, potentially leading to inconsistent or unexpected UI behavior.

To address this issue, the solution was to move the import of global CSS files out of the decorator and instead add them to the `globalImports` array in the Cosmos configuration. This ensures that the global CSS is loaded first, establishing a consistent base for all subsequent styles in the application.

By making this change, we improve the clarity of the documentation and provide a clear, recommended approach for handling global CSS in React Cosmos projects.

### DIFF
--- a/docs/pages/docs/fixtures/decorators.mdx
+++ b/docs/pages/docs/fixtures/decorators.mdx
@@ -30,3 +30,22 @@ A decorator file can also export an array of decorators:
 ```jsx filename="cosmos.decorator.js"
 export default [Decorator1, Decorator2, Decorator3];
 ```
+
+## Importing Global CSS
+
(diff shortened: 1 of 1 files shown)


ovidiuch:
@receter Useful addition to the docs, thanks!

PR #1684, opened 2025-03-09
Redesign home screen

ovidiuch:
Realization: It's not a good UX to rely on the renderer preview for the Cosmos UI main content. An overlay is still needed to react instantly to selecting/deselecting fixtures. Plus the hack to hide the "No fixture selected." message inside iframes doesn't work when renderer is server-side.


packages/react-cosmos-dom/src/DomFixtureLoader.tsx
```diff
@@ -37,5 +41,9 @@ const containerStyle: React.CSSProperties = {
 };
 
 function renderMessage(msg: string) {
+  // XXX: This is a hack to prevent the "No fixture selected." message from
+  // showing up when the renderer is loaded inside the Cosmos UI iframe.
+  if (msg === 'No fixture selected.' && isInsideWindowIframe()) return null;
```
ovidiuch, line 46: This doesn't work on the server.

PR #1704, opened 2025-04-06
Vite error handling & onboarding

ovidiuch:
- [x] Generate html file automatically if user doesn't have an index.html
- [x] Resolve relative vite/ paths without trying to resolve `vite` module path.
- [x] Improve main script detection and communicate failure to detect it. Pick up all script tags from index.html. Match any when single, match based on common src when multiple.
- [x] Auto add main script to index.html when no script exists.
- [x] Rename `vite.indexPath` to `vite.mainScriptUrl` and no longer check the file system.

> Refactor: Find main script URL (src attribute) instead of resolving the main script file path. This allows resolving a virtual script file and makes it possible to run Cosmos without any main script file. As for the Rollback plugin, `transform` can only be used when a script file exists, so it's between the `resolveId` and `load` hooks. The `load` hook is OK but the `id` will differ between existing and virtual files—in the former it's the original URL and in the latter it's the resolved file path. The avoid this inconsistency the `resolveId` hook can be used, provided the `enforce: 'pre'` option is used on the plugin.

Clean up
- [x] Decide whether to add `parse5` dependency or remove it.
- [x] Capitalize Cosmos, Vite & Webpack in log messages.
- [x] Only log `Using vite config...` once.


ovidiuch:
After going back to the drawing board a couple of times I'm finally happy with the result here. I've rethinked the Vite renderer integration from the ground up. Here's a summary.

Some people don't have an `index.html` when they try to use Cosmos with Vite because the host project is set up to be a component library. Because a Vite project revolves around an `index.html` though, it's still needed. So Cosmos will now generate a default `index.html` for you on start if it doesn't already exists.

The `vite.indexPath` setting has been renamed to `vite.mainScriptUrl` and it works completely different. Cosmos no longer cares about the file system when dealing with the main script because it replaces it with the Cosmos renderer at run time anyway. So instead of dealing with files and file paths like before Cosmos now reads the `index.html` contents and finds all the script tags.

The main script URL is detected based on common patterns (eg `"/src/main.tsx"`). If more matches are found then the best match is chosen. If there's just a script tag then it uses it as the main script URL. If `vite.mainScriptUrl` is configured then it looks for that one.

In all cases **the main script module doesn't even need to exist**, because Cosmos hooks into Vite's module resolution and automatically injects the Cosmos renderer under the main script URL.

Finally, if the user has an `index.html` without any script tags, Cosmos will transform the index.html on the fly to add a default main script that will then be intercepted to inject the Cosmos renderer.

So in conclusion:

- An index.html is no longer required.
- A script tag is no longer required.
- The main script URL will be detected automatically more intelligently.
- The main script URL doesn't need point to a real module.

There are some cases when Cosmos will fail with a message:

- `vite.mainScriptUrl` is set but it doesn't match any of the user scripts.
- Cosmos finds multiple scripts but none of the URLs can be clearly distinguished as the main one (eg. `"/src/foo.tsx"` and `"/src/bar.tsx"`). In this case `vite.mainScriptUrl` needs to be set.

Hopefully this makes life more easier for future React Cosmos Vite user. Please install the latest `canary` version to try this out. I encourage you to throw anything at it and try to break it because I'm curious to know if I missed something!

PR #1733, opened 2025-07-26
Toolbar position

ovidiuch:


ovidiuch:
Dropping this idea because having the renderer header on the bottom doesn't feel good in practice. But opened a PR to be able to come back to this later if needed.

PR #1772, opened 2026-08-28
Fix serving assets from paths containing dot directories

### LINKED ISSUE #1771: React Cosmos 7.4 fails to serve assets when installed with pnpm
### What's wrong?

React Cosmos 7.4.0 starts successfully when installed with pnpm, but loading
the playground in a browser fails because `/playground.bundle.js` returns 404.

The server logs:

```text
NotFoundError: Not Found
    at createHttpError (.../node_modules/.pnpm/send@1.2.1.../send/index.js:861:12)
    at SendStream.error (.../send/index.js:168:31)
    at SendStream.pipe (.../send/index.js:468:14)
    at sendfile (.../express/lib/response.js:1014:8)
    at ServerResponse.sendFile (.../express/lib/response.js:411:3)
    at .../react-cosmos/dist/devServer/expressApp.js:11:13
```

React Cosmos 7.4 upgraded from Express 4.22.1 / `send` 0.19.0 to
Express 5.2.1 / `send` 1.2.1.

With pnpm, the resolved playground bundle has an absolute path containing a
`.pnpm` directory, for example:

```text
.../node_modules/.pnpm/react-cosmos-ui@7.4.0.../node_modules/react-cosmos-ui/dist/playground.bundle.js
```

`send` 1.2.1 treats the intermediate `.pnpm` directory as a dotfile path and
returns 404 with the default `dotfiles: "ignore"` behavior.

The favicon and UI plugin endpoints appear to be affected for the same reason.

### Steps to reproduce

1. Install React Cosmos 7.4.0 in a project using pnpm.
2. Run `pnpm cosmos`.
3. Open the reported URL in a browser.
4. Observe that `/playground.bundle.js` returns 404 and the error above is
   printed by the server.

The same project works with React Cosmos 7.3.0.

### Screenshots

Not applicable.

### Useful info

- React Cosmos: 7.4.0
- Package manager: pnpm 11
- Node.js: 24.19.0
- Express: 5.2.1
- `send`: 1.2.1

I confirmed the behavior directly with the resolved bundle path:

```text
res.sendFile(filePath)                         -> 404
res.sendFile(filePath, { dotfiles: "allow" }) -> 200
```

### Additional context

The affected calls are in:

- `packages/react-cosmos/src/devServer/expressApp.ts`
- `packages/react-cosmos/src/corePlugins/pluginEndpointPlugin.ts`

One possible fix is to pass the filename relative to its containing directory:

```ts
res.sendFile(path.basename(filePath), {
  root: path.dirname(filePath),
});
```

This allows package-manager paths containing `.pnpm` without globally allowing
a requested filename that is itself a dotfile.

Alternatively, `{ dotfiles: "allow" }` fixes these exact-file routes.

ovidiuch:
Express 5 (send >= 1) responds with 404 when any segment of the served path is a dotfile. Package managers like pnpm resolve modules inside a .pnpm directory, so every res.sendFile call with an absolute path 404'd: the playground bundle, its sourcemap, the favicon and UI plugin modules.

Send the file name relative to its own directory instead, which scopes the dotfile check to the file we actually serve rather than to wherever the package happens to live on disk.

Fixes #1771

### DIFF
--- a/packages/react-cosmos/src/corePlugins/pluginEndpointPlugin.ts
+++ b/packages/react-cosmos/src/corePlugins/pluginEndpointPlugin.ts
@@ -1,6 +1,7 @@
 import path from 'node:path';
 import type { Request, Response } from 'express';
 import type { CosmosServerPlugin } from '../cosmosPlugin/types.js';
+import { sendFile } from '../shared/sendFile.js';
 import { resolveSilent } from '../utils/resolveSilent.js';
 
 export const pluginEndpointPlugin: CosmosServerPlugin = {
@@ -26,7 +27,7 @@ export const pluginEndpointPlugin: CosmosServerPlugin = {
         return;
       }
 
-      res.sendFile(resolvedPath);
+      sendFile(res, resolvedPath);
     });
   },
 };
--- a/packages/react-cosmos/src/devServer/__tests__/devServerUiPlugin.ts
+++ b/packages/react-cosmos/src/devServer/__tests__/devServerUiPlugin.ts
@@ -19,10 +19,23 @@ const port = 5000 + viteWorkerId();
 const testFsPath = path.join(__dirname, '../__testFs__');
 const pluginPath = path.join(testFsPath, `plugin-${viteWorkerId()}`);
 
+// Package managers like pnpm resolve modules inside a dot directory
+const dotPluginPath = path.join(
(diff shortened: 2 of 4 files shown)


birtles:
Looks good to me!

birtles (approved):

